Issue #2, opened 2015-09-26
Impossible to install via pip - no links on PyPI page


lorenzo-pasa:
When trying to install Voc as suggested in the documentation:
```
pip install voc
```
an error is raised:
```
Collecting voc
  Could not find a version that satisfies the requirement voc (from versions: )
  No matching distribution found for voc
```
The version of pip is not really the problem (it doesn't work with pip 6.1.1 nor with pip 7.1.2).
Trying to install it with the verbose option leads to the probable root cause:
```
(voc)p@pc:~/voc$ pip install -v voc
Collecting voc
  Getting page https://pypi.python.org/simple/voc/
  Starting new HTTPS connection (1): pypi.python.org
  "GET /simple/voc/ HTTP/1.1" 200 111
  1 location(s) to search for versions of voc:
  * https://pypi.python.org/simple/voc/
  Getting page https://pypi.python.org/simple/voc/
  Analyzing links from page https://pypi.python.org/simple/voc/
  Could not find a version that satisfies the requirement voc (from versions: )
Cleaning up...
No matching distribution found for voc
Exception information:
Traceback (most recent call last):
  File "/home/p/.virtualenvs/voc/local/lib/python2.7/site-packages/pip/basecommand.py", line 211, in main
    status = self.run(options, args)
  File "/home/p/.virtualenvs/voc/local/lib/python2.7/site-packages/pip/commands/install.py", line 294, in run
    requirement_set.prepare_files(finder)
  File "/home/p/.virtualenvs/voc/local/lib/python2.7/site-packages/pip/req/req_set.py", line 334, in prepare_files
    functools.partial(self._prepare_file, finder))
  File "/home/p/.virtualenvs/voc/local/lib/python2.7/site-packages/pip/req/req_set.py", line 321, in _walk_req_to_install
    more_reqs = handler(req_to_install)
  File "/home/p/.virtualenvs/voc/local/lib/python2.7/site-packages/pip/req/req_set.py", line 461, in _prepare_file
    req_to_install.populate_link(finder, self.upgrade)
  File "/home/p/.virtualenvs/voc/local/lib/python2.7/site-packages/pip/req/req_install.py", line 250, in populate_link
    self.link = finder.find_requirement(self, upgrade)
  File "/home/p/.virtualenvs/voc/local/lib/python2.7/site-packages/pip/index.py", line 571, in find_requirement
    'No matching distribution found for %s' % req
DistributionNotFound: No matching distribution found for voc
```
[...]

PR #4, opened 2015-09-27
Add important information to README

lorenzo-pasa:
This pull request adds to the README file some important information:
 * an alternative way to install VOC via pip (as discussed in issue #2 )
 * Python 3.x is required to run VOC (otherwise [some lambda functions raise errors](https://github.com/pybee/voc/blob/master/voc/java/klass.py#L92), due to `print` being a statement in Python 2.x)

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -89,6 +89,11 @@ Documentation
 
 Documentation for VOC can be found on `Read The Docs`_.
 
+Supported Python Versions
+-------------------------
+
+VOC requires Python 3.x
+
 Why "VOC"?
 ----------
 
--- a/README.rst
+++ b/README.rst
@@ -52,9 +52,16 @@ and development effort.
 Quickstart
 ----------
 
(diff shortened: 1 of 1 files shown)


freakboy3742:
Thanks of the report - I tweaked the text, but should have all the same details.

lorenzo-pasa:
Thanks... also for having fixed the issue about pip installation !

Actually, it seems there is another small imprecision in the Quickstart section: the relative path to the file *example.class* when running the `java` command.
Instead of just `example` it should be `org.pybee.example`
That is, the complete command should be:
`$ java -XX:-UseSplitVerifier -classpath python.jar:. org.pybee.example`

It's such a small correction that I'm not sure if open an issue, create a pull request or simply notify you about it with this comment =)

freakboy3742:
Thanks - I've just corrected that extra problem. I swear that sometimes, I actually read the instructions that I write... :-)

PR #26, opened 2015-11-15
Str.__add__ and Str.__multiply__ implementation

petrblahos:
Implemented __add__ between a str and anything.
Implemented __mul__ between a str and int.
A small clarification of the docs.


freakboy3742:
Thanks for the contribution, Petr!

PR #57, opened 2015-12-25
Explicitly use python3 instead of python

rubacalypse:
Some str tests were failing (e.g. str_add_bool) even though python3 and Java
were producing identical error messages. The problem was that the command
invocation was using "python" which ran python2 on my OSX machine.

### DIFF
--- a/tests/utils.py
+++ b/tests/utils.py
@@ -117,7 +117,7 @@ def runAsPython(test_dir, main_code, extra_code=None, run_in_function=False, arg
         args = []
 
     proc = subprocess.Popen(
-        ["python", "test.py"] + args,
+        ["python3", "test.py"] + args,
         stdin=subprocess.PIPE,
         stdout=subprocess.PIPE,
         stderr=subprocess.STDOUT,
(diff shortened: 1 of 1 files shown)


tests/utils.py
```diff
@@ -117,7 +117,7 @@ def runAsPython(test_dir, main_code, extra_code=None, run_in_function=False, arg
         args = []
 
     proc = subprocess.Popen(
-        ["python", "test.py"] + args,
+        ["python3", "test.py"] + args,
```
freakboy3742: Agreed that this is a good safety mechanism; however, it would probably be better to use `sys.executable`, which will ensure that the subprocess uses *exactly* the same Python interpreter as the test environment, and allow for a couple of odd situations where the executable won't be called `python`.

ankush981:
@rubacalypse Well, if you follow the instructions, you will end up with `python3` as the default command. I had no problems with this as `virtualenv -p $(which python3) env` sets up the environment as desired.

freakboy3742:
To be fair - at the time @rubacalypse opened this ticket, the instructions didn't exist :-)

On the basis that there are now clear instructions that should avoid the python3 confusion, I'm going to close this ticket.

freakboy3742:
Actually... rethinking this - `sys.executable` is better practice in general anyway.

ankush981:
@freakboy3742 Never heard of `sys.executable`, to be honest! :stuck_out_tongue_winking_eye: But do give it a thought and let us know.

PR #60, opened 2016-01-01
Corrected typo where command is -XX instead of -CC

bsdtux:
While going through the ReadMe.rst I noticed that it used -XX in the command but in the description afterwards it used -CC. After testing on my machine and looking through the documentation or jvm options I did not see any reference to -CC as a set of options.


freakboy3742:
Thanks for the report! I'm guessing autocorrect got a little overzealous.

PR #63, opened 2016-01-02
Updated Tutorials and Getting Started

bsdtux:
Created a first pass at an initial getting-started guide and updated the tutorial-0 to use a working example.py file.


freakboy3742:
Thanks! Great contribution!

PR #70, opened 2016-01-04
Fixed inplace test 

bsdtux:
Test was causing an anomaly where it was the inplace tests to use a tuple even when specifying a string.


freakboy3742:
Thanks for the contribution @bsdtux, but I've fixed this one independently in 6d112c8472475c807e9520038eff1661e977c575, along with correcting all the tests that were failing.

PR #72, opened 2016-01-07
Implemented Str * Str

bsdtux:
Implemented str * str functionality


python/common/org/python/types/Str.java
```diff
@@ -287,6 +287,9 @@ public boolean __setattr_null(java.lang.String name, org.python.Object value) {
         else if (other instanceof org.python.types.Float) {
             throw new org.python.exceptions.TypeError("can't multiply sequence by non-int of type '" + other.typeName() + "'");
         }
+        else if (other instanceof org.python.types.Str) {
+            throw new org.python.exceptions.TypeError("can't multiply sequence by non-int of type '" + other.typeName() + "'");
+        }
```
freakboy3742: Is there any reason that this can't just be else { raise TypeError }? Are there any error messages here other than "can't multiply sequence by non-int of type ..."? This might mean we end up with a bunch of other passing multiply tests as a side effect.

tests/datatypes/test_str.py
```diff
@@ -235,7 +235,7 @@ class InplaceStrOperationTests(InplaceOperationTestCase, TranspileTestCase):
         'test_multiply_frozenset',
         'test_multiply_list',
         'test_multiply_set',
-        'test_multiply_str',
+        'test_multiple_str',
```
freakboy3742: This looks like it might be typo.

python/common/org/python/types/Str.java
bsdtux: Let me change that, and run all the test. If they pass I believe a else statement would cut down on number of statements needed.

freakboy3742:
A couple of comments inline. To clean up the commit history, read [this article](https://github.com/ginatrapani/todo.txt-android/wiki/Squash-All-Commits-Related-to-a-Single-Issue-into-a-Single-Commit) about squashing commits.

PR #77, opened 2016-01-28
Correct instructions in Getting Started and Tutorial-0

ankush981:
Minor corrections:
- In getting-started.rst, the `cd voc` should come immediately after the repo is cloned.
- In tutorial-0.rst, since we are inside `tutorial/tutorial0`, the `classpath` should be modified to `../dist` rather than `../voc/dist`.


docs/tutorials/tutorial-0.rst
```diff
@@ -55,7 +55,7 @@ This classfile can run on any Java 6 (or higher) VM. To run the project, type:
 
 .. code-block:: bash
 
-    $ java -classpath ../voc/dist/python-java.jar:. python.example.__init__
+    $ java -classpath ../dist/python-java.jar:. python.example.__init__
```
freakboy3742: The problem here isn't the classpath - it's the directory in which "mkdir tutorial0" is run. The instructions assume people would go back to the home directory (the directory with `voc` and `env` in it), which is an oversight on my part. The tutorial code shouldn't be dropped in the VOC install directory.

The better fix here is to a "cd .." to the end of the getting started instructions after `ant java`, and a note that this is the "tutorial home directory". Tutorial 0 should then start "If you're not already in the tutorial home directory (the same directory where `voc` and `env` are located), change into that directory".

PR #78, opened 2016-01-29
Improve instructions for Getting Started

ankush981:
Made two changes to improve the Getting Started part:
- Added the `env` directory to `.gitignore`
- Added `cd voc` to Getting Started as we need to be inside the repo directory.


freakboy3742:
Doesn't this patch put the ``cd`` in the wrong place? If you follow this patch, the virtual environment will be created *inside* the ``voc`` directory (which explains why you've had to add the virtual environment to the ``.gitignore`` file). Your previous patch had the ``cd`` after the ``virtualenv`` was created, and removed the cd from before the ``ant java`` call, which would result in that placement.

freakboy3742:
For the record - you don't need to close a pull request if you want to make a change. If you push a commit to the same branch that the pull request came from, the pull request will be updated with your changes. Don't think of the pull request as a patch - think of it as a ticket in a bug tracking system. You can continue to update the details on a single ticket until the project maintainer (me, in this case) is happy with the patch (the combined result of a series of commits), at which point I'll merge the pull request.

ankush981:
Ah, I see! But doesn't that kind of "taint" the commit history with stuff like `Changed file`, `Corrected errors`, `Again corrected errors`, `Corrected one error I missed out earlier`, etc?

freakboy3742:
Yes it does; however, that process can, in itself, be enlightening - for example, chasing down somewhere where you accidentally introduced a bug due to a search-and-replace failure. There's also a merge commit which covers all the little commits, so the "lots of little commits" problem isn't critical.

If you're still disturbed (or embarrassed :-) by the commit history, you can always squash the commits - if you use ``git reset HEAD^`` to unroll the commits, make a single unified commit, and the ``git push --force``, the pull request will be updated and will only show the single commit. The downside to this approach is that you're re-writing history, so any historical review comments will become detached, so it's generally frowned upon - but if what you're squashing is incidental, or reflects a fundamental change in the patch, it's usually OK.

ankush981:
Cool! :+1:

ankush981:
Makes sense. Yes, I'm aware of the dangers of changing history in a shared workflow. I guess I'll let things be. Once again, thanks for your patient explanations! :smile:

PR #80, opened 2016-01-30
Corrected Getting Started guide

ankush981:
Moved the `cd voc` part to its right location. I've always had the habit of creating my virtual environments right inside the project directory, and I realize I ended up doing that the last time. How I got the setup to work last time is still a mystery to me! 

In any case, we're now going to have parallel directories: `env`, `voc`, `tutorial-0`, `tutorial-1`, `tutorial-2`, etc.

I also added some minor formatting corrections.


ankush981:
Yipee! :dancer: :dancers: :+1: :smile: :smile:

PR #84, opened 2016-02-04
Implement test_unary_plus for Str

ankush981:
Added test for unary positive to the String class.


freakboy3742:
The CI failure appears to be part of a bigger problem - I've addressed that independently.

ankush981:
Thanks! . . .

Yay! Me is an open source contributor now! :dancer: :dancer:

PR #86, opened 2016-02-05
Wiki link on readme

ankush981:
Every time I have to struggle a lot to find the "contribution explanation" link on the wiki. It's buried two levels deep, and is neither in the README nor in Read The Docs. I'm adding it to README to make it easier for new people to figure out what to do.


freakboy3742:
A great suggestion! Longer term, we probably want to polish up the content on that wiki page and put it into the formal documentation - the wiki page is a great solution for right now, but I think we can do much better.

Issue #13, opened 2015-10-01
Use a single Java session for tests


freakboy3742:
The current test runner validates Java output by starting a fresh VM for every test. However, JVM warmup is time-expensive operation. 

Ideally, we'd be able to use a single JVM instance, and run each test in that instance, rather than restarting the JVM each time. 

If necessary, this could be done on a per-TestCase or per test Module level.

cflee:
I tried tweaking the TieredCompilation and TieredStopAtLevel options on the [Java HotSpot VM](http://docs.oracle.com/javase/8/docs/technotes/guides/vm/performance-enhancements-7.html) but a rough test showed it was at best equal, and usually worse than without those options.

I was also looking at [Drip](https://github.com/ninjudd/drip) as well, which looks very interesting, but I'm not sure if it's feasible to keep a sufficiently large pool of JVMs running such that we can use and dispose of one every 0.x seconds without ever having to wait for the pool to replenish.

On the other hand, after a bunch more work, I do have a very, very rough proof of concept working using [nailgun](https://github.com/martylamb/nailgun). Nailgun works by keeping a JVM running a server that listens to a socket for commands, and accompanying C client that issues commands to run a particular class. However, it doesn't seem to be able to run the transpiled classes directly, maybe because the classes in tests/temp/* don't exist when Nailgun is started, or because it keeps being replaced.

(There is a python port of the client, but I haven't tried that yet.)

 [Classpath-manager](https://github.com/stuartsierra/classpath-manager) tries to address this by having a class that reads a list of configurable classpaths, creates a new URLClassLoader and loads them, then spawns a new Thread, attaches the ClassLoader to it, then invokes the main() method of the desired class within the Thread. In other words, this acts as a layer of indirection so that the ever-changing freshly transpiled classes in tests/temp/* can get freshly loaded for each test. I did have to monkey-patch it to avoid using the smart ClassLoaderCache, as reusing the ClassLoader doesn't work.

The good news:

```
python3 setup.py test -s tests.datatypes.test_int.UnaryIntOperationTests
unchanged: 7.62s
nailgun: 1.761

python3 setup.py test -s tests.datatypes.test_int
unchanged: 723.179s
nailgun: 94.342s

python3 setup.py test
unchanged: <no result, never ran this on my machine>
nailgun: 610.148s
```

The bad news: 

[...]

PR #88, opened 2016-02-06
Implement unary `not` on String

ankush981:
The branch name is misleading - it's the implementation for `not` and not `invert`. But I guess branch names are minor issues? :v:

### DIFF
--- a/python/common/org/python/types/Str.java
+++ b/python/common/org/python/types/Str.java
@@ -334,8 +334,8 @@ public class Str extends org.python.types.Object {
     @org.python.Method(
         __doc__=""
     )
-    public org.python.Object __not__() {
-        return new org.python.types.Bool(this.value.length() == 0);
+    public org.python.Object __bool__() {
+        return new org.python.types.Bool(this.value.length() > 0);
     }
 
     @org.python.Method(
--- a/python/common/org/python/types/Str.java
+++ b/python/common/org/python/types/Str.java
@@ -330,6 +330,13 @@ public class Str extends org.python.types.Object {
     public org.python.Object __neg__() {
         throw new org.python.exceptions.TypeError("bad operand type for unary -: 'str'");
     }
+    
+    @org.python.Method(
+        __doc__=""
+    )
+    public org.python.Object __not__() {
+        return new org.python.types.Bool(this.value.length() == 0);
+    }
 
     @org.python.Method(
         __doc__ = ""
--- a/tests/datatypes/test_str.py
+++ b/tests/datatypes/test_str.py
@@ -20,7 +20,6 @@ class UnaryStrOperationTests(UnaryOperationTestCase, TranspileTestCase):
     values = ['""', '"This is a string"']
 
     not_implemented = [
-        'test_unary_not',
         'test_unary_invert',
     ]
 


freakboy3742:
You're correct that the branch name doesn't matter. 

While this change does make the test pass, I'm going to hold off merging for the moment, pending some further exploration. __not__ isn't the right place to be addressing this. If I were to subclass Str, and override __bool__, I would expect the behaviour of "not" to change as well - but if VOC uses the special method __not__, this won't happen.

So - in order to maintain compatibility with CPython, the fix really needs to be based in __bool__.

ankush981:
Agree on that. Please hold, then, and let me try to make it work with `__bool__`

ankush981:
Well, well. I was able to solve it without doing anything special. I thought I was supposed to provide the `not` functionality, but knowing that VOC does this hooking-up internally was what clicked. Now I've just provided a `__bool__()` method while VOC does the rest. The unit testing remains a mystery, though. Now it says "ok". How does it arrive at its answers?

freakboy3742:
Awesome - I told you that you would be able to sort it out :-)

To answer your other question: the unit testing is based on live testing actual code. The test case specifies a Python script. The test harness then starts a subprocess, and runs that script under CPython, capturing the output. This becomes the "expected" result. The harness then compiles the script into Java  bytecode using VOC, and invokes a second subprocess to generate the "actual" result. Some light post-processing is done to both the "expected" and "actual" results to normalise them (for example, to make sure floating point numbers and stack traces are formatted the same way), and then the two results are compared. If they match, the test passes. If they don't, the test fails, and the diff between expected and actual is displayed so you can work out what went wrong.

The weakness to this approach is that it won't catch when Java is producing the right output, for the wrong reason. For example, if the input code is "print(2 + 2)", and VOC converts this to "System.out.println("4")" instead of "System.out.println(2+2)", the test will still pass, even though it hasn't correctly interpreted the addition inside the print statement. However, this is a bit of an edge case, and hasn't bitten us so far.

The other downside is speed. The reason the full suite takes 40 minutes to run is because of the startup time of the Java VM.  ideally, we would start a single VM, and re-use it over and over again - but then you hit problems with caching inside the VM. Ticket #13 is tracking this problem; but it's not a simple one to fix.

On top of the test case itself, there's another layer that rolls out the n x m, "all operators on all data types" matrix of test cases. That's what you're interacting with when you delete the "test_str_not" line - the interface that determines which test cases in the matrix are known failures.

ankush981:
@freakboy3742 Awesome stuff! And thanks again for the detailed explanation. I still don't understand the "ok" output, though. What does the test harness mean by "ok"? I noticed that "ok" is output in those cases where the unit test is missing. Does that mean the harness assumes that "I can't test this so I'll assume all is well"?

freakboy3742:
"OK" means "the test ran and got the output that was expected." 

If, by "missing", you mean "not listed in the list of tests" - that's correct. The test harness for datatypes is dynamically generated - the code is in tests/utils.py. The "not_implemented" lists of strings containing tests names is the list of tests that are known failures because the code isn't implemented; the code that generates the test suite automatically applies the "expectedFailure" decorator to those tests. 

If a test name *doesn't* appear in that list, it's assumed that it's a normal test - and a normal test should always pass (i.e., give an OK output).

Issue #36, opened 2015-12-04
Complete implementation of operations on standard types


freakboy3742:
Java has a specific set of allowed operator and operations. Python has a set of operators and operations as well. However, the two don't match exactly. For example, Java will allow any object to be added to a `java.lang.String`; Python does not. On the other hand, Python allows multiplication of `str` by `int` (producing a duplicated string); Java does not. 

In order to replicate Python behavior in Java, Python's logic for all the basic operations needs to be implemented in the VOC Java support library.

The test suite contains around 2800 tests in the that are currently marked as "expected failures". These reflect an attempt to do "every operation between every base data type". The task: pick a data type, and write the implementation of one of the dunder operation (e.g., `__add__` or `__xor__`) or comparison (`__lt__` or `__eq__`) methods. 

The tests will pass (and become unexpected successes) when the output of Python code doing that operation is the same when run through CPython and VOC - and I mean byte-identical, down to the text and punctuation of the error message if appropriate.

If you want to see an example of what is involved, check out the implementation of [__add__](https://github.com/pybee/voc/blob/master/python/common/org/python/types/Int.java#L135) for integers. If you have an `int`, Python will allows you to add an `int` or a float` to it; all other types raise a `TypeError`. The list of operations on `int` that still need to be implemented can be [found here](https://github.com/pybee/voc/blob/master/tests/datatypes/test_int.py); if you add a new operation, delete the line that corresponds to that test, and the test will report as a pass, rather than an expected fail.

freakboy3742:
When you submit a PR for this, prefix the PR message with "Refs #34", so it will be associated with this ticket.

ankush981:
Folks, I'm picking up the `test_unary_positive` test in `tests/test_str.py`. Please don't work on it so that we can avoid collisions! :smile:

ankush981:
I forgot to mention, I'll be doing all the unary operator tests for String. Please stay away -- they're mine! :stuck_out_tongue_closed_eyes:

ankush981:
Hello, people. I'll be picking up the following tests for string types:

```
'test_eq_bytearray',
'test_eq_bytes',
'test_eq_class',
'test_eq_complex',
'test_eq_frozenset',
```

cflee:
I'm working through the Int class' unary and binary operations.

mzwaen:
Hi! I'm working on the data type Boolean, starting with comparison methods.

khoobks:
I'm working through the List class

[...]

PR #92, opened 2016-03-10
Implement some Int operations

cflee:
Implemented the three remaining tests for the unary operators (`+`, `-`, `~`) and some for binary `&`.

I'm not sure what to do with the remaining tests for `&`, I can't seem to run them:
    AttributeError: type object 'BinaryIntOperationTests' has no attribute 'test_and_bytearray'
    RuntimeError: Unknown constant type <class 'bytes'>

Refs #36


freakboy3742:
Fantastic work! Thanks! 

There's one more small change required: your PR causes one test to be an unexpected pass, which is flagged by the CI build: `tests.datatypes.test_int.InplaceIntOperationTests.test_add_bool` (Note - it's the *inplace* tests, not the *binary* test). If you remove the line that puts that test in the `not implemented` list, the PR should be ready for merge.

Regarding the problems with bytearray test - I think you'll be able to see the source of the problem if you look at the source code for `types/ByteArray.java` itself :-) If you're looking for your next contribution, filling that hole would be a great place to start.

cflee:
Right - I have to admit that I was waiting for Travis CI to run the full set of tests since it takes so long.. I will remove that line.

Some questions though:

I am working my way through some of the other binary operators, but I was a bit mystified about how some particular tests are passing. Let's take `tests.datatypes.test_int.BinaryIntOperationTests.test_and_bool` for example. The Java implementation is hard-coded to return int 0, which is wrong, but it's not caught as the values in test_int.py don't include an odd number. Should I uncomment the '-37' so that it also has a negative, odd number like in the `SAMPLE_DATA` in utils.py? That makes the `test_and_bool` test fail as expected.

Also, do you have a preference for whether the class names are fully specified in the Java source files? For instance in /python/common/org/python/types/Int.java, sometimes the `instanceof` parameter is `org.python.types.Int`, sometimes it's just `Float`, which actually works perfectly fine and is easier to read.

freakboy3742:
The speed of the test suite is definitely a concern for me. It's the subject of issue #13; but it isn't an easy fix (at least, I haven't been able to think of one). I'm open to any suggestions on ways to make the testing faster.

Q1: I certainly won't warrant that the test framework itself is bug free. The problem you describe, and the fix you've proposed both make sense to me.

Q2: I completely agree that short names are easier to read... right up until there's a name collision and you find things failing because it's referencing the builtin type with a given name, rather than the Python type. I've been bitten by that a couple of times so far, which has led to me erring on the side of "explicit"  full naming of classes. This avoids any confusion, even though it is a lot more long-winded.

Again - I don't doubt that I've missed a few places, though. I don't think I'd consider that a "bug", per se; more a thing that I'd probably clean up if I noticed it and I was in the area.

cflee:
Yes, I was thinking about avoiding the JVM startup and it certainly looks non-trivial. Also, I entirely missed that some of the package class names do overlap with the native boxed ones - good point.

I uncommented the '-37', and it raises this interesting failure, amongst others:

```
FAIL: test_power_float (tests.datatypes.test_int.BinaryIntOperationTests)
  File "/Users/cflee/Desktop/voc/tests/utils.py", line 499, in func
    self.assertBinaryOperation(x=value, y=example, operation=operation, format=self.format)
  File "/Users/cflee/Desktop/voc/tests/utils.py", line 520, in assertBinaryOperation
    """ % kwargs, "Error running %(operation)s with x=%(x)s and y=%(y)s" % kwargs)
  File "/Users/cflee/Desktop/voc/tests/utils.py", line 326, in assertCodeExecution
AssertionError: 'NaN\n' != '(2223.33566194017+4218.35887675393j)\n'
- NaN
+ (2223.33566194017+4218.35887675393j)
 : Global context: Error running x ** y with x=-37 and y=2.3456

```

It seems that negative numbers raised to a non-integer exponent yield a complex number. As Complex is not implemented yet, I've added `test_power_float` back to `not_implemented` instead of removing `-37` again, since that should be more useful for flushing out bugs.

Apart from that, the binary operations should be covered. I tried looking at the in-place operations, but the remaining ones seem non-trivial - for instance, supporting `+= 5.5` on an Int, as then the Int needs to change into a Float.

It seems like that would require the Int to encapsulate a Float, and then either the Int or the parent Object has to passthrough method calls to the Float if it's set. If there are other cases of these types changing, getting the Object class to determine the 'actual' object to call would be cleaner. Nonetheless, that sounds messy; I'm not sure if there's a better way.

(I'm much more familiar with Java than with Python, so please excuse me if I'm not so conversant with Python fundamentals. :grin:)

freakboy3742:
All looks good - thanks! Your instinct with adding test_power_float back to the "not_implemented" list looks right to me. 

However, I'm a little surprised by the error you're seeing. A negative number raised to a float power shouldn't be generating a complex number, so there must be something else going on. If you do the same operation (``-37**2.345``) at the Python prompt, you definitely don't get a complex number, so I don't know what is going on there.

Regarding the changes required by inplace - that's a good point that I hadn't considered; more thought is definitely required. My immediate reaction is that it might be necessary to tackle that one a little higher up, in the of the INPLACE_ADD python bytecode. At that point, we'll be in a position to do the int->float conversion if necessary. 

On a separate note: this set of commits definitely qualifies you for a BeeWare challenge coin. Do you attend any Python conferences? If so, I can give you one in person; otherwise, I'll have to mail you one.

cflee:
> However, I'm a little surprised by the error you're seeing. A negative number raised to a float power shouldn't be generating a complex number, so there must be something else going on. If you do the same operation (-37**2.345) at the Python prompt, you definitely don't get a complex number, so I don't know what is going on there.

That was nearly the reaction I posted in my earlier comment, but after very long time poking through the source and the interactive shell, it seems to be an [operator precedence](https://docs.python.org/3/reference/expressions.html#operator-precedence) issue: 

```
>>> x = -37
>>> y = 2.3456
>>> x ** y
(2223.33566194017+4218.35887675393j)
>>> -37 ** 2.3456
-4768.414105207686
>>> (-37) ** 2.3456
(2223.33566194017+4218.35887675393j)
>>> (-37) ** 2     
1369
```

Also, the [Python 3 reference](https://docs.python.org/3/reference/expressions.html#the-power-operator) notes: "Raising a negative number to a fractional power results in a complex number. (In earlier versions it raised a ValueError.)" It was this that prompted me to check the precedence rules since that's the only explanation left.

> Regarding the changes required by inplace - that's a good point that I hadn't considered; more thought is definitely required. My immediate reaction is that it might be necessary to tackle that one a little higher up, in the of the INPLACE_ADD python bytecode. At that point, we'll be in a position to do the int->float conversion if necessary.

Cool, this would make a lot more sense – I'm still trying to spelunk around the source to figure out how this whole thing works, so I was looking at it from a purely Java perspective.

> On a separate note: this set of commits definitely qualifies you for a BeeWare challenge coin. Do you attend any Python conferences? If so, I can give you one in person; otherwise, I'll have to mail you one.

That's awesome! Not really, my experience largely revolves around web apps in PHP and Java, with a lot more Ruby and JavaScript lately, so I've really not involved in Python... I'll ping you via email!

freakboy3742:
Ah - that makes sense - good catch. I guess the fix here is to stick brackets through the test program so we can guarantee that operator precedence in the test is actually what we want to test.

freakboy3742:
FYI - I've just added parentheses to the test suite so the operator order should be unambiguous.

freakboy3742:
... or, on second thought - I've just reverted that change. The problem was with the manual testing, not the test suite. Note to self: don't code while distracted :-)

cflee:
Yup, I was just about to comment - the test suite uses the x and y variables instead, so there's no impact from doing `print(x ** y)` in the test case. :grin:

PR #95, opened 2016-03-19
Windows 10 Compatibility

khoobks:
Changes to support running the unit tests on Windows.

- Symlinks not supported on Windows
- Directory separators ('\' vs '/')
- Path separators (';' vs ':')
- Line endings ('\r\n' vs '\n')
- Slight variations in the format of exceptions

The changes were 'tested' on the BinaryStrOperationTests with the following result
Ran 171 tests in 276.103s
OK (expected failures=32)

### DIFF
--- a/build-java.xml
+++ b/build-java.xml
@@ -25,9 +25,9 @@
 
     <jar jarfile="${dist}/Python-Java-${python-version}-b${release}.jar" basedir="${build}/java"/>
 
-    <symlink
-        link="${dist}/python-java.jar"
-        resource="${dist}/Python-Java-${python-version}-b${release}.jar"
+    <copy
+        tofile="${dist}/python-java.jar"
+        file="${dist}/Python-Java-${python-version}-b${release}.jar"
         overwrite="true" />
   </target>
 
--- a/tests/utils.py
+++ b/tests/utils.py
@@ -34,6 +34,7 @@ def setUpSuite():
         stdin=subprocess.PIPE,
         stdout=subprocess.PIPE,
         stderr=subprocess.STDOUT,
+        shell=True,
     )
 
     try:
@@ -147,8 +148,13 @@ def runAsJava(test_dir, main_code, extra_code=None, run_in_function=False, args=
     if args is None:
         args = []
 
+    classpath = os.pathsep.join([
+        os.path.join('..', '..', 'dist', 'python-java.jar'),
+        os.path.join('..', 'java'),
+        os.curdir,
+    ])
     proc = subprocess.Popen(
-        ["java", "-classpath", "../../dist/python-java.jar:../java:.", "python.test.__init__"] + args,
+        ["java", "-classpath", classpath, "python.test.__init__"] + args,
         stdin=subprocess.PIPE,
         stdout=subprocess.PIPE,
         stderr=subprocess.STDOUT,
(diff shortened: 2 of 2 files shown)


khoobks:
Many failing tests. Closing this Pull Requests until I can further investigate the issue.

cflee:
It's the `shell=True` change to the "ant java" bit in `setUpSuite()` that is causing the breakage on Linux/OS X – the behaviour is a [little different](https://docs.python.org/3.4/library/subprocess.html#popen-constructor) with that enabled, so the "java" parameter goes to the shell instead of to ant. This makes ant try to build the android target as well.

It works if it was changed to `"ant java"` or `["ant java"]` instead, but now I'm curious why that's necessary only for ant and not for the other subprocess calls in `compileJava()`, `runAsPython()`, `runAsJava()`, etc.

khoobks:
Oh fantastic! Thanks for that tip cflee. 

Without a Linux/OSX machine handy that would have been difficult to track down.

cflee:
Yeah, I am on OS X so I figured it might be faster for me to fetch and test, since I've been poking at the test suite for #13. However, there are _lots_ of failures even after that fix, I suspect the regex needs some work. Here is an example:

```
FAIL: test_add_dict (tests.datatypes.test_int.BinaryIntOperationTests)
  File "/Users/cflee/Desktop/voc-dev/voc/tests/utils.py", line 511, in func
    self.assertBinaryOperation(x=value, y=example, operation=operation, format=self.format)
  File "/Users/cflee/Desktop/voc-dev/voc/tests/utils.py", line 532, in assertBinaryOperation
    """ % kwargs, "Error running %(operation)s with x=%(x)s and y=%(y)s" % kwargs)
  File "/Users/cflee/Desktop/voc-dev/voc/tests/utils.py", line 338, in assertCodeExecution
AssertionError: 'Exception in thread "main" java.lang.Exc[138 chars]:3\n' != "### EXCEPTION ###\nTypeError: unsupporte[54 chars]:3\n"
- Exception in thread "main" java.lang.ExceptionInInitializerError
+ ### EXCEPTION ###
- Caused by: org.python.exceptions.TypeError: unsupported operand type(s) for +: 'int' and 'dict'
? ---------------------------------
+ TypeError: unsupported operand type(s) for +: 'int' and 'dict'
      test.py:3
 : Global context: Error running x + y with x=42 and y={}
```

Here is what I get from java_out (I assume with `\n` newlines):

```
Exception in thread "main" java.lang.ExceptionInInitializerError
Caused by: org.python.exceptions.TypeError: unsupported operand type(s) for +: 'int' and 'dict'
	at org.python.types.Int.__add__(Int.java:192)
	at python.test.__init__.<clinit>(test.py:3)

```

You can try running `tests.test_utils.JavaNormalizationTests`, I think those contain output which is obtained on Linux. Which reminds me, I need to add some tests as well.

khoobks:
Yeah I saw those. I think I have a fix for it but this time I'll run the full set of tests instead of being lazy and just running one test suite.

khoobks:
I tried the ["ant java"] option you have above but no luck on Windows. The "ant java" way works though.

freakboy3742:
Thanks for the helpful pointers @cflee. For the record, I'm open to any suggestions of a better way to normalise Java and Python stack trace output - regexes were the best way I could think of, but I'm fully aware that they can be fragile, and when they break, they tend to break hard.

PR #98, opened 2016-03-20
Implement TestDaemon - single JVM instance for tests

### LINKED ISSUE #13: Use a single Java session for tests

cflee:
This fixes #13, but is likely to conflict with #96 as both pull requests changed the regex and subprocess stuff in utils.py.

- Implement a TestDaemon Java class that accepts class names on stdin, and loads + runs the required classes, then sends the output on stdout/stderr. The voc classes are kept loaded to improve performance.
- Add TestDaemon to ant build for `ant java` target.
- Keep a fallback to individual JVM per test when `args` is non-empty, as `sys.argv` will be wrong since it picks up the TestDaemon's parameters from the JVM instead of the entry main() method's `String[] args` parameter.
- Fix an inconsistent test implementation in `test_interface`, as provided by @khoobks in #96 at [d8cfe6d6](https://github.com/pybee/voc/pull/96/commits/d8cfe6d6cdda1b5f482b597255f426e24b4b07b9). No idea how this passes!
- Current test suite runtime is improved to 9-11 min on Travis CI, about 5x speedup.

Potential issues:
- Missing tests for adjusted `JAVA_EXCEPTION` regexes. I'm not sure how to generate the `<clinit>` vs non-`<clinit>` kind of exceptions to use for the test.
- Fallback to individual JVM may have a large impact on performance on test runtime if there are many tests that provide the args. However, this seems non-fixable unless there's an alternate way to implement `sys.argv`.
- I have observed this issue where running the tests with no built files causes `ant java` to run to compile and build the jars, but then the java invocation for TestDaemon fails as it can't find the files / can't find the TestDaemon class. Restarting the test runner fixes this. Not sure if it's reproducible for others.
  - `ant clean && python3 setup.py tests` triggers this on OS X.

### DIFF
--- a/python/common/org/python/ImportLib.java
+++ b/python/common/org/python/ImportLib.java
@@ -79,7 +79,7 @@ public class ImportLib {
                 if (!name.equals("*")) {
                     try {
                         if (native_import) {
-                            java.lang.Class java_class = java.lang.Class.forName(java_name.toString().replace("/", ".") + name);
+                            java.lang.Class java_class = java.lang.Thread.currentThread().getContextClassLoader().loadClass(java_name.toString().replace("/", ".") + name);
                             parent_module.__setattr__(name, org.python.java.Type.pythonType(java_class));
                         } else {
                             python_module = importPythonModule(java_name.toString() + name);
@@ -106,7 +106,7 @@ public class ImportLib {
             throws java.lang.ClassNotFoundException {
         org.python.types.Module python_module;
         try {
-            java.lang.Class java_class = java.lang.Class.forName(java_name.replace("/", "."));
+            java.lang.Class java_class = java.lang.Thread.currentThread().getContextClassLoader().loadClass(java_name.replace("/", "."));
             python_module = null;
         } catch (java.lang.ClassNotFoundException e) {
             python_module = new org.python.java.Module(java_name.replace("/", "."));
@@ -136,7 +136,7 @@ public class ImportLib {
             throws java.lang.ClassNotFoundException {
         org.python.types.Module python_module;
         try {
-            java.lang.Class java_class = java.lang.Class.forName(java_name.replace("/", ".") + ".__init__");
+            java.lang.Class java_class = java.lang.Thread.currentThread().getContextClassLoader().loadClass(java_name.replace("/", ".") + ".__init__");
             java.lang.reflect.Constructor constructor = java_class.getConstructor();
             python_module = (org.python.types.Module) constructor.newInstance();
             modules.put(java_name, python_module);
--- a/tests/utils.py
+++ b/tests/utils.py
@@ -18,7 +18,6 @@ from voc.transpiler import Transpiler
 
 # A state variable to determine if the test environment has been configured.
 _suite_configured = False
-_jvm_configured = False
 
 def setUpSuite():
     """Configure the entire test suite.
@@ -283,7 +282,6 @@ class TranspileTestCase(TestCase):
 
     def setUpClass():
         global _jvm
-        global _jvm_configured
         test_dir = os.path.join(os.path.dirname(__file__))
         _jvm = subprocess.Popen(
             ["java", "-classpath", "../dist/python-java-testdaemon.jar", "python.testdaemon.TestDaemon"],
@@ -292,15 +290,12 @@ class TranspileTestCase(TestCase):
             stderr=subprocess.STDOUT,
             cwd=test_dir,
         )
-        _jvm_configured = True
 
     def tearDownClass():
         global _jvm
-        global _jvm_configured
-        if _jvm_configured:
+        if _jvm is None:
             # use communicate here to wait for process to exit
             _jvm.communicate("exit".encode())
-            _jvm_configured = False
 
     def assertBlock(self, python, java):
         self.maxDiff = None
--- a/tests/utils.py
+++ b/tests/utils.py
@@ -217,8 +217,8 @@ def compileJava(java_dir, java):
 
 
 JAVA_EXCEPTION = re.compile(
-    '(((Exception in thread "\w+" )?org\.python\.exceptions\.(?P<exception1>[\w]+): (?P<message1>[^\r?\n]+))|' +
-    '((Exception in thread "[\w-]+" )?([^\r?\n]+\r?\n)+' +
+    '(((Exception in thread "[\w-]+" )?org\.python\.exceptions\.(?P<exception1>[\w]+): (?P<message1>[^\r?\n]+))|' +
+    '([^\r\n]*?\r?\n((    |\t)at[^\r\n]*?\r?\n)*' +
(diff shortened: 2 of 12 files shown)


python/common/org/python/ImportLib.java
```diff
@@ -79,7 +79,7 @@
                 if (!name.equals("*")) {
                     try {
                         if (native_import) {
-                            java.lang.Class java_class = java.lang.Class.forName(java_name.toString().replace("/", ".") + name);
+                            java.lang.Class java_class = java.lang.Thread.currentThread().getContextClassLoader().loadClass(java_name.toString().replace("/", ".") + name);
```
freakboy3742: I take it that this is required to make sure that, during testing, the non-caching class loader is in use? Does this effectively mean that any use of `Class.forName()` will fail under test?

tests/utils.py
```diff
@@ -254,6 +270,27 @@ class TranspileTestCase(TestCase):
     def setUp(self):
         setUpSuite()
 
+    def setUpClass():
+        global _jvm
+        global _jvm_configured
```
freakboy3742: Do we need both `_jvm` and `_jvm_configured`? Couldn't we just use `_jvm is None` as as marker for "we have no JVM", and when you invoke `runAsJava()`, start a new JVM if one doesn't exist? That would also avoid duplicating the `Popen` logic.

tests/utils.py
```diff
@@ -254,6 +270,27 @@ class TranspileTestCase(TestCase):
...
+            cwd=test_dir,
+        )
+        _jvm_configured = True
+
+    def tearDownClass():
+        global _jvm
+        global _jvm_configured
+        if _jvm_configured:
+            # use communicate here to wait for process to exit
+            _jvm.communicate("exit".encode())
```
freakboy3742: Shouldn't this be `encode('utf-8')`?

freakboy3742:
This looks great - and the speedup is amazing. 

I've made minor review comments; the only one that's really concerning to me is the need to avoid `Class.forName()`, because it's such a common idiom; I just want to make sure I understand the consequences of that change. 

I did a merge with the current trunk; but I got a couple of test failures - they all appear to be due to a buffer flushing issue. A test that outputs a couple of lines out stdout, then raises an exception seems to match the exception fine, but the output is missing. `DictTests.test_getitem`, `RangeTests.test_creation` and `RangeTests.test_step` are all examples - there's about a dozen all up that are failing for me. I'm not sure if this is due to an error in my merge, or a problem that is only appearing on my machine.

cflee:
The need to avoid `Class.forName()` only affects code that goes into the runtime jar; the generated python.* class bytecode continues to use `Class.forName()` just fine.

What happens is that there's no way to 'unload' or not-cache classes in the ClassLoader, as by design they don't reload classes that have been loaded before. This makes it necessary to create a new ClassLoader for each test, to allow it to pick up the newly generated `python.test.__init__` class for the test.

However, that only gets the test runtime down to 35 min or so. Hence, for performance reasons we cache the loaded classes that are from python-java.jar. These classes are marked with the ClassLoader from the first test.

When we move onto the second test, the second ClassLoader loads the new `python.test.__init__` fine. However, later when `org.python.ImportLib.__import__()` is invoked, the ImportLib class used is the cached one from the first ClassLoader. The `__import__()` method (or the other methods in the class) will try to do `Class.forName("python.test.__init__")`, but by default it will gets the classes from the ClassLoader that created it, so that's the first ClassLoader. This yields the wrong class, as the first ClassLoader has an "old copy" of the class, that doesn't match the .class file on-disk right now.

Bottom line: any use of `Class.forName()` in code that goes into the runtime python-java.jar will cause failure during tests, when the underlying .class file is changed (compileJava) within the same test class, on the second time that a class is loaded.

cflee:
Rebased onto latest master with force push onto this PR and made the following changes:
- Remove redundant `_jvm_configured` global
- Specify 'utf-8' when encoding strings, as mentioned above

I see the 12 test failures locally as well, will look into them.

cflee:
Looks like my changes to the regex to handle the additional lines in the stack trace between exception name and 'Caused by:' were assuming that the `Exception in thread "main"` bit is present. Adjusted the regex to handle that.

freakboy3742:
I've just merged, on the basis that a 5 minute test run is much better than a 45 minute test run :-)

freakboy3742:
I'd still like to look a bit more at the `Class.forName()` issue, though. If I'm understanding you correctly, if we could get `org.python.ImportLib` and `org.python.types.Type` to be loaded from a non-caching Class loader, the problem wouldn't exist, because by default, if class A tries to load class B, the class loader that loaded A will be used to load B. 

To be clear: I have no idea how we'd actually make that change - just verifying if my understanding is correct. The best idea I've got is a subclass of URLClassLoader that explicitly refuses to load `ImportLib` or `Type` at all, and use *that* as the loader for the runtime classes.

My concern is that the problem we're hitting here is an edge case caused by our test environment, but it's fundamentally affecting the way that we can write production code. There's no problem writing code to make it more testable, but if the test environment is imposing odd requirements on production code, it's something that is likely to cause confusion in future. 

However, If avoiding `Class.forName()` it's the only way to do it, that's fine - I'd just like to make sure it's the last resort. Alternatively, if we can come up with a good generic explanation why production code shouldn't be using `Java.getClass()` I'd be happy, because then we can provide that as general advice for potential VOC module implementors. Does that make sense?

cflee:
I definitely see where you're coming from - I was very reluctant to touch the existing codebase outside of the test suite, but that was the best solution I have right now.

I did try to not-cache `org.python.ImportLib` and some of the other classes earlier (since rewritten out of git history), by always loading it fresh from the new ClassLoader, but there were a slew of strange errors from NameError, AttributeError to particularly alarming IllegalAccessErrors. I tried digging into that a little bit, but after that I figured that the current workaround was the lowest-resistance path.

> I'd still like to look a bit more at the Class.forName() issue, though. If I'm understanding you correctly, if we could get org.python.ImportLib and org.python.types.Type to be loaded from a non-caching Class loader, the problem wouldn't exist, because by default, if class A tries to load class B, the class loader that loaded A will be used to load B.

I'm still spelunking through how the `<clinit>` stuff in the generated classes interact with the runtime jar,  at least for the small few-line test cases, so I can't definitively say if this is sufficient. Shall we track this in a new issue? I think this PR thread will get lost since it's been closed.

freakboy3742:
Ok - that makes sense. The patch you've provided will certainly suffice for now - I'm already loving the newfound speed :-) 

Opening a new ticket for "improving the test runner" is probably worthwhile, rather than piling on this closed ticket.

cflee:
I've opened issue #102, in case anyone else tries to follow this thread in future.

PR #99, opened 2016-03-21
Implementing __add__ for booleans

mzwaen:


freakboy3742:
Thanks for the contribution! You've got the right idea - the patch just needs a couple of small tweaks.

If you look at the continuous integration results above (click on the "details" link next to the "Travis CI build failed" message above), you'll see that there are 13 unexpected successes. That's because you've modified the definition of `__add__`, but you haven't updated the test suite to reflect those changes. As a result, the tests for `__add__` on a boolean are passing, where previously they were failing.

To fix this, you'll need to edit `tests/datatypes/test_bool.py`. That file contains the definitions for tests to be performed on a boolean. The long "not_implemented" lists all the tests that are expected to fail because the operators aren't implemented yet. Find the entries that correspond to the unexpected successes in the CI run, delete them, commit that change, and push to Github. That should generate a fresh CI build which will let you check if you've done the right thing.

If any of this doesn't make sense, let me know and I'll try to explain better.

mzwaen:
Hi! Thanks for the explanation! All checks seem to have passed now. In this case, do we need to make a new pull request or is the original pull request automatically updated with the changes?

freakboy3742:
No new pull request required - the original is automatically updated.

Congratulations! You're now a contributor to an Open Source project! Thanks for contributing!

freakboy3742:
If you've got a Twitter handle, I'd love to give you credit on the BeeWare Twitter feed.

mzwaen:
Hi,
I am not particularly active on twitter (to say the least:-) but here is my handle : @m_zwae
Thanks!

PR #103, opened 2016-03-21
Fix race condition between compile and testdaemon startup

cflee:
- As `setUpClass()` is runs before `setUp()`, ensure that the defined
`setUpSuite()` which runs `ant java` to build the jars is run before
testdaemon is started up.
- This was not an issue before the introduction of testdaemon because
only `setUp()` was defined.
- This should fix some spurious build pass/fail results due to the old
python-java.jar being used on testdaemon startup, before ant completes
the build.

:flushed: :sweat: :sweat_smile:

### DIFF
--- a/tests/utils.py
+++ b/tests/utils.py
@@ -277,10 +277,8 @@ def cleanse_python(input):
 
 
 class TranspileTestCase(TestCase):
-    def setUp(self):
-        setUpSuite()
-
     def setUpClass():
+        setUpSuite()
         global _jvm
         test_dir = os.path.join(os.path.dirname(__file__))
         _jvm = subprocess.Popen(


cflee:
Okay, this is very strange, the same commit [passes](https://travis-ci.org/cflee/voc/builds/117470760) on my repo but fails here due to `test_exit_from_static (tests.stdlib.test_sys.SysModuleTests) ... unexpected success`. I am restarting it on my repo's Travis to see if it passes then. If it persists and remains isolated to that test, I'm tempted to add a `@skip` to avoid this particular test messing with everyone.

freakboy3742:
@cflee I'm getting the same failure on my machine (OS/X). The problem is that the test is *already* skipped - the problem is an unexpected success, not a failure.

cflee:
I can't seem to replicate this on my OS X machine though, even with a fresh git clone. It also keeps passing on Travis on my fork, which is really annoying.

The only thing that looks suspicious to me is [TestDaemon.java:78](https://github.com/pybee/voc/blob/402c841f0cad464596f328e88eae4699f67c6280/python/testdaemon/python/testdaemon/TestDaemon.java#L78), that println() is probably redundant and possibly harmful, but removing it hasn't made any impact on my test runs.

I guess what I can do is to (i) break the exit handling, let testdaemon exit, and add restart handling to utils.py, (ii) add unused args to that test to make it fallback to individual JVM, (iii) use `@skip` instead of `@expectedFailure` to entirely suppress it for now, or (iv) try to fix the test to make it pass consistently.

PR #104, opened 2016-03-22
Implement static sys exit

cflee:
I don't actually know if this is a correct implementation, but it makes `test_exit_from_static` pass!

I copied the try-catch-endtry opcodes from the main method to the static block part, in order to catch the `org.python.exceptions.SystemExit` exception thrown by `sys.exit()`. Testdaemon also now handles the corresponding `System.exit()`/`ExitException` being thrown from the init block too.


freakboy3742:
I tried this approach back when I first wrote the `test_exit_from_static` method - while it definitely makes the test pass as written, I'm pretty sure there are some ordering problems with more complex projects - in particular, when there is a main block and static initialiser, the order of execution is different in Python than to Java, and that can lead to the exception being handled differently in the two platforms.

cflee:
Right. I did not realise there are differences there - closing this.

PR #106, opened 2016-03-22
Partial implementation of the List type

khoobks:
Partial implementation of List type operations.


cflee:
Phew! I thought it was a `test_exit_from_static` failure, but it's not for once!

Since I am looking at this anyway, just something minor - I don't know how strict the standards are for this project but usually in Java there are spaces around the `=` and `<`:
```diff
- for(int i=0; i<count; i++) {
+ for(int i = 0; i < count; i++) {
```

And a space after a cast:
```diff
- (org.python.types.Bool)this.value.get(i).__le__(otherList.value.get(i));
+ (org.python.types.Bool) this.value.get(i).__le__(otherList.value.get(i));
```

tests/datatypes/test_list.py
```diff
@@ -58,10 +58,6 @@ class UnaryListOperationTests(UnaryOperationTestCase, TranspileTestCase):
     values = ['[]', '[1, 2, 3, 4, 5]']
 
     not_implemented = [
-        'test_unary_positive',
-        'test_unary_negative',
-        'test_unary_not',
-        'test_unary_invert',
     ]
```
freakboy3742: If you remove the last of the non-implemented tests, you should be able to delete the not_implemented marker entirely (and if you can't, you should, and it's a bug). Once we've got a completely compliant implementation, there's no need for cruft to hang around telling us that there's nothing that isn't implemented.

freakboy3742:
What @cflee said - plus, spaces after for loops, too - `for` isn't a function call, so `for (...)`, not `for(...)` :-)

khoobks:
Thanks for the feedback.

I'll incorporate the changes over the next few days.

khoobks:
Corrected styling issues and fixed some classpaths for windows.


Refs #36

PR #107, opened 2016-03-24
Test other types against NoneType

cflee:
Implementation of NoneType has been added previously to pass the tests for NoneType against other types, this adds tests in the other direction:
- Modify `utils.py` to include NoneType in the binary/inplace operation tests
- Update the list of tests that are unimplemented
- Fix binary and inplace str-modulo-NoneType error messages to make them pass their tests


cflee:
Oops, forgot to quote the `'None'` in the values, will update this.

PR #109, opened 2016-03-24
Implement methods of Float

soummyaah:
Refs #36


freakboy3742:
@soummyaah Thanks for the contribution - your implementation looks great!. However, if you check the CI results (click on the red cross next to the most recent commit), you can see that there are 6 "unexpected success" results - these are tests that now pass, but are still listed in the "not_implemented" list. From the look of it, they're all `__le__` tests with datatypes that can't be compared. Delete those lines from the "not_implemented" list, push the update to GitHub, and your patch should be ready for merge!

soummyaah:
@freakboy3742 
Yeah, I was not sure what you wanted me to do with that, so I had left it until I received your response. 
I've removed them for now. But I'll be implementing all the methods for float in a day or two. So, I can either send a new PR for that and you can merge this one, or I can keep working on this PR itself. Whatever suits you.

freakboy3742:
@soummyaah Awesome! I'll hold off merging for now - let me know when you're done and I'll pull it in.

soummyaah:
@freakboy3742 
Hey, I was implementing __ne__ for Float and was wondering if I should implement it like this:
```
public org.python.Object __ne__(org.python.Object other) {
        return (!__eq__(other)); //Or whatever way necessary to negate the bool.
    }
```
Would this be preferred in place of basically copying from the __eq__ method and changing the operator?
This would keep the __eq__ and __ne__ tightly bound.
Also, I was trying to run the tests locally, however it was throwing this error: 
` 'module' object has no attribute 'test_codecs' `
Could you direct me to a way to solve this?

soummyaah:
@freakboy3742  
Hey, could you help me out with my query above?
Thanks,

freakboy3742:
@soummyaah Apologies for the delay in responding - I've been travelling for DjangoCon Europe, and I've only just returned home.

Having `__ne__` defer to `__eq__` seems like a good approach to me - that ensures that the two methods remain compatible. Your patch looks great, so I've just merged it.

As for the error you're seeing - that looks odd. Can you:
 * provide a full log of the output when you run the tests (including the command you're using to invoke the tests)
 * confirm that you have a `tests/test_codecs.py` in your checkout
 * confirm that you can successfully import the `test_codecs` file from a Python prompt.

PR #111, opened 2016-03-25
Dict type impl

khoobks:
Implemented the Dict Type.

I had to add an `isHashable` function in order to implement `__setitem__` and `__getitem__`

I originally added a `Hashable` interface to closely follow the Python specification for hashability - the presence or absence of a `__hash__` function - but ran into difficulties when trying to apply it to `Super`.


freakboy3742:
Can you elaborate on the problem with super? A Hashable interface would make more sense, assuming there isn't a fundamental underlying problem.

freakboy3742:
Can you elaborate on the problem with super? A Hashable interface would make more sense, assuming there isn't a fundamental underlying problem.

freakboy3742:
Can you elaborate on the problem with super? A Hashable interface would make more sense, assuming there isn't a fundamental underlying problem.

python/common/org/python/types/Dict.java
```diff
@@ -6,7 +6,7 @@
 
     /**
      * A utility method to update the internal value of this object.
-     *
+     * <p>
```
freakboy3742: You've done this in a couple of patches - is there a reason for it?

khoobks:
Well the problem I ran into is that the instance of `Super` needs to have a `__hash__` function if the class that it wraps is hashable otherwise it should not have a `__hash__` function. I don't know of a way to dynamically making a class implement an interface at runtime. Hence the `isHashable` function.

Looking at the code again, I have introduced a bug because `Super.isHashable` is infinitely recursive. This also hides a further problem that the code in `Super` needs to be able to work out if something is hashable purely from the `Class` definition since `Super` may not have an `instance` set.

OK so back to the drawing board.
- Static method or constant won't work because a subclass cannot override it
- `Hashable` interface; best candidate so far except that all `org.python.Object` instances will still have a `__hash__` function (see problem with `Super` above). The contract will that you can only validly call the `__hash__` function if the class also implements `Hashable` otherwise it is a meaningless implementation detail.

khoobks:
With the `<p>` tag, my IDE (IntelliJ IDEA 2016.1) is doing it automatically. You need to use them if you want to have an empty line in the generated javadoc.

http://stackoverflow.com/questions/21886533/p-tag-in-javadoc-in-android-studio

I am not sure why it is just inserting `<p>` instead of `<p/>`.

cflee:
I don't really know how `Super` affects this, but could you invoke `__hash__()` on the item and don't catch the `TypeError: unhashable type: 'Widget'` that may be thrown?

freakboy3742:
Is there any reason you couldn't put `__hash__` on the interface for Object, and then have it raise `AttributeError` if it's accessed on a class where it shouldn't exist? This should satisfy the requirements for Java, but would retain the dynamic behaviour needed for Python.

khoobks:
Pretty much did what freakboy3742 said in the comment above.

Issue #76, opened 2016-01-21
Complete implementation of Python builtins


freakboy3742:
`org.Python` is a Java implementation of all the Python builtins - methods like `print()`, `input()`, `list()`, `abs()`, `min()` and so on.

An implementation of these builtins is required.

freakboy3742:
See also [Ouroborous](https://github.com/pybee/ouroborous), a related project to develop a standalone, pure python version of the standard library. Ideally, it should be possible to use Ouroboros to generate the builtins and standard library.

cflee:
Where should the tests for these go? I was looking at adding tests for `abs(x)`, `bool(x)` for Int but I'm not sure if that should go into `test_int.IntTests`, or go into a new type of TestCase in `utils.py`.

freakboy3742:
I think an entirely new test module is called for - something to parallel datatypes, stdlib and structures. There's probably enough tests for each method that `test_abs.py` containing AbsTests, `test_bool.py` containing BoolTests etc wouldn't be excessive.

cflee:
I'm trying to figure out how comprehensive the tests are meant to be, since that affects how they should be defined. I quite like how the current UnaryOperationTestCase, etc are 'generating' all the possible combinations in a compact format, without cluttering up the individual test files. It means that we humans won't accidentally miss something out when implementing, particularly a strange edge/error case.

However, I'm not sure whether to do that for functions/methods - I can imagine the combinations exploding quite quickly, especially if each argument is tested with all data types, and with all possible combinations of arguments. It seems like the fraction of tests that would result in exercising some kind of error or another might be quite high.

(I've hacked up a `BuiltinFunctionTestCase` to just run all the builtin functions against the test values of each type, but that doesn't take into account 2+ arguments yet.)

freakboy3742:
For functions that take a single argument, I can see this being a good idea; however, it's going to get out of control pretty quickly once you move into functions with multiple arguments - if only because the multi-argument functions all have different combinations of arguments.

If you can put together a variant of UnaryOperationTestCase that can validate the simple, single argument methods (the ones like abs, bool, and so on), I think that would be worthwhile; however, for the more complex methods, I think aiming for branch coverage is more than enough.

cflee:
[...]

PR #112, opened 2016-03-27
Builtin tests and some builtin implementations

cflee:
This PR includes:
- Add new builtin test case that tests them against single arguments of all data types
  - Except set, frozenset and dict as those are unordered and the `assertEqual()` barfs on them occasionally
- Generate the builtin tests for the builtin functions
- Fix a nasty logic bug that wasn't exiting testdaemon, causing the process to run out of file descriptors once the new tests are added
- Implement some builtins' functionality: `abs()`, `bool()`, `all()`
  - Mark some unary-not tests as passing due to implementation of `bool()` on some data types.
- Fixed a missing optional args on the `round()` builtin

Outstanding issues:
- `Object.typeName()` seems to return names like `test.Widget`, but the expected TypeError message is supposed to be `Widget`. Not sure what is the best way to deal with this, the test is marked as expected failure now.


cflee:
- Re-merged master in to resolve the merge conflict, and resolve the non-conflicting (but not valid Java) double definition of `Tuple.__bool__()` that arose from the merge.
- Added NoneType to all the lists of non-implemented tests for builtins.
- Fix issues with the TypeError message generated in `org.Python` for non-iterable objects, and objects with no `len()`.
  - In particular, relying on the implicit `toString() -> __str__() -> __repr__()` on the `org.python.types.Type` class when doing `org.python.types.Type.pythonType(iterable.getClass())` doesn't work properly as it becomes `<class 'NoneType'>` instead of the expected `NoneType`. Switched to `iterable.typeName()` instead.

freakboy3742:
This is awesome - thanks!

cflee:
Oops, forgot to mention the issue: ref #76.

PR #114, opened 2016-04-05
Implementation of multiply methods of float

cherub96:
multiply_float,multiply_int,multiply_str,multiply_bool implemented.
Also, we have added the eq_str method for float.java


python/common/org/python/types/Int.java
```diff
@@ -108,9 +108,15 @@ public Int(long value) {
         } else if (other instanceof org.python.types.Float) {
             return new org.python.types.Bool(((double) this.value) == ((org.python.types.Float) other).value);
         } else if (other instanceof org.python.types.Bool) {
-            return new org.python.types.Bool(
-                (this.value == 0 && !((org.python.types.Bool) other).value)
-            );
+            if(this.value==0)
```
freakboy3742: Two minor style issues - there should be a space after the if, and if statements should *always* have braces, even if syntactically they aren't strictly required.

freakboy3742:
@cherub96 Looking good - almost ready for merge, but there's two small changes required.

I've flagged one of them in the code - they're minor code style issues.

The other is a piece of the patch that is still required - *deleting* code, rather than adding it. If you look in the timeline above at the commits you've made, you'll see there's a little red cross next to them. This means that the automated test suite failed when running with your patch. If you click on the cross, you'll be taken to a page that shows the test results - in particular, at the bottom, the summary says:

    Ran 3954 tests in 740.586s
    FAILED (skipped=2, expected failures=2155, unexpected successes=11)

That is, 11 tests passed, when the test suite thought they shouldn't. If you look through the test results, you should be able to find them - they're the 11 test results tied to the fix you've just made. The test suite doesn't know that you've added multiply, and it expects the tests to fail - but you *have* implemented them, and so they're passing unexpectedly.

To fix this, open up `tests.datatypes.test_float`, and delete the 11 lines that correspond to the tests that are now passing. Commit that change, push to Github, and the test suite should run again. When the red cross is replaced by a green tick, the patch will be ready to be merged.

If you need any more pointers, let me know!

cherub96:
@freakboy3742 , I have tried implementing only 5 tests for float class. So, which are the 11 test cases I need to remove?

freakboy3742:
The 11 tests that are passing are the ones from the log that end with `... unexpected success` - for example:

* test_eq_str (tests.datatypes.test_float.BinaryFloatOperationTests)
* test_multiply_bool (tests.datatypes.test_float.BinaryFloatOperationTests)
* test_multiply_float (tests.datatypes.test_float.BinaryFloatOperationTests)
* test_multiply_bool (tests.datatypes.test_float.InplaceFloatOperationTests)

(Note that there are both BinaryFloatOperationTests and InplaceFloatOperationTests)

The lines that need to be deleted are the ones that correspond to the tests that have unexpectedly passed - for the four tests listed above, that would be:
* [Line 73](https://github.com/pybee/voc/blob/master/tests/datatypes/test_float.py#L73)
* [Line 127](https://github.com/pybee/voc/blob/master/tests/datatypes/test_float.py#L127)
* [Line 133](https://github.com/pybee/voc/blob/master/tests/datatypes/test_float.py#L133)
* [Line 250](https://github.com/pybee/voc/blob/master/tests/datatypes/test_float.py#L250)

There are 7 others that need to be deleted - I'll leave working out which ones as an exercise :-)

cherub96:
I removed the test cases from test_float file and added the functionalities for set,dict,none types in Float.java file. Is it working fine now?

cflee:
Hi @cherub96, could you try to compile (`ant java`) and run the float tests (`python3 setup.py test -s tests.datatypes.test_float`) on the cherub96:master branch that you have submitted for this PR? It doesn't seem to compile successfully right now, so the tests cannot be run.

PR #119, opened 2016-04-08
Patch 3

cherub96:
Implementation of Boolean class.


python/common/org/python/types/Bool.java
```diff
@@ -65,41 +65,135 @@ public Bool(long int_val) {
...
+        } else if (other instanceof org.python.types.Str) {
+            throw new org.python.exceptions.TypeError("unorderable types: bool() < " + other.typeName() + "()");
+        } else if (other instanceof org.python.types.Dict) {
+            throw new org.python.exceptions.TypeError("unorderable types: bool() < " + other.typeName() + "()");
+        } else if (other instanceof org.python.types.NoneType) {
+            throw new org.python.exceptions.TypeError("unorderable types: bool() < " + other.typeName() + "()");
+        } else if (other instanceof org.python.types.Set) {
+            throw new org.python.exceptions.TypeError("unorderable types: bool() < " + other.typeName() + "()");
+        }
         throw new org.python.exceptions.NotImplementedError("bool.__lt__() has not been implemented.");
```
freakboy3742: What valid comparisons remain to be implemented? Is there any reason that the last four comparisons (Str, Dict, None, Set) could be collapsed into a single "else - unorderable type"?

cherub96:
@freakboy3742  I have removed the extra lines and merged them into a single else statement.

freakboy3742:
Fantastic! Thanks for the Bool and Float patches!

PR #120, opened 2016-04-09
WIP - Inbuilt sum() method

SrishtiSengupta:
Please review.
Ref #76 

-- Team l1ghtsab3r


cflee:
@SrishtiSengupta Looks like the CI build is failing because the `__sum__()` method is being invoked with the `wrong number of arguments`.

If you look at some of the other builtin methods in `org.Python` like [bin(number)](https://github.com/SrishtiSengupta/voc/blob/master/python/common/org/Python.java#L285), you might notice that the `@Method` annotation is slightly different. It contains the `args` component, which are for compulsory arguments. 

Also, if you scroll down slightly further to `bool([x])`, you'll see that it has `default_args` instead, for the optional arguments. These will be `null` if not provided, so your code needs to handle that and substitute it with the default value.

Finally, I see that you have found the tests for `sum()` - great! When you're done, the lines for tests that are passing should be deleted instead of commented out. However, note that the default `BuiltinFunctionTestCase` will only try to invoke the function with one argument - you can add tests to exercise it with various values of the optional `start` argument if you like.

freakboy3742:
Unfortunately, it someone else has submitted a fully working version of the sum builtin, so I'm going to close this ticket. Thanks for the contribution anyway - sorry we couldn't merge it.

Issue #38, opened 2015-12-04
Complete implementation of methods on standard types


freakboy3742:
The builtin types of Python have a number of methods associated with them - for example, `dict` objects have `update()`, `items()`, `keys()`, `pop()` and so on. These methods need to be implemented in Java. In many cases, there will be a simple mapping to an analogous function in the underlying Java type; however, sometimes you'll need to write an implementation using the available Java primitives.

Tests are are also required; they should go in the module corresponding to the data type [in the tests/datatypes directory](https://github.com/pybee/voc/blob/master/tests/datatypes).

Tests are based on comparing the output of some sample Python code when run through CPython to the output when run through VOC. When the two outputs are identical - and I mean byte-identical, down to the text and punctuation of the error message if appropriate - the test passes.

If you want an example to follow, look at [the implementation of `clear()` on `dict`](https://github.com/pybee/voc/blob/master/python/common/org/python/types/Dict.java#L120):
* The `@org.python.Method` annotation is used to expose the clear() method as a symbol that should be available at runtime. 
* The argument list for all exposed symbols is `List args, Map kwargs`. 
* These arguments are checked to make sure they are both empty (`clear()` takes no arguments)
* The body code defers to an implementation that *has* no arguments. This is provided so that if we need to call `clear()` from within Java natively, we don't have to jump through the hoops of creating an empty list and dict.
* The "empty args" implementation defers to the `java.lang.Map.clear()` method which performs the same task.
* Finally, the method returns a Python `None`.

TalSkverer:
Hi there, first timer here, trying to help :smile: 

I'm working on the `dict` builtin type, but I've got a encountered problems and I need some assistance:
1. `dict` functions like `pop()`, `get()` can get an additional, optional value to be returned on default, yet the Java functions only have one argument. Is this correct? Could another optional argument be added?
2. Should I add my tests to `DictTests` class or create my own?
3. I did not manage to figure out how to run any test. Do I run `setup.py` again and use voc as usual? 
4. I tried to run `ant java` to get the support library as well, but it returned an error.

Thanks in advance!

freakboy3742:
Hi Tal! Thanks for offering to help out!

To answer your questions:
[...]

PR #122, opened 2016-04-15
Rich comparison methods

cflee:
Ref #38:

> > The “rich comparison” methods. When called directly (`{}.__lt__(1)`) they return `"NotImplemented"`, and when called using `{} < 1` they raise the following excpetion: `TypeError: unorderable types: dict() < int()`. What should the code do, then?
> 
> Implement `__lt__()` as if it were being invoked directly as a function. The code to convert the `NotImplemented` into a `TypeError` will go in the implementation of the `COMPARE_OP` Python opcode.

This PR:
- Modifies various comparison methods like `__lt__()` to act as if they were called directly, such that they return `NotImplemented` instead of throwing `TypeError`, and also add the `args` value to the `@Method` annotations so that they can be invoked directly
- Adds a new bunch of tests, `test_direct_lt_[datatype]` etc. to verify this
- ~~Adds a new helper `comparison` method to `org.python.Object`, which implements the [reflected operation logic](https://docs.python.org/3/reference/datamodel.html#object.__lt__)~~
- ~~Modifies `COMPARE_OP` opcode to invoke the helper `comparison` method instead of the comparator method directly~~
- Implements the [reflected operation logic](https://docs.python.org/3/reference/datamodel.html#object.__lt__) in `ast.py:visit_Compare`, such that the logic is directly generated into bytecode for each compare operator.
- Implements the `issubclass()` and `isinstance()` builtin methods (refs #76) to support the reflected operation logic.
- Modifies the standard `Object.__ne__()` logic to depend on `__eq__()`, as defined in the reflected operation logic, see also [python issue4395](http://bugs.python.org/issue4395). Removed `__ne__()` from all datatypes.
- Disabled stackmap verification using `-noverify` for Java 8 JVM when running tests, as otherwise there will be a `java.lang.VerifyError: Expecting a stackmap frame at branch target xx`.
- **Note:** Some of the `test_direct_ne_*` tests will fail on CPython 3.4.2 due to a bug [fixed in 3.4.3](https://docs.python.org/3.4/whatsnew/changelog.html#id6) (see [python issue21408](http://bugs.python.org/issue21408)). As VOC currently targets the latest release of python 3.4, this quirk is not implemented.

It simplifies implementing the comparison methods on datatypes, as fewer combinations need to be handled. For example, Int and Float don't have to know about Complex; it's up to Complex to handle reflected comparisons with Int and Float.


cflee:
Potential issues that require review:

* Duplication of operator to method name mapping from Python into Java, do these need to be shifted back into opcode layer?
* Should I just copy and paste the new method from `org.python.types.Object` to `Super` and `BaseException` to replace the `return null` stubs?
* This adds a bunch of tests, since now `__lt__()` is tested as well as `<`, so the test runtime increases again...

Also, I have merged master in to resolve the conflicts (and make some edits to make sure everything passes) but it still says that there are conflicts here for some reason. [Here is the green Travis build](https://travis-ci.org/cflee/voc/builds/123293669) of 3c3e3f1 on my fork.

python/common/org/python/types/Dict.java
```diff
@@ -190,11 +190,11 @@ public boolean __setattr_null(java.lang.String name, org.python.Object value) {
             // exception is thrown.
             org.python.Object hashcode = item.__hash__();
```
freakboy3742: Codacy Issue found: [Avoid unused local variables such as 'hashcode'.](https://www.codacy.com/app/freakboy3742/voc/file/2412210734/issues/source?bid=3235301&fileBranchId=3235629#l191)

cflee:
Oops, so the conflict persisted because I merged in an old `master`. Just pushed another merge commit to fix that.

python/common/org/python/types/NotImplementedType.java
```diff
@@ -50,43 +50,51 @@ public boolean __setattr_null(java.lang.String name, org.python.Object value) {
...
-        );
+        return org.python.types.NotImplementedType.NOT_IMPLEMENTED;
     }
 
     @org.python.Method(
-        __doc__ = ""
+        __doc__ = "",
+        args = {"other"}
     )
     public org.python.Object __le__(org.python.Object other) {
```
freakboy3742: Codacy Issue found: [Method names should not contain underscores](https://www.codacy.com/app/freakboy3742/voc/file/2412364122/issues/source?bid=3235301&fileBranchId=3235629#l91)

python/common/org/python/types/NotImplementedType.java
```diff
@@ -50,43 +50,51 @@ public boolean __setattr_null(java.lang.String name, org.python.Object value) {
...
-        );
+        return org.python.types.NotImplementedType.NOT_IMPLEMENTED;
     }
 
     @org.python.Method(
-        __doc__ = ""
+        __doc__ = "",
+        args = {"other"}
     )
     public org.python.Object __mul__(org.python.Object other) {
```
freakboy3742: Codacy Issue found: [Method names should not contain underscores](https://www.codacy.com/app/freakboy3742/voc/file/2412364122/issues/source?bid=3235301&fileBranchId=3235629#l99)

python/common/org/python/types/NotImplementedType.java
```diff
@@ -100,7 +108,8 @@ public boolean __setattr_null(java.lang.String name, org.python.Object value) {
     }
 
     @org.python.Method(
-        __doc__ = ""
+        __doc__ = "",
+        args = {"other"}
     )
     public org.python.Object __getitem__(org.python.Object other) {
```
freakboy3742: Codacy Issue found: [Method names should not contain underscores](https://www.codacy.com/app/freakboy3742/voc/file/2412364122/issues/source?bid=3235301&fileBranchId=3235629#l114)

python/common/org/python/types/NotImplementedType.java
```diff
@@ -50,43 +50,51 @@ public boolean __setattr_null(java.lang.String name, org.python.Object value) {
...
+        return org.python.types.NotImplementedType.NOT_IMPLEMENTED;
     }
 
-        @org.python.Method(
-        __doc__ = ""
+    @org.python.Method(
+        __doc__ = "",
+        args = {"other"}
     )
     public org.python.Object __gt__(org.python.Object other) {
```
freakboy3742: Codacy Issue found: [Method names should not contain underscores](https://www.codacy.com/app/freakboy3742/voc/file/2412364122/issues/source?bid=3235301&fileBranchId=3235629#l64)

python/common/org/python/types/NotImplementedType.java
```diff
@@ -50,43 +50,51 @@ public boolean __setattr_null(java.lang.String name, org.python.Object value) {
     }
 
     @org.python.Method(
-        __doc__ = ""
+        __doc__ = "",
+        args = {"other"}
     )
     public org.python.Object __ge__(org.python.Object other) {
```
freakboy3742: Codacy Issue found: [Method names should not contain underscores](https://www.codacy.com/app/freakboy3742/voc/file/2412364122/issues/source?bid=3235301&fileBranchId=3235629#l56)

python/common/org/python/types/NotImplementedType.java
```diff
@@ -50,43 +50,51 @@ public boolean __setattr_null(java.lang.String name, org.python.Object value) {
...
-        );
+        return org.python.types.NotImplementedType.NOT_IMPLEMENTED;
     }
 
     @org.python.Method(
-        __doc__ = ""
+        __doc__ = "",
+        args = {"other"}
+    )
+    public org.python.Object __eq__(org.python.Object other) {
```
freakboy3742: Codacy Issue found: [Method names should not contain underscores](https://www.codacy.com/app/freakboy3742/voc/file/2412364122/issues/source?bid=3235301&fileBranchId=3235629#l72)

freakboy3742:
Gah - Apologies for the comment noise on this. I've been trying out some automated code linting tools... this one might be a wash.

cflee:
No worries, I was casually ruminating a few weeks ago about suggesting an automatic code-style checker too.. I guess it needs some rules configuration at the very least.

python/common/org/python/Object.java
```diff
@@ -32,6 +32,11 @@
     public org.python.Object byValue();
 
     /**
+     * Perform the appropriate reflected comparison operation
+     */
+    public org.python.Object comparison(org.python.Object other, String operator);
```
freakboy3742: For historical reasons, and to preserve the `__`-method-means-internals metaphor, this should probably be called `__cmp__`. Strictly, Python3 doesn't need a __cmp__ operator, but if we're going to introduce one for our own purposes, it wouldn't hurt to retain consistency.

python/common/org/python/types/Object.java
```diff
@@ -32,6 +32,84 @@ void setValue(org.python.Object obj) {
...
+        // 4.3: Objects of different types, except different numeric types, never compare equal.
+        if (operator.equals("==")) {
+            return new org.python.types.Bool(false);
+        } else if (operator.equals("!=")) {
+            // Non-identical instances of a class normally compare as non-equal unless the class defines the __eq__() method.
+            return new org.python.types.Bool(true);
+        }
+        // 4.3: The <, <=, > and >= operators will raise a TypeError exception when comparing a complex number with another built-in numeric type, when the objects are of different types that cannot be compared, or in other cases where there is no defined ordering.
+        throw new org.python.exceptions.TypeError(String.format(
+        "unorderable types: %s() %s %s()", this.typeName(), operator, other.typeName()));
```
freakboy3742: This should be indented so it looks like a continuation, rather than a new statement.

freakboy3742:
Ok - on the whole this looks pretty good. 

In a comment, I noted that `comparision` should be `__cmp__`; however, as you note in your first question, it's probably a moot point anyway, because that whole method should be in the opcode layer, not the Java code. 

If I'm reading this right, that would also resolve your second question, since the opcode wouldn't care if it has a `Super`, `BaseException` or an `Object`. 

As for making the test suite run longer - that doesn't bother me. I'd rather the code work :-) 

I'm also wondering if the reflected operations docs you referenced has a bug in it - shouldn't `__ge__` be the reflection of `__lt__`, not `__le__`? Similarly for `__le__` and `__gt__`, instead of `__ge__`?

cflee:
> In a comment, I noted that `comparision` should be `__cmp__`; however, as you note in your first question, it's probably a moot point anyway, because that whole method should be in the opcode layer, not the Java code.

I initially tried to do this directly in the opcode layer by compiling it as a separate function and transcribing the `javap` output into `opcodes.py`, but I was running into this `VerifyError`. (It was an `astore_2` related thing earlier but essentially it's the `Local variable table overflow` that is the issue?)

```
Error: A JNI error has occurred, please check your installation and try again
Exception in thread "main" java.lang.VerifyError: Local variable table overflow
Exception Details:
  Location:
    python/test/__init__.<clinit>()V @233: aload_1
  Reason:
    Local index 1 is invalid
  Bytecode:
    0x0000000: 014b b200 0813 000e bb00 0359 1300 10b8
    0x0000010: 0012 b700 18b9 001c 0300 57bb 0022 59bb
    0x0000020: 0024 5905 b700 2659 bb00 2959 04b7 002b
    0x0000030: b900 2c02 0057 59bb 0029 5905 b700 2bb9
    0x0000040: 002c 0200 57b7 0032 4bb2 0008 1300 0eb9
    0x0000050: 0035 0200 c000 0313 0039 2ab9 003b 0300
    0x0000060: bb00 2259 bb00 2459 04b7 0026 59bb 0029
    0x0000070: 5906 b700 2bb9 002c 0200 57b7 0032 4bb2
    0x0000080: 0008 1300 0eb9 0035 0200 c000 0313 0041
    0x0000090: 2ab9 003b 0300 b200 0813 000e b900 3502
    0x00000a0: 00c0 0003 1300 43b9 0045 0200 c000 4904
    0x00000b0: bd00 3c59 03b2 0008 1300 0eb9 0035 0200
    0x00000c0: c000 0313 0039 b900 4502 00b2 0008 1300
    0x00000d0: 0eb9 0035 0200 c000 0313 0041 b900 4502
    0x00000e0: 00b2 004b 4b2a b600 512b b600 51a5 001f
    0x00000f0: 2ab6 0051 2bb6 0057 a500 142b 2ab9 005b
    0x0000100: 0200 4b2a b200 4ba5 0005 2db0 2a2b b900
    0x0000110: 5b02 004b 2ab2 004b a500 052a b01c a000
    0x0000120: 142b 2ab9 005b 0200 4b2a b200 4ba5 0005
    0x0000130: 2ab0 bb00 5f59 1300 6106 bd00 5259 032a
    0x0000140: b900 6301 0053 5904 1300 6753 5905 2bb9
    0x0000150: 0063 0100 53b8 0068 b700 6ebf 53bb 0071
    0x0000160: 59b7 0073 b900 7503 0057 b200 7957 b1b1
    0x0000170:                                        

	at java.lang.Class.getDeclaredMethods0(Native Method)
	at java.lang.Class.privateGetDeclaredMethods(Class.java:2701)
	at java.lang.Class.privateGetMethodRecursive(Class.java:3048)
	at java.lang.Class.getMethod0(Class.java:3018)
	at java.lang.Class.getMethod(Class.java:1784)
	at sun.launcher.LauncherHelper.validateMainClass(LauncherHelper.java:544)
	at sun.launcher.LauncherHelper.checkAndLoadMain(LauncherHelper.java:526)
```

Basically if there's some way to get space to store and retrieve variables (object references and booleans) it would be very helpful, otherwise I'll probably try to figure out how to get the references to the right spot on operand stack by hand using `DUP` and stuff. It seems that the x and y are already placed on the operand stack just before this `COMPARE_OP`'s bytecode starts, since that's where the `invokeinterface` for `org.python.Object.[comparator]` would happen.

> I'm also wondering if the reflected operations docs you referenced has a bug in it - shouldn't `__ge__` be the reflection of `__lt__`, not `__le__`? Similarly for `__le__` and `__gt__`, instead of __ge__?

It is a little strange but unfortunately it doesn't seem to be a bug. Try this:

```python
class Ge:
    def __init__(self, val):
        self.val = val
    def __ge__(self, other):
        return self.val

x = 5
y1 = Ge(True)

# True: using __ge__
print(y1 >= x)
# True: __le__ reflects to __ge__
print(x <= y1)

# TypeError: __lt__ reflects to __gt__ (not implemented) not __ge__
try:
    print(x < y1)
except TypeError as err:
    print(err)
```

sjoerdjob:
@cflee are you still working on this? If not, I would not mind taking a further stab at it.

cflee:
@sjoerdjob 
I was briefly re-examining this a few weeks back, but didn't have any good ideas on making the `comparison()` (or `__cmp__()`) method logic work on the operand stack without storing anything in local variables.

Please go ahead! I would love to see this landed in some other form, it makes almost all the basic types' comparison methods simpler.

cflee:
Ok, while debugging this, I realised that the subclass checking is wrong. I was using `java.lang.Class.isInstance()`, which examines the Java class hierarchy, not the correct one for Python classes.

Since `org.Python.issubclass()` is still unimplemented, this will still be incorrect even after I sort out the bytecode stuff. The remaining snag seems to be that this:
```
if (var10004 == NotImplementedType.NOT_IMPLEMENTED) {
    if (var2.getClass() != var1.getClass() && !var1.getClass().isInstance(var2)) {
```

is coming out as:
```
if (var10004 == NotImplementedType.NOT_IMPLEMENTED && var2.getClass() != var1.getClass()) {
    if (!var1.getClass().isInstance(var2)) {
```

cflee:
For future reference, the reflected operation logic implemented in `ast.py` is an approximation of:

```
// Let x and y be the left and right operands respectively, i.e. `x < y`.
org.python.Object y = new org.python.types.Float(2.0);
org.python.Object x = new org.python.types.Int(5);

org.python.Object result = org.python.types.NotImplementedType.NOT_IMPLEMENTED;
boolean rhs = x.type() != y.type() && org.Python.isinstance(y, x.type());

if (rhs) {
    result = y.__gt__(x); // reflected
}

if (result == org.python.types.NotImplementedType.NOT_IMPLEMENTED) {
    result = x.__lt__(y); // normal

    if (result == org.python.types.NotImplementedType.NOT_IMPLEMENTED) {
        if (!rhs) {
            result = y.__gt__(x); // reflected

            if (result == org.python.types.NotImplementedType.NOT_IMPLEMENTED) {
                if (operator is ==) {
                    // == defaults to false instead of TypeError
                    result = new org.python.types.Bool(0);
                } else if (operator is !=) {
                    // != defaults to true instead of TypeError
                    result = new org.python.types.Bool(1);
                } else {
                    // there is a conditional to handle the changed TypeError message
                    // for python 3.6+
                    result = new org.python.exceptions.TypeError(String.format(
                        "unorderable types: %s() %s %s()", x.typeName(), "<", y.typeName()));
                }
            }
        }
    }
}
```

This logic may be useful for implementing reflected operands for math operators (refs #213) as that has similar [subclass priority rules](https://docs.python.org/3/reference/datamodel.html#object.__radd__).

cflee:
After nearly 11 months, this is ready for review! I have updated the overview in initial PR message as well.

python/common/org/Python.java
```diff
@@ -814,15 +814,9 @@ public static void initializeModule(java.lang.Class cls, java.util.Map<java.lang
...
             throw new org.python.exceptions.TypeError("isinstance expected 2 arguments, got 1");
         } else if (class_or_type_or_tuple instanceof org.python.types.Tuple) {
-            java.util.List<org.python.Object> classes = ((org.python.types.Tuple) class_or_type_or_tuple).value;
-            for (org.python.Object klass: classes) {
-                if (((org.python.types.Bool) org.Python.isinstance(object, klass).__bool__()).value) {
-                    return new org.python.types.Bool(true);
-                };
-            }
-            return new org.python.types.Bool(false);
+            return org.Python.issubclass(object.type(), class_or_type_or_tuple);
```
sjoerdjob: In Python 2/3, you can override the `isinstance` check using a meta class. So this code will have to be changed at a later date, but that can be done for a separate issue.

python/common/org/python/types/Dict.java
```diff
@@ -286,11 +267,11 @@ public boolean __setattr_null(java.lang.String name, org.python.Object value) {
...
 
-            org.python.Object value = this.value.get(item);
-            if (value == null) {
-                throw new org.python.exceptions.KeyError(item);
-            }
-            return value;
+        org.python.Object value = this.value.get(item);
```
sjoerdjob: This change in indent seems "off". Now the nested code no longer looks nested.

python/common/org/python/types/Dict.java
```diff
@@ -308,7 +289,7 @@ public void __setitem__(org.python.Object item, org.python.Object value) {
             // exception is thrown.
             org.python.Object hashcode = item.__hash__();
 
-            this.value.put(item, value);
+        this.value.put(item, value);
```
sjoerdjob: Should this "dedent" also be undone?

python/common/org/python/types/List.java
```diff
@@ -167,8 +168,14 @@ public List(org.python.Object [] args, java.util.Map<java.lang.String, org.pytho
             int count = Math.min(size, otherSize);
 
             for (int i = 0; i < count; i++) {
-                org.python.types.Bool b =
-                    (org.python.types.Bool) this.value.get(i).__lt__(otherList.value.get(i));
+                org.python.Object result = this.value.get(i).__lt__(otherList.value.get(i));
```
sjoerdjob: Shouldn't this go through the `compare` again, letting `compare` deal with formatting the exception, handling the type error, etc?

(I may be wrong, and there's a good reason for doing it inline here, though).

tests/builtins/test_isinstance.py
```diff
@@ -70,6 +67,24 @@ class MyClass2(MyClass1):
             print(isinstance(obj, MyClass2))
             """, run_in_function=False)
 
+    def test_subclass_grandparent(self):
```
sjoerdjob: Maybe should be called `test_isinstance_grandparent`? A similar test should perhaps than also be built for `issubclass(MyClass3, MyClass1)`.

python/common/org/python/types/Dict.java
cflee: You're right – fixed in 63a856a.

python/common/org/python/types/Dict.java
cflee: Same – fixed in 63a856a.

python/common/org/python/types/List.java
cflee: Sorry, I'm not too sure what you mean by `compare` here. Do you mean the logic implemented in the ast? If so, I think there's no nice way to invoke that since there is no longer a `__cmp__()` Java method.

tests/builtins/test_isinstance.py
cflee: Hmm, there is is a `issubclass(MyClass3, MyClass1)` at [`test_issubclass.py:75`](https://github.com/pybee/voc/blob/31e0cee28f9e6ff7b4c8cdcacf9ed3cae5d1b222/tests/builtins/test_issubclass.py#L75).

Here I was following the previous `test_subclass()` test case, but I think I will rename them to `test_parent()` and `test_grandparent()` instead.

python/common/org/python/types/List.java
sjoerdjob: I'm talking about calling the following method

```public int compare(org.python.Object o1, org.python.Object o2) {```

(in `Python.java`).

Comment (cflee) [inline python/common/org/python/types/List.java]: I see! Hmm, `__SortedObjectComparator.compare()` could do with some refactoring since the logic seems to be repeated twice inside after all, and I wouldn't want to instantiate what looks like a inner utility class elsewhere.

However, now that I'm looking at it, `__SortedObjectComparator.compare()` plus all the `List` dunder comparison methods _don't_ implement the subclass priority rule, so they might be faulty on top of not being DRY. If CPython _does_ do all that, I guess the `org.python.types.Object.__cmp__()` helper method might have to be revived...

Comment (sjoerdjob) [inline python/common/org/python/types/List.java]: Looking at the Python source code, I *think* the way forward would be to have a method `do_richcompare`: https://github.com/python/cpython/blob/master/Objects/object.c#L655

Comment (sjoerdjob) [inline python/common/org/python/types/List.java]: (Also, there should probably be a `PyObject_RichCompareBool` which does some extra checks.

```
>>> class F:
...     def __init__(self, v):
...         self.v = v
...     def __eq__(self, other):
...         return self.v == other.v + 1
...
>>> k = F(5)
>>> k == k
False
>>> [k] == [k]
True
>>> {'hi': k} == {'hi': k}
True
```

Looking at the Python source code, it seems the "normal" compare (`COMPARE_OP`) goes to `PyObject_RichCompare`, while containers (list/dict) use `PyObject_RichCompareBool`). Of course that's not a precise argument, but it does indicate that maybe there should be several methods.

Comment (cflee) [inline python/common/org/python/types/List.java]: I really should've spent more time poking at the cpython source earlier >_<

This PR's current implementation seems largely similar to `PyObject_RichCompare`, but with a few differences: (i) == instead of defaulting to False should return a Bool of whether they're the same object, similarly for !=, (ii) the operators can return non-Bool (and non-NotImplemented) values, which I'm not sure works properly now, (iii) 100 byte limit on type name in TypeError string.

I should patch and test (i) and (iii), and look into the behaviour of (ii).

As for `PyObject_RichCompareBool`, it seems to be primarily coercing the result to a Bool, with an optimization for == and != with the same object.

I agree that some implementation of `PyObject_RichCompare` should be available as a method. It seems to be needed to correctly implement other datatypes like slice, tuple, complex, [list](https://github.com/python/cpython/blob/v3.6.0/Objects/listobject.c#L2251) in cpython, and it's quite annoying to have to duplicate the logic across all of them.

cc @freakboy3742: 
1. Would it be better to still duplicate this logic in the AST code, or have it invoke the static method that is needed to implement the datatypes correctly?
2. Where would be a good place to put this static method?

Comment (sjoerdjob) [inline python/common/org/python/types/List.java]: With respect to 2. (I know you CC'd @freakboy3742, but I think I have a possible answer, if he disagrees he can correct me).

In Python source code, `PyObject_RichCompare(Bool)` is in `object.c`. To me, this signifies that the method should probably be a static method `RichCompare(Bool)` on `org.python.types.Object`.

Also, from the current list of `.java` files, I can't find a better location.

Comment (cflee) [inline python/common/org/python/types/List.java]: @sjoerdjob Right, I mostly meant to direct question 1 at him.

I do agree, but this time I'll probably do it on `org.python.types.Object` as you point out, instead of the `org.python.Object` interface as I did on a previous iteration...

Comment (cflee) [inline python/common/org/python/types/List.java]: @sjoerdjob This particular line is addressed in 2dc1655.

File: python/common/org/Python.java [hunk 15]
```diff
@@ -814,15 +814,9 @@ public static void initializeModule(java.lang.Class cls, java.util.Map<java.lang
...
         } else if (class_or_type_or_tuple instanceof org.python.types.Type) {
-            return new org.python.types.Bool(object.type() == class_or_type_or_tuple);
+            return org.Python.issubclass(object.type(), class_or_type_or_tuple);
```
eliasdorneles: Isn't it better to use `object.__class__` here instead of `object.type()` ?
That `Object.type()` public method should probably be deprecated.

voc/python/ast.py
```diff
@@ -1709,23 +1709,53 @@ def visit_Compare(self, node):
...
-                        '(Lorg/python/Object;)Lorg/python/Object;'
-                    )
+                    JavaOpcodes.LDC_W(oper_symbol),
+                    JavaOpcodes.LDC_W(oper),
+                    JavaOpcodes.LDC_W(reflect_oper),
+                    JavaOpcodes.INVOKESTATIC(
+                        'org/python/types/Object',
+                        '__cmp__',
+                        '(Lorg/python/Object;Lorg/python/Object;Ljava/lang/String;Ljava/lang/String;'
+                        + 'Ljava/lang/String;)Lorg/python/Object;'
```
freakboy3742: This can get cleaned up if you use the `args` and `returns` arguments to `INVOKESTATIC` - this allows you to specify the arguments as a list, rather than a concatenated string.

python/common/org/python/types/Object.java
```diff
@@ -979,4 +990,110 @@ public void __delitem__(org.python.Object index) {
...
     }
+
+    // Need to implement Is __eq__ Is Not __ne__
+    // In __contains__ (reversed operands) Not In __not_contains__ (reversed operands)
+    //
+    // for Is and Is Not, falls through to here only when either side is a constant in python
+    // which ends up as org.python.types.[Int|Float|Complex], otherwise it's dealt with
+    // by object reference comparison IF_ACMEQ. so we fall through to __eq__!
+    public enum CMP_OP {
+        GE(">=", "__ge__", "__le__"),
```
freakboy3742: Is this right? Shouldn't the reflection of `__ge__` be `__lt__`, rather than `__le__` (similarly for the other operations)? And isn't is the reflection of `__eq__` `__ne__` (and vice versa)? Or am I missing what the reflection is used for here?

tests/utils.py
```diff
@@ -313,7 +313,7 @@ def setUpClass(cls):
             os.path.join('dist', 'python-java-support.jar'),
         ])
         cls.jvm = subprocess.Popen(
-            ["java", "-classpath", classpath, "python.testdaemon.TestDaemon"],
+            ["java", "-noverify", "-classpath", classpath, "python.testdaemon.TestDaemon"],
```
freakboy3742: Isn't this only needed if you're using Java 8?

python/common/org/Python.java
cflee: I was going off this statement:
> This type can be retrieved by calling the `type()` method on the Java instance (which is part of the `org.python.Object` interface)

But sure, `org.Python.type(object)` works too. Fixed in a963a1a.

voc/python/ast.py
cflee: Cool, I didn't know I could do that! Fixed in f45a1cd.

python/common/org/python/types/Object.java
cflee: It's rather strange mathematically for `ge/gt/le/lt`, but:

From [python/cpython/Objects/object.c at v3.6.0](https://github.com/python/cpython/blob/v3.6.0/Objects/object.c#L647):
```
/* Map rich comparison operators to their swapped version, e.g. LT <--> GT */
int _Py_SwappedOp[] = {Py_GT, Py_GE, Py_EQ, Py_NE, Py_LT, Py_LE};
static const char * const opstrings[] = {"<", "<=", "==", "!=", ">", ">="};
```

From [3. Data model – Python 3.5](https://docs.python.org/3.5/reference/datamodel.html?highlight=data%20model#object.__lt__):
> There are no swapped-argument versions of these methods (to be used when the left argument does not support the operation but the right argument does); rather, `__lt__()` and `__gt__()` are each other’s reflection, `__le__()` and `__ge__()` are each other’s reflection, and `__eq__()` and `__ne__()` are their own reflection.

As for `eq/ne`, let's say we're evaluating `x == y`. The reflection is in the sense that if `x.__eq__(y)` returns NotImplemented, then they will perform the reflected operation on y with x as parameter, i.e. `y.__eq__(x)`.

tests/utils.py
cflee: Yes, it _was_ needed for my local tests on Java 8 – but the need seems to have gone away now that the AST bit is no longer fiddling with `IF` opcode jump targets. Fixed in 0777522.

python/common/org/python/types/Object.java
sjoerdjob: The swapped ops are not used for "negation" but for reverting the order of arguments.

Mathematically speaking, `x.__ge__(y)` if and only if `y.__le__(x)` [*if* `x` and `y` are of the right type.

Issue #123, opened 2016-04-24
Error in test suite when running under Python 3.4.4


freakboy3742:
Via Kesha Shah:
The CI server runs under Python 3.4.2, and is (usually) in a fully passing state. 

However, if you run the test suite under Python 3.4.4, you will get 3 test failures:
```
======================================================================
FAIL: test_list (tests.builtins.test_int.BuiltinIntFunctionTests)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/Users/keshashah/VOC/voc-dev/voc/tests/utils.py", line 641, in func
    self.assertBuiltinFunction(x=example, f=function, operation=operation, format=self.format)
  File "/Users/keshashah/VOC/voc-dev/voc/tests/utils.py", line 661, in assertBuiltinFunction
    """ % kwargs, "Error running %(operation)s with f=%(f)s, x=%(x)s" % kwargs)
  File "/Users/keshashah/VOC/voc-dev/voc/tests/utils.py", line 373, in assertCodeExecution
    self.assertEqual(java_out, py_out, context)
AssertionError: "### [52 chars]tring or a number, not 'list'\n    test.py:3\n" != "### [52 chars]tring, a bytes-like object or a number, not 'l[17 chars]:3\n"
  ### EXCEPTION ###
- TypeError: int() argument must be a string or a number, not 'list'
+ TypeError: int() argument must be a string, a bytes-like object or a number, not 'list'
?                                           +++++++++++++++++++++
      test.py:3
 : Global context: Error running f(x) with f=int, x=[]
 
======================================================================
FAIL: test_none (tests.builtins.test_int.BuiltinIntFunctionTests)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/Users/keshashah/VOC/voc-dev/voc/tests/utils.py", line 641, in func
    self.assertBuiltinFunction(x=example, f=function, operation=operation, format=self.format)
  File "/Users/keshashah/VOC/voc-dev/voc/tests/utils.py", line 661, in assertBuiltinFunction
    """ % kwargs, "Error running %(operation)s with f=%(f)s, x=%(x)s" % kwargs)
  File "/Users/keshashah/VOC/voc-dev/voc/tests/utils.py", line 373, in assertCodeExecution
    self.assertEqual(java_out, py_out, context)
[...]

PR #124, opened 2016-04-24
Modified error message to be compatible with Python 3.4.4

keshashah:
Fix for Issue #123

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,6 @@
 language: python
 python:
-  - "3.4"
+  - "3.4.4"
 install:
   - "pip install ."
   - "ant java"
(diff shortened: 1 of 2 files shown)


keshashah:
Upgraded CI to 3.4.4 and it passes now. Safe to merge. 
Will check for a matrix that runs CI across both Python versions- 3.4.2 and 3.4.4

freakboy3742:
FYI - I've addressed the multi-platform build in 499cd511.

I also modified the error message handling in 5d378ce3, so both 3.4.2 and 3.4.4 will pass.

Thanks for the contribution, and for the help diagnosing why you were getting errors!

PR #128, opened 2016-04-27
Run CI on Python versions 3.4.2 and 3.4.4

tapasweni-pathak:


freakboy3742:
While I can see that this change passes CI, it isn't immediately clear to me why the change is needed. Does the extra layer of "matrix/include" change anything in particular with the build process, or is it just a different (or preferred) way of spelling the same thing?

tapasweni-pathak:
This is related to the discussion we had over email for the issue #123. We need to run a matrix build on CI so that both 3.4.2, 3.4.4, are tested with every commit. Isn't this what we need?

cflee:
It appears that the [latest master build](https://travis-ci.org/pybee/voc/builds/125977519) is already testing on both 3.4.2 and 3.4.4.

I think that the matrix include/exclude part is only necessary if you want to add/remove specific combinations, instead of letting Travis generate the combinations. In this case it's generating 2×1 builds, which seems to be exactly what is needed.

tapasweni-pathak:
@cflee Yes. I got it now. When I started working on it, (enabling CI for my repo took some time), it was not there, only 3.4.2 was there. I didn't check it and submitted the PR. I'll close this one.

PR #134, opened 2016-06-01
Add next() builtin implementation and tests.

ericatkin:


project.properties
```diff
@@ -12,4 +12,4 @@
 
 android.library=true
 # Project target.
-target=android-14
+target=android-23
```
freakboy3742: Not completely sure of the android etiquette here - is there a reason to keep this "up to date"? I originally thought the best approach was "keep this number as low as possible" for backwards compatibility reasons.

python/common/org/Python.java
```diff
@@ -1005,10 +1005,24 @@ public static void initializeModule(java.lang.Class cls, java.util.Map<java.lang
...
             "Return the next item from the iterator. If default is given and the iterator\n" +
-            "is exhausted, it is returned instead of raising StopIteration.\n"
+            "is exhausted, it is returned instead of raising StopIteration.\n",
+        args = {"iterator"},
+        default_args = {"_default"}
     )
-    public static org.python.Object next() {
-        throw new org.python.exceptions.NotImplementedError("Builtin function 'next' not implemented");
+    public static org.python.Object next(org.python.Object iterator, org.python.Object _default) {
+        if (iterator instanceof org.python.Iterable) {
```
freakboy3742: `instanceof` works; but if we can wrap this into the `utils.isinstance` mechanism, that would be even better.

project.properties
ericatkin: 23 is all I had. I wasn't sure either. I have no problem reverting to 14.

python/common/org/Python.java
freakboy3742: Scratch that - forgot this is VOC, not Batavia.

PR #137, opened 2016-06-02
correct relative folders

julisi:
starting in tutorial0 directory:

   "This will produce an ``__init__.class`` in the ``python/example`` namespace.
    This classfile can run on any Java 6 (or higher) VM. To run the project, type:

      .. code-block:: bash

          $ java -classpath ../voc/dist/python-java.jar:. python.example.__init__
          Hello, World

   Congratulations! You've just run your first Python program under Java using
   VOC! Now you're ready to get a little more adventurous."

the /voc doesn't work.  The command line should read:

`$ java -classpath ../dist/python-java.jar:. python.example.__init__`

### DIFF
--- a/docs/tutorials/tutorial-0.rst
+++ b/docs/tutorials/tutorial-0.rst
@@ -55,7 +55,7 @@ This classfile can run on any Java 6 (or higher) VM. To run the project, type:
 
 .. code-block:: bash
 
-    $ java -classpath ../voc/dist/python-java.jar:. python.example.__init__
+    $ java -classpath ../dist/python-java.jar:. python.example.__init__
     Hello, World
 
 Congratulations! You've just run your first Python program under Java using


freakboy3742:
I know you had this problem; however, I think the issue is that you've made the tutorial0 directory in the wrong location. You *should* have a directory structure that looks like:

* tutorial
  - voc
  - tutorial0

But it looks like you've got

* tutorial
  - voc
    * tutorial0

The key piece of documentation is the line:

    Let’s start by creating a tutorial0 directory alongside the voc directory you just cloned into:

However, it's easy to see how you made this mistake, and you're not the first person to make it. The text clearly needs to be clarified. I'm open to suggestions on how to improve it - perhaps including a "your directory tree should now look like..." example?

julisi:
Ah, that makes sense. I will look at the wording and see if I can suggest
something.

freakboy3742:
Closing on the basis that #142 cleared up this problem.

PR #138, opened 2016-06-02
Make note about setting up Ubuntu Dev Environment

gopar:
In Ubuntu, you need to install `libpython3.4-testsuite` to be able to run the test suite. Without it, running the tests will just fail.


katlings:
I accomplished the same by running `pip install test` - for a python-based alternative :)

PR #139, opened 2016-06-02
Refs #36 - throw correct TypeErrors for dict multiply

katlings:
Looks like multiplying a dict by a sequence in Python prioritizes throwing the sequence TypeError instead of a generic TypeError. 

#firstpr!


katlings:
Updated tabs -> spaces

PR #148, opened 2016-06-03
working on __mod__ and __imod__

jphilipsen05:


freakboy3742:
Thanks for the contribution - I'm going to close this pull request, though. The tests that it fails point to the fact that the implementation is a little too simplistic - there's some other logic required. I'm about to commit a change that introduces more data tests to highlight the problem, so if you want to take a second swing at this problem, feel free.

PR #152, opened 2016-06-04
Enable Bytes class via LOAD_CONST Python opcode translation

katlings:
:v:


freakboy3742:
Due to the problems we've been having with Travis, I merged this externally; it was committed in 236521f.

PR #155, opened 2016-06-04
Fixing typo for Hello World

michellejl:
Changed a line from "Hello, World" to "Hello World!" so that it matches what is in the print function earlier in the example.


phildini:
:ship:, looks good to me!

PR #160, opened 2016-06-05
Initial work on filter & iterables

Hwesta:
* Filter tests take 2 arguments instead of 1
* `filter` & `iter` raise correct exceptions for non-Iterables
* Test set/frozenset/dicts on functions with 2 args
* Convert all functions that check for an iterable to just call `iter`, so the TypeError is always raised in the same place
* Bug: Memory addresses can be more than 8 characters

Refs #76


Hwesta:
Found some issues with sets & dicts - this needs a bit more work before merging.

Hwesta:
Updated for dict & set changes - also implemented iter-related stuff.

PR #161, opened 2016-06-05
Implement float.hex()

encukou:
I need a bit more insight into what's going on in float-land before I implement `float.__repr__`. So, I implemented `float hex()` to tell what the bits are.


phildini:
I think this looks good! Thanks for adding the test.

PR #170, opened 2016-06-14
Implementation of builtin method `sum`.

sjoerdjob:
Should I also add extra testcases for `builtin_sum`, or are the default testcases good enough?

### DIFF
--- a/python/common/org/Python.java
+++ b/python/common/org/Python.java
@@ -1480,10 +1480,30 @@ public class Python {
             "\n" +
             "Return the sum of an iterable of numbers (NOT strings) plus the value\n" +
             "of parameter 'start' (which defaults to 0).  When the iterable is\n" +
-            "empty, return start.\n"
+            "empty, return start.\n",
+        args = {"iterable"},
+        default_args = {"start"}
     )
-    public static org.python.Object sum() {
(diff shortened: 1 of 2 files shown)


sjoerdjob:
(see also Issue #76)

freakboy3742:
Some additional test cases would certainly be a good idea - however, I'll merge what you've got here on the basis that it looks like it should work as is. If you want to provide a follow up pull request that contains a few tests for sum, I'll gladly merge that, too.

sjoerdjob:
I was happy to oblige, see pull requests #172 and #173.

zeerorg:
@sjoerdjob @freakboy3742 a very basic mistake "value" is used for storing start index and sum 
```
Test Case:
print(sum([1,2,3,4])) # output : 10
print(sum([1,2,3,4], 2)) # output : 12
```

eliasdorneles:
@zeerorg `start` is not an index, it's the initial value for the sum and that's expected behavior, see the docstring:

```
>>> sum?
Docstring:
sum(iterable[, start]) -> value

Return the sum of an iterable of numbers (NOT strings) plus the value
of parameter 'start' (which defaults to 0).  When the iterable is
empty, return start.
```

Also, when you think you found a bug, please create a new issue, don't comment in old ones.

PR #172, opened 2016-06-15
Respect @expectedFailure when processing `not_implemented`.

sjoerdjob:
During the processing of `not_implemented`, the value of
`__unittest_expecting_failure__` was enforced, instead of marking it
when not already set.

### DIFF
--- a/tests/utils.py
+++ b/tests/utils.py
@@ -535,7 +535,10 @@ class NotImplementedToExpectedFailure:
         if self._testMethodName in getattr(self, 'not_implemented', []):
             # Mark 'expecting failure' on class. It will only be applicable
             # for this specific run.
-            self.__unittest_expecting_failure__ = True
+            method = getattr(self, self._testMethodName)
+            wrapper = lambda *args, **kwargs: method(*args, **kwargs)
+            wrapper.__unittest_expecting_failure__ = True
+            setattr(self, self._testMethodName, wrapper)
         return super().run(result=result)
 
 
--- a/tests/utils.py
+++ b/tests/utils.py
@@ -528,6 +528,17 @@ class TranspileTestCase(TestCase):
         return out
 
 
+class NotImplementedToExpectedFailure:
+    def run(self, result=None):
+        # Override the run method to inject the "expectingFailure" marker
+        # when the test case runs.
+        if self._testMethodName in getattr(self, 'not_implemented', []):
+            # Mark 'expecting failure' on class. It will only be applicable
+            # for this specific run.
+            self.__unittest_expecting_failure__ = True
+        return super().run(result=result)
+
+
 SAMPLE_DATA = {
     'bool': [
             'True',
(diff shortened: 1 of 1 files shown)


tests/utils.py
```diff
@@ -649,7 +649,7 @@ def run(self, result=None):
         for test_name in dir(self):
             if test_name.startswith('test_'):
                 expect_failure = hasattr(self, 'not_implemented') and test_name in self.not_implemented
-                getattr(self, test_name).__dict__['__unittest_expecting_failure__'] = expect_failure
+                getattr(self, test_name).__dict__.setdefault('__unittest_expecting_failure__', expect_failure)
```
freakboy3742: I think the fix you're looking for here is to correct the evaluation of expect_failure, rather than using setdefault on `__unittest_expecting_failure`. The implementation provided here would do; even that could be simplified to `expect_failure = getattr(self, 'not_implemented', False)`.

sjoerdjob: The failure was the result of copy-pasting the solution from other places (Binary..., Builtin...), while the Unary case sometimes has no `not_implemented`. I think `setdefault` was the least invasive solution, because on other places there is no `expect_failure` intermediary.

Of course, could also write

    expect_failure = getattr(getattr(self, test_name), '__unittest_expecting_failure__', False) or test_name in getattr(self, 'not_implemented', [])
    getattr(self, test_name).__dict__['__unittest_expecting_failure__'] = expect_failure

which looks a 'tad' convoluted (especially the `expect_failure`).

Anyway, after re-thinking it a bit, I think it might be better to replace the custom `run` re-implementations with an extra Mixin class `NotImplementedToExpectedFailure` or something. That way it is more reasonable to get away with a bit more messy code, as it's really nicely moved away to its own class. However, in that case I'm not quite sure if it would be better to have all separate test-cases inherit from the mixin, or to make `UnaryOperationTestCase` extend from the mixin.

freakboy3742: The thing is, `setdefault` doesn't address the problem here. From the docs for `setdefault`:
```
D.setdefault(k[,d]) -> D.get(k,d), also set D[k]=d if k not in D
```

In this case, the problem isn't whether `__unittest_expecting_failure__` is already set; it's whether the value we're using exists or not. Hence, the fix is `getattr` on `not_implemented`, rather than worrying about `__unittest_expecting_failure__`.

sjoerdjob: The `getattr` is already being done on line 651 (actually: `hasattr` and using it).

Anyway, I see Travis-CI also reports a failure.

I'm preparing a pull-request which does the refactoring I was thinking about, but first running the tests locally to make sure no unexpected failures pop up.

sjoerdjob: Actually, after some further research, the value of `getattr(self, test_name).__dict__` is shared across all classes. The proper way to mark the test as 'expecting failure' in this case is by marking `__unittest_expecting_failure__` on the instance of the `TestCase`, based on the `TestCase()._testMethodName`.

I'm running a comparison of results on master and the branch to make sure they produce the same results before pushing again.

sjoerdjob:
This has a lot less "unexpected success" results than previous tries, because it now handles all the cases correctly.

The last remaining "unexpected success" are where the dictionary iteration order (or set iteration order) influences the results, which also happens in on master.

sjoerdjob:
Funnily enough, it turns out setting `__unittest_expecting_failure__` on the `TestCase` instance only works in 3.4.4, not 3.4.2.

freakboy3742:
Fantastic - that's a nice looking cleanup. 

The dictionary/set ordering thing is a known problem. I've added *some* fixes to compensate for this, but for the most part, it comes down to constructing test cases that aren't subject to ordering problems - either by adding an explicit sorting step, or by doing tests that explicitly avoid ordering problems. If you've got any other suggestions, I'd love to hear them.

PR #175, opened 2016-06-16
Replace comments with @expectFailure

sjoerdjob:
So that it shows up as a TODO during testing.

### DIFF
--- a/tests/structures/test_for.py
+++ b/tests/structures/test_for.py
@@ -1,3 +1,5 @@
+from unittest import expectedFailure
+
 from ..utils import TranspileTestCase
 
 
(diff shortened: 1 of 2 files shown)


sjoerdjob:
I've checked why Travis-CI marked the commit as failing, but I don't really understand it. No "FAIL" in the output, but setup.py exited with exit code 1. Any clue as to why?

cflee:
@sjoerdjob there is 1 unexpected success, which overall fails the test suite:

```
test_dict (tests.builtins.test_print.BuiltinPrintFunctionTests) ... unexpected success
```

freakboy3742:
Dict ordering *shouldn't* be a problem any more - as of 54aa78c, dictionary output should be being normalised (see https://github.com/pybee/voc/blame/master/tests/utils.py#L618). I would have thought the output would have been caught and normalized in this case, but evidently not...

freakboy3742:
@sjoerdjob I've had a closer look at this, and I'm almost certain that if you merge with master, this problem will go away. The code that is currently in your branch has two typos in the output substitutions - one on [Line 622](https://github.com/sjoerdjob/voc/blob/22c48148f23e614dae3571400495d3ba391fa420/tests/utils.py#L622) and one on [Line 626](https://github.com/sjoerdjob/voc/blob/22c48148f23e614dae3571400495d3ba391fa420/tests/utils.py#L626). These are both dictionary substitutions, so if the dictionary ordering on Python 3.4.4 on the CI box output either of those two formats (which is entirely likely), the output normalisation wouldn't be happening. 

If you update your branch you'll also find that the test that is failing as an "unexpected success" is now marked as a pass, whereas on your branch, it is [still in the `not_implemented` list](https://github.com/sjoerdjob/voc/blob/22c48148f23e614dae3571400495d3ba391fa420/tests/builtins/test_print.py#L16). By correcting the substitution pattern, the `dict` test for `print` now works, which it didn't previously due to dictionary ordering.

sjoerdjob:
@freakboy3742 I just rebased, so within ~40 minutes Travis-CI hopefully will say 'ok'.

PR #181, opened 2016-06-28
my first commit

Dexdev08:
deleted developers.svg line as practice -- for my contribution

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -7,9 +7,6 @@ VOC
 .. image:: https://travis-ci.org/pybee/voc.svg?branch=master
     :target: https://travis-ci.org/pybee/voc
 
-.. image:: https://badges.gitter.im/pybee/developers.svg
-    :target: https://gitter.im/pybee/developers
-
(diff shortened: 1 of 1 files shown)


freakboy3742:
Thanks for the contribution, but *why* are you contributing this patch? That link is live, even though it's not an especially active forum.

Dexdev08:
noob here... errr... i was just poking around the tutorial... and literally copying the image. >.<

freakboy3742:
Ah - no problems then! The screenshot from the tutorial is from a slightly older version of the code; it's meant more as a "here's the sort of thing you should be looking for", rather than "here's a change you should make". That's probably worth mentioning in the tutorial, though. We're working on a rework of the website content at the moment, so I'll make sure to include this.

Closing this PR (since it isn't a change we want to apply). However, assuming it was a change that we *did* want to make, you've done exactly the right thing. The only improvement you could make would be the commit/PR message itself; "My first commit" doesn't really tell anyone what the change is. Something like "Removed gitter button" would be more helpful.

PR #182, opened 2016-07-03
WIP: Methods for Float

soummyaah:


soummyaah:
@freakboy3742 
Hey, please check the implementation for ne and round. 
For ne, I noticed that round works in this way that if 4.5(even) is being rounded, it would round to 4. However, if 5.5(odd) is being rounded, it would round to 6. I'm not sure if I should implement it in this way or something else.

cflee:
@soummyaah the behaviour of `round` you described is documented [in the Built-In Types docs](https://docs.python.org/3/library/stdtypes.html#numeric-types-int-float-complex) as:

> `round(x[, n])`
>
> _x_ rounded to _n_ digits, rounding half to even. If _n_ is omitted, it defaults to 0.

That's why 4.5 rounds to 4, and 5.5 rounds to 6. The key phrase is 'round half to even' – you should be able to find more resources explaining this on the internet using that.

python/common/org/python/types/Float.java
```diff
@@ -578,7 +588,22 @@ public boolean __setattr_null(java.lang.String name, org.python.Object value) {
         __doc__ = ""
     )
     public org.python.Object __round__(org.python.Object ndigits) {
-        throw new org.python.exceptions.NotImplementedError("float.__round__() has not been implemented.");
+        /*if((org.python.types.Int)ndigits != new org.python.types.Int(0)) {
```
phildini: Did you mean to include this? If it's commented out, should it be removed?

python/common/org/python/types/Float.java
```diff
@@ -169,7 +171,15 @@ public Float(double value) {
         __doc__ = ""
     )
     public org.python.Object __ne__(org.python.Object other) {
-        throw new org.python.exceptions.NotImplementedError("float.__ne__() has not been implemented.");
+        if (other instanceof org.python.types.Int || other instanceof org.python.types.Float ||
```
phildini: I think this looks fine for the types you're including, but in my testing you _can_ compare a string to a float.

```
>>> float(1.0) != "1.0"
True
```

So I'm super curious why the string test passes for float, given your code. Is this a python 3.4 vs 3.5 difference? (I'm on 3.5)

python/common/org/python/types/Float.java
soummyaah: Hey!
Yes, this was not implemented properly which is why I commented it out. I'll be sending in more commits soon, which would rectify this.

python/common/org/python/types/Float.java
freakboy3742: Really? Is that *with* the quotes? float(1.0) should be returning a float, which definitely isn't equal to the *string* 1.0...

freakboy3742:
@soummyaah As @cflee says, "do what CPython does" is always the preferred behavior. You may need to add some tests to make sure that this specific 'round to half' behavior is preserved.

python/common/org/python/types/Float.java
soummyaah: Hey, yes. Float and string datatypes are comparable. However, it would always result in False. I'm on Python 3.4 and it gives the same results as yours.
Also, what do you mean that the string test was passing? As far as I can see, the test_ne_str was in the not_implemented list.
I've added code to include the string datatype in my implementation now.
Thank you for pointing it out!

python/common/org/python/types/Float.java
```diff
@@ -352,39 +365,16 @@ public boolean __setattr_null(java.lang.String name, org.python.Object value) {
...
-                        result += other_val;
-                    } else if (other_val < 0 && result > 0) {
-                        // second operand is negative, ensure that result is negative
-                        result += other_val; // subtract other_val, which is negative
-                    }
-                    if (other_val < 0 && result >= 0
-                            && Double.doubleToRawLongBits(result) != NEGATIVE_ZERO_RAW_BITS) {
-                        result *= -1;
-                    }
+                    // Reference: http://stackoverflow.com/a/4412200
```
soummyaah: @freakboy3742 Should I add the reference link in this way or just summarize it to a paragraph?
Also, please review it.

python/common/org/python/types/Float.java
```diff
@@ -352,39 +365,16 @@ public boolean __setattr_null(java.lang.String name, org.python.Object value) {
...
-                        // second operand is negative, ensure that result is negative
-                        result += other_val; // subtract other_val, which is negative
-                    }
-                    // edge case where adding -0.0 to 0.0 doesn't yield the expected -0.0
-                    // do this only if it is definitely not -0.0 already
-                    if (other_val < 0.0 && result >= 0.0
-                            && Double.doubleToRawLongBits(result) != NEGATIVE_ZERO_RAW_BITS) {
-                        result *= -1;
-                    }
+                    double result = (((((double) this.value) % other_val) + other_val) % other_val);
```
phildini: I'm fairly certain this works.

```
>>> float(-1) % float(10)
9.0
>>> float(1) % float(10)
1.0
>>> (-1.0 % 10.0 + 10.0) % 10.0
9.0
>>> (1.0 % 10.0 + 10.0) % 10.0
1.0
```

but I feel like some of the "modulo" tests down below should be removed from the `not_implemented` list to prove that it works?

soummyaah: Yes, it does work for float. The test that failed was a different one, I'm fixing that right now.
Though the same method does throw an error for Int:
>>> -3.14159%-5
-3.14159
However, my code returns
-3.1415900000000008
I'm trying to fix it, but I'm not really sure how to proceed from here though.

python/common/org/python/types/Float.java
freakboy3742: Given that Stack Overflow content might change over time, yes - include the paragraph as an explanation.

The code itself looks fine; however, it needs tests. There's a very specific set of combinations of values that need to be tested, and the only way to ensure that they are *all* reliably tested is an explicit test for this behavior.

python/common/org/python/types/Float.java
freakboy3742: The test runner already has handling for this - [see the `SAMPLE_SUBSTITITIONS`](https://github.com/pybee/voc/blob/master/tests/utils.py#L635). This is a list of strings that are "known equivalents" - the substitutions will turn replace any of the items in the list with the item that is the key. You can use this to remove minor "imperfections" between test output an actual value.

freakboy3742:
This is looking pretty good - the only thing I can pick up at this point is a minor code formatting issue - `if(...)` vs `if (...)` - i.e., if and else aren't function calls, so there should be a space between the keyword and the bracket. But this is a relatively minor thing.

Is there any other reason that you think this should still be a WIP? Or is it ready for merge in your opinion?

soummyaah:
No, I wanted to send in the implementation for __pow__ as well. But I'll do that in another PR. I've updated the code in Float.py to follow conventions. You can go ahead and merge.

Also, how about adding a linter to travis? Though I'm not sure how it would run on the different Java and Python files.

freakboy3742:
I want to add a linter; the catch has been finding one that reports in a helpful fashion to GitHub. If possible, I'd like to see it report in the same way that a test failure reports - "tests passed", but "Python lint failed". If you wanted to take a few hours and see if you can hook something up, that would be very helpful.

soummyaah:
Yeah, I'll do that. I'm making an issue about it for now. But I'll work on it over the weekend.

PR #185, opened 2016-07-05
Refs #36 Implement unary operators for range

jshipley:
Implement invert, negative, positive, and bool (used for unary not) for range type
Also added a ValueError exception when the range step is 0

### DIFF
--- a/python/common/org/python/types/Range.java
+++ b/python/common/org/python/types/Range.java
@@ -122,6 +122,13 @@ public class Range extends org.python.types.Object implements org.python.Iterabl
     }
 
     @org.python.Method(
+        __doc__ = "Implement __bool__(self)."
+    )
+    public org.python.types.Bool __bool__() {
+        return new org.python.types.Bool(this.start != this.stop);
+    }
+
+    @org.python.Method(
         __doc__ = ""
     )
     public org.python.Object __invert__() {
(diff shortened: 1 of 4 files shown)


python/common/org/python/types/Range.java
```diff
@@ -118,4 +121,50 @@ public Range(org.python.Object start, org.python.Object stop, org.python.Object
         }
     }
 
+    @org.python.Method(
+        __doc__ = "Implement __len__(self)."
+    )
+    public org.python.types.Int __len__() {
```
jshipley: Should __len__ and __bool__ be returning an org.python.Object?

freakboy3742: Yes, it should - Java will actually let you get away with this, because the return type doesn't formally form part of the method prototype, but for consistency, it should return `org.python.Object`.

jshipley: OK. It looks like \_\_repr\_\_ is returning a String, and \_\_iter\_\_ is returning an Iterable. Should they also have the return type org.python.Object?

freakboy3742: `__repr__` should be returning an Object, but IIRC `__iter__` needs to return an `org.python.Iterable` (the interface, not the class) to make iteration operations work.

Issue #171, opened 2016-06-14
for/else is not supported


sjoerdjob:
After some experimenting, I ran into a case where the transpiler gives an error.

```
for k in []:
    pass
else:
    print("nobreak")
```

This is as far as I could get to a minimal (non-)working example. Replacing the `print` with `pass` makes it get optimised out by Python already, so the transpiler works.

sjoerdjob:
Never mind. I saw that there's a test-case for it, but it is commented out.  Does it make sense to replace the comment-markers with an `expectFailure` decoration instead? That way it shows up in the test-run as a todo.

freakboy3742:
Thanks for the report, and sorry for the confusion. Yes, it would probably make sense for that test to be marked as an expected failure; I've certainly done that with some other tests. If you want to submit a pull request for this, I'll gladly merge it!

Manishearth:
This was done in https://github.com/pybee/voc/commit/9877c3652206bde636b6d540d70a3eff54e7587f

jshipley:
I've spent some time looking at this, and I'd like to try to fix it. I've learned how it's supposed to work in python bytecode, I just need to learn how to put that into voc and translate it to java bytecode.

Every SETUP_LOOP has a POP_BLOCK at the end. Currently, the POP_BLOCK doesn't do much, but it should be used start the "else" clause or to jump past the "else" clause if the loop was terminated by a BREAK_LOOP.

So far, all I've done is write a few more test cases. When the behavior of SETUP_LOOP, BREAK_LOOP, and POP_BLOCK is correct, then for/else and while/else with or without break statements should all work.

sjoerdjob:
Yes, I have also spent some time looking into this, and I believe your assumptions are correct.

What I did see is that determining the full extent of the loop can be quite complicated, especially for a `while True` loop, which the current approach does not fully recognize (correctly). I would be happy to review some testcases.

freakboy3742:
Fantastic! And congratulations on jumping knee deep into bytecode wrangling :-)

The area where you'll be digging is `voc/python/utils.py`, and in particular the `ForLoop` structure, and the `find_blocks()` and `extract_command()` methods.

[...]

PR #189, opened 2016-07-14
for/else, while/else, and nested try/catch tests

jshipley:
By the time work for #171 is done, all of these tests should pass.


sjoerdjob:
One thing I noticed is that if one uses  `while not …` you get a `POP_JUMP_IF_TRUE` instead of `…FALSE`. It would be good to have at least one test for that as well. The `while 1` test only succeeds because the `if` statement also creates a `POP_JUMP_IF_FALSE`.

PR #190, opened 2016-07-16
Updated jar file to reflect new build name

sesamemucho:
I updated the getting-started and tutorial0 documents to reflect the new name of the python java jar file (dist/python-java.jar->dist/python-java-support.jar).


freakboy3742:
Thanks for catching that!

PR #192, opened 2016-07-19
Implement `map()`.

sjoerdjob:
Implemented the multi-argument map case.

### DIFF
--- a/python/common/org/python/types/Map.java
+++ b/python/common/org/python/types/Map.java
@@ -2,13 +2,13 @@ package org.python.types;
 
 public class Map extends org.python.types.Object implements org.python.Iterable {
     private org.python.Object callable;
-    private java.util.List<org.python.Iterable> iterators;
+    private org.python.Iterable[] iterators;
 
     public Map(org.python.Object callable, org.python.Object [] iterables) {
         this.callable = callable;
-        this.iterators = new java.util.ArrayList();
-        for (org.python.Object iterable: iterables) {
-            this.iterators.add(org.Python.iter(iterable));
+        this.iterators = new org.python.Iterable[iterables.length];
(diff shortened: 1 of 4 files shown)


freakboy3742:
Great stuff! The only request I have is a minor naming change; for consistency, the class should be `Map`, so that it matches the object type in Python itself. Otherwise, it looks ready to merge.

sjoerdjob:
Will do!

python/common/org/python/types/Map.java
```diff
@@ -0,0 +1,35 @@
...
+    )
+    public org.python.Object __next__() {
+        org.python.Object[] args = new org.python.Object[iterators.size()];
+        int i = 0;
+        for (org.python.Iterable iterator: iterators) {
+            args[i] = iterator.__next__();
+            assert(args[i] != null);
+            i++;
+        }
+        return ((org.python.Callable)callable).invoke(args, new java.util.HashMap());
```
sjoerdjob: Would it be better to create a cached hashmap? Or is it preferable to just create a new empty version all the time?

python/common/org/python/types/Map.java
```diff
@@ -0,0 +1,35 @@
+package org.python.types;
+
+public class Map extends org.python.types.Object implements org.python.Iterable {
+    private org.python.Object callable;
+    private java.util.List<org.python.Iterable> iterators;
```
sjoerdjob: Since the size of this does not change, would it be better to create an array (`[]`) instead of an `ArrayList`?

freakboy3742: I think you might be right - creating the array should be faster than instantiating the object.

sjoerdjob: I'll make the change. By then it should be ready to merge.

PR #195, opened 2016-07-21
Added code for float eq list

gholder:


sjoerdjob:
Is working on #122 not preferable instead, as it solves the general problem instead of this specific one?

PR #196, opened 2016-07-21
Add documentation on how to deal with method signatures.

sjoerdjob:
This took me some time to figure out, so I thought it would be good to have for other developers.

### DIFF
--- a/docs/internals/index.rst
+++ b/docs/internals/index.rst
@@ -8,3 +8,4 @@ Project Internals
    contributing
    releases
    roadmap
+   signatures
--- a/docs/internals/signatures.rst
+++ b/docs/internals/signatures.rst
@@ -1,0 +1,153 @@
+Python signatures for java-defined methods
+==========================================
+
+
+In essence, a method is not much different from a Java method. One defines a
+method as follows. Here is a simple example of a 1-argument function.
+
+.. code-block:: java
+
+    @org.python.Method(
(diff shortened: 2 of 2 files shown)


docs/internals/index.rst
```diff
@@ -8,3 +8,4 @@ Project Internals
    contributing
    releases
    roadmap
+   signatures
```
sjoerdjob: I can easily imagine you'd prefer having this somewhere else. Does anybody have a good suggestion?

docs/internals/signatures.rst
```diff
@@ -0,0 +1,153 @@
...
+                val = other;
+            }
+        }
+        return val;
+    }
+
+A function accepting keyword arguments
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Of course, sometimes you want to use keyword arguments. Unfortunately, that is
```
sjoerdjob: The `kwargs` is not really used in the codebase, so I am a bit edgy before documenting how it should be used when there is no example to look at.

docs/internals/index.rst
freakboy3742: It's as good as anywhere for now.

docs/internals/signatures.rst
freakboy3742: It's an odd edge case; I'm happy to commit this as is, and I'll flesh out the rest at some point.

PR #201, opened 2016-07-21
Error msg too verbose

CodeAmend:
The python error is ONLY "string indices must be integers"
no need for ', not " + index.typeName()'

Also deleted all unexpected successes that were a result of this fix.


CodeAmend:
I think we got it this time.

PR #210, opened 2016-07-26
List methods

soummyaah:


soummyaah:
So, I'm not sure how the test suite works here. The error currently raised by the Java implementation is : Caused by: java.lang.IllegalArgumentException: wrong number of arguments
This might be caused because I added the method to contain an argument variable whereas it didn't exist anywhere before. I'm not sure how to tackle this though.

cflee:
@soummyaah If you look at other methods within List, you'll see that there is the `args` parameter you can give to the `@org.python.Method` annotation – that's probably what you want here, since the `x` argument is required. You can look at the `@Method` definition to see what other options are possible, like for methods where there are arguments with default values.

tests/utils.py
```diff
@@ -591,6 +591,8 @@ def run(self, result=None):
             '-5',
             '-3',
             '5',
+            '1',
+            '-1'
```
freakboy3742: Is the reason for this to check if the list contains 1 and -1? I'm just wary of adding more data samples unless there's a specific reason why they're needed, as there's a performance overhead to adding more examples.

soummyaah: I added 1 to verify that the Int.__eq__(Bool) ran properly. Adding those showed me that there was an error in the implementation for __eq__ which I've now modified. 
I didn't have a specific reason for adding the -1. I'll remove that.

Issue #212, opened 2016-07-28
Args not defined for inbuilt methods


soummyaah:
For most of the methods, args have not been defined, which leads to the following problem.

5*6 is executed succesfully
However, 5.**mul**(6) throws an error. For every method, the args need to be defined.

The stack trace for:

```
# test.py
a = 4
print(a.__mul__(5))
```

Comes out to be:

```
Exception in thread "main" java.lang.IllegalArgumentException: wrong number of arguments
    at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
    at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
    at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
    at java.lang.reflect.Method.invoke(Method.java:498)
    at org.python.types.Function.invoke(Function.java:275)
    at org.python.types.Method.invoke(Method.java:44)
    at python.test.__init__.module$import(test.py:2)
    at python.test.__init__.main(test.py)
```

The operators are checked, however the underlying methods throw an error.
This arose in a conversion with @cflee 
He suggested to bridge this gap by modifying the tests/utils.py to test the math operators.

freakboy3742:
This is fairly simple to do, it's just labour intensive - if you're interested in doing this, get in touch and I can give you some pointers.

Melvie:
I'd like to take a crack at this bug ! Any advice/guidance would be awesome !

Melvie:
@soummyaah Could you provide an advice on how to start solving this problem?

eliasdorneles:
Hey @Melvie, I think I can help with this one.

So, the builtin functions are define in the file [python/common/org/Python.java](https://github.com/pybee/voc/blob/master/python/common/org/Python.java)

You can search that file for `NotImplementedError` to find the Python builtin functions that are missing implementation and find some that you'd like to implement. Basically, you'd have to write a Java version of the builtin function as it works on Python 3.4.

There is some documentation for how to write the Java functions with Python signatures here: https://voc.readthedocs.io/en/latest/internals/signatures.html

I highly recommend doing it in a test-driven manner, you can find the test files for the builtins under the [tests/builtins](https://github.com/pybee/voc/tree/master/tests/builtins) directory, you should add tests of your own for the function that you'll write.

tulikaiam:
@freakboy3742 Can you help me out how to start off with it

[...]

PR #213, opened 2016-07-28
WIP: Inbuilt mathematical operations

soummyaah:
Refs #212


cflee:
I suspect this will become as involved as #122 – need to solve the same issue of using `__add__()` vs `__radd__()`, and the subclass gets priority stuff, as defined in the [data model docs](https://docs.python.org/3/reference/datamodel.html?highlight=__mul__#object.__mul__).

goanpeca:
@soummyaah any update in this work? we need a rebase and define i this work still applies, or if it has been implemented on another PR

phildini:
Hi there! We're closing this at this point due to inactivity, but feel free to re-open if you're interested in continuing it! Thanks!

PR #219, opened 2016-08-02
Test cases to formalize CALL_FUNCTION_VAR_KW, CALL_FUNCTION_VAR and C…

aftalavera:
…ALL_FUNCTION_KW


freakboy3742:
Fantastic! Thanks for those updates. I'm looking into function calls right now as part of #218; these extra tests will be very helpful.

PR #220, opened 2016-08-02
Modified Int pow method to include modulo

soummyaah:


soummyaah:
I tried running
```
print(pow(2,3))
```
However, this throws error:
```
Exception in thread "main" org.python.exceptions.RuntimeError: java.lang.NullPointerException
	at org.python.types.Function.invoke(Function.java:291)
	at org.python.types.Function.invoke(Function.java:232)
	at python.example.__init__.module$import(example.py:14)
	at python.example.__init__.main(example.py)
```
The following piece of code however does not throw any error:
```
print(2**3)
print(pow(2,3,3))
```
Can you guide me towards solving this? I tried looking in the utils.py, but according to what I saw this error shouldn't exist.

PR #225, opened 2016-08-15
Bytes type

thefon:
Signed-off-by: Adrian Smith <fon@thefon.net>


thefon:
Failing only due to travis timeout.

freakboy3742:
Almost there - there's just the 52 unexpected successes to deal with. Delete them from the various `not_implemented` lists where they appear, and this will be good to merge!

thefon:
Ahh yes, I was looking for that.  I think many of them are in other tests
from what I am working on (Bytes) that I have indirectly made work.

Also had to hardcode all the differences between Python 3.5 (my system),
and Python 3.4 (Travis).   :(

thefon:
Yay!

PR #234, opened 2016-08-16
Init the JVM once

skimbrel:
Save some time in testing by using a single JVM rather than booting one up for every test class.

Have tested this under pytest's parallel runner and it Just Works™ because xdist relies on subprocesses, meaning the globals work as intended.

Fingers crossed this helps Travis run better?

### DIFF
--- a/tests/utils.py
+++ b/tests/utils.py
@@ -27,6 +27,10 @@ _suite_configured = False
 # Set during setUpSuite()
 _output_dir = ''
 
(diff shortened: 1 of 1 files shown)


freakboy3742:
Looks like this one won't work - it definitely speeds things up on a shorter test run, but on the whole suite (on my laptop, and on CI, it raises `java.lang.OutOfMemoryError: Permgen space`.

PR #238, opened 2016-08-16
Int imul still to be implemented

DylanGraham:
TODO: create test for __imul__ of an int


DylanGraham:
Hm, maybe I have a little more work to do:
FAILED (failures=2, expected failures=42, unexpected successes=11)

PR #241, opened 2016-08-18
Small Python 3 checking fixes.

mithro:


freakboy3742:
I like the idea; but if you put logic in the setup.py, you can't create a wheel. A better place for that logic would be in the mainline for starting the voc command line tool.

mithro:
Why does the check in setup.py prevent creating a wheel?

freakboy3742:
AFAIK, you can't have install-modifying logic in a setup.py. The setup.py isn't executed as part of a wheel install, so if there is anything install-location-dependent in your setup.py, it won't be run.

mithro:
This check just prevents you from (hopefully accidentally) running setup.py with Python 2?

When creating wheels they will already be tagged as Python 3 specific if I understand correctly?

mithro:
@freakboy3742 Friendly ping?

freakboy3742:
@mithro Apologies for the delay in addressing this - it fell of my radar.

mithro:
It fell off my radar too which is why I only pinged then :).

PR #244, opened 2016-08-23
WIP: Implement Complex function call

soummyaah:


soummyaah:
So, currently my PR has implementations for `complex(1,2)` or `complex("1")`. 
However, I'm yet to implement `complex(3+4j)`. 
As per my understanding, what voc should do is interpret the `3+4j` as a complex and wrap it in that datatype class. Now I'm kind of confused about where and how this should be implemented.

python/common/org/Python.java
```diff
@@ -501,10 +501,22 @@ public static void initializeModule(java.lang.Class cls, java.util.Map<java.lang
...
-    )
-    public static org.python.types.Complex complex() {
-        throw new org.python.exceptions.NotImplementedError("Builtin function 'complex' not implemented");
+            "This is equivalent to (real + imag*1j) where imag defaults to 0.\n",
+        default_args = {"real", "imag"}
+    )
+    public static org.python.types.Complex complex(org.python.Object real, org.python.Object imag) {
+        org.python.types.Float real_val;
+        org.python.types.Float imag_val;
+        if(real instanceof org.python.types.Str && imag != null) {
```
freakboy3742: A completely cosmetic code style thing, but since you're going to need another commit to do a final CI build - there should be a space between `if` and the opening bracket.

PR #251, opened 2016-09-15
fixed list and tuple error messages as per tests

ciacon:


freakboy3742:
@ciacon As I suspected, this didn't pass CI because of 22 test failures, all related to the error message that you've changed.

The thing that was a red flag was that you've made a substantive change to code, but no tests were modified. If you're changing an error message, I would have expected to see at least *one* test change as well.

Now, in this case, the problem is that an error message has evidently changed over time. The test servers run on Python 3.4; I'm guessing you're running on Python 3.5. We may need to modify the CI server to include Python 3.5; but along with that, we'll need to modify the test framework to account for the change in error message. There is already [precedent for this](https://github.com/pybee/voc/blob/master/tests/utils.py#L256) for other messages; an analogous change will be needed for this error message.

ciacon:
I was indeed working on python 3.5. Running py.test reduced the count of errors thrown giving me confidence to pass this on to you folks. I guess I had not gotten a full glimpse of the problem at hand.

PR #263, opened 2016-09-30
Added the `get(key, [default])` method to org.python.types.Dict

darkryder:


python/common/org/python/types/Dict.java
```diff
@@ -297,10 +297,15 @@ public void __delitem__(org.python.Object item) {
...
+        __doc__ = "",
+        default_args = {"other", "default_value"}
     )
-    public org.python.Object get(org.python.Object other) {
-        throw new org.python.exceptions.NotImplementedError("dict.get() has not been implemented.");
+    public org.python.Object get(org.python.Object other, org.python.Object default_value) {
+        try {
+            return this.__getitem__(other);
+        } catch (org.python.exceptions.KeyError e){ // allow unhashable type error to be percolated up.
+            return default_value != null ? default_value: org.python.types.NoneType.NONE;
```
freakboy3742: Ternary if syntax is a recipe for confusion :-) Roll this out as a normal if statement.

darkryder: Done. Do you need a squashed history?

freakboy3742: No need to do the squash - we can do that at the time of merge.

PR #264, opened 2016-09-30
Proof-of-concept PR to show that operators can depend on each other.

darkryder:
Going through the code, I found the same logic rewritten multiple times. From a higher perspective, it seems to me that it could be made more DRY by coupling operator logic together.

IMO this makes the code easier to read as well as easier to squash bugs. Can I get a review on the concept?

For instance:
```
# Write logic only for __eq__ and __gt__ hooks
a != b can be solved by !(a == b)
a >= b can be solved by (a > b || a == b) 
a < b can be solved by (!a>b && a != b)
a <= b can be solved by (a<b || a==b) # or even (!a>b)
```

### DIFF
--- a/python/common/org/python/types/Bool.java
+++ b/python/common/org/python/types/Bool.java
@@ -79,14 +79,13 @@ public class Bool extends org.python.types.Object {
         __doc__ = ""
     )
     public org.python.Object __le__(org.python.Object other) {
-        if (other instanceof org.python.types.Int) {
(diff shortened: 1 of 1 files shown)


freakboy3742:
I can see what you're getting at here, and from a logical perspective, you're completely correct. However, doing the checks by combination means you're doing a lot more function call, type conversions, and so on. 

If the logic for `__ge__`, `__le__` etc were complex, then using this sort of combination approach would make a lot of sense; however, in this case, it's relatively easy to get right on the first attempt, and will optimise to faster code because there are fewer calls/conversions required.

Issue #270, opened 2016-10-12
Update python version requirement to 3.5+


manparvesh:
Please update the python version requirement to 3.5+ instead of 3.4+.
It took me a lot of time to figure this out, since I had Python 3.5.2 installed on my system and still, while installing VOC, I got error message saying:
`VOC requires Python 3.4+`  
This was resolved once I changed this in the `setup.py` file.

freakboy3742:
As noted on #271, it's not quite as simple as just updating a version reference; there a fundamental differences between Python 3.4, 3.5 and 3.6.

If you're particularly invested in Python 3.5, merging the work from the master branch into the 3.5 branch and making whatever adjustments are required would be deeply appreciated.

I'll keep this ticket open as a flag for Python 3.5+ support.

manparvesh:
I understand. I would go through this again properly in that case.

brechtm:
This breaks installation of briefcase, even though my use case doesn't require VOC (creating a macOS .app)!

brechtm:
Installation from wheel is still possible. You might want to perform this check on import of voc instead.

freakboy3742:
@brechtm Thanks for the report - that's clearly an unintended side effect.

shaib:
Since #326 was merged, is this still pertinent?
Note that #326 does not provide full Python 3.5 support, not all the tests pass, but the complete breakage should no longer be there.

freakboy3742:
@shaib *This* issue can be closed; but there's at least two follow up issues - making the 3.5 test suite pass (and adding it to CI); and modifying the import process so that it doesn't break briefcase (although that might be a fix to Briefcase that is required).

PR #271, opened 2016-10-12
Update python version requirement to 3.5+

manparvesh:
As mentioned in the issue #270, I think that the python version requirement should be updated to 3.5+ instead of 3.4+. This is a minor fix, but please do correct me if I am wrong.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,6 @@
 language: python
 python:
-  - "3.4.2"
-  - "3.4.4"
+  - "3.5.2"
 install:
   - "pip install ."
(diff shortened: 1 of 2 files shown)


freakboy3742:
Thanks for the suggestion; however, as you'll see from the CI results, it isn't as simple as just "bumping the version". There are significant differences between Python 3.4, 3.5 and 3.6 bytecode, which need to be accounted for. 

For the moment, I'm focussing on Python 3.4; the long term plan is to have separate branches for each version of Python, and release versions of VOC that correspond to each Python version.

PR #272, opened 2016-10-16
Implement str.lower() and str.capitalize()

eliasdorneles:
So this adds an initial implementation for `str.lower()` and `str.capitalize()`.

Does this look good?


freakboy3742:
Ok - that failure is *really* odd - I wouldn't have expected that at all. Can you run the failing tests locally to see if you get the same problem? I wouldn't completely rule out that there's something odd going on with the build box.

eliasdorneles:
I tried running locally (Python 3.4.3) and didn't got the same failures (which seem to be all unexpected successes).
I'll dig with a bit more patience later.

PR #287, opened 2016-10-29
Add implementation of list.pop()

stummjr:
This implements `list.pop()`. Does it look good?


freakboy3742:
Looks great! Thanks @stummjr!

PR #291, opened 2016-11-03
Fix arguments handling (fixes #289)

### LINKED ISSUE #289: Incorrect behavior when giving wrong arguments
I was testing out the behavior of the method signatures before sending #288, and found some unexpected behaviors.

There is probably a better way of testing this, but I did my tests adding these two test methods to the List object:

```
    @org.python.Method(
        __doc__ = "",
        default_args = {"arg1", "arg2"}
    )
    public org.python.Object some_method(org.python.Object arg1, org.python.Object arg2) {
        System.out.printf("some_method: arg1: %s\n", arg1);
        System.out.printf("some_method: arg2: %s\n", arg2);
        return org.python.types.NoneType.NONE;
    }

    @org.python.Method(
        __doc__ = "",
        args = {"arg1"},
        varargs = "others"
    )
    public org.python.Object method_varargs(org.python.Object arg1, org.python.types.Tuple others) {
        System.out.printf("method_varargs: arg1: %s\n", arg1);
        System.out.printf("method_varargs: others: %s\n", others);
        return org.python.types.NoneType.NONE;
    }
```

And then I wrote some different combinations of passing arguments, and these two cases in which I pass unexpected arguments result in unexpected behavior:

```
li = [1, 2, 3]

print("\nli.some_method(1, 'two', 3)")
li.some_method(1, 'two', 3)

print("\nli.method_varargs()")
li.method_varargs()
```

This is the output for the above program:

```
li.some_method(1, 'two', 3)
some_method: arg1: 1
some_method: arg2: two

li.method_varargs()
Exception in thread "main" java.lang.ArrayIndexOutOfBoundsException: -1
    at java.util.ArrayList.elementData(ArrayList.java:418)
    at java.util.ArrayList.get(ArrayList.java:431)
    at org.python.types.Function.adjustArguments(Function.java:251)
    at org.python.types.Function.invoke(Function.java:320)
    at org.python.types.Method.invoke(Method.java:41)
    at python.argstest.__init__.module$import(argstest.py:7)
    at python.argstest.__init__.main(argstest.py)
```

In both cases I expected to get a `TypeError` complaining about passing wrong number of arguments, but in the first case the extra argument was silently ignored and in the second case the missing argument causes a Java runtime level exception.

eliasdorneles:
This is an attempt to fix #289 , implementing proper handling of arguments.

It's currently a work in progress, I'm creating the pull request to run the tests on Travis.

The changes are:

- [x] fix name handling for closure functions
- [x] throw TypeError for unexpected args
- [x] throw TypeError for missing args

### DIFF
--- a/voc/python/blocks.py
+++ b/voc/python/blocks.py
@@ -204,7 +204,7 @@ class Block:
                 JavaOpcodes.DUP(),
         )
 
-        self.add_str(function.name)
+        self.add_str(function.code.co_name)
 
         # Add the code object
         self.add_opcodes(
--- a/python/common/org/python/types/Function.java
+++ b/python/common/org/python/types/Function.java
@@ -1,4 +1,5 @@
 package org.python.types;
+import java.util.ArrayList;
 
 
 public class Function extends org.python.types.Object implements org.python.Callable {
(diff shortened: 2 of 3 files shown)


voc/python/blocks.py
```diff
@@ -204,7 +204,7 @@ def add_callable(self, function, closure=False):
                 JavaOpcodes.DUP(),
         )
 
-        self.add_str(function.name)
+        self.add_str(function.code.co_name)
```
eliasdorneles: @freakboy3742 for some reason, `function.name` ends up being "invoke" for closure functions, `function.code.co_name` seemed to work better.
I'm not completely sure what I'm doing here though, does this look okay?

freakboy3742: Looks good to me - the code object should be the canonical python name for the function.

eliasdorneles:
@freakboy3742 It looks like the build failed because it was too long, Travis error was: `Your test run exceeded 120 minutes. One possible solution is to split up your test run.`

freakboy3742:
Yeah - we need to do something about the timing on the test suite. I've restarted that specific build; hopefully it will finish slightly faster.

freakboy3742:
There's something *really* odd going on here. The test passed on rebuild, but is marked as a failure. I'm going to merge anyway, and hope this clears up on a fresh build.

eliasdorneles:
Nice, the merge commit passed.

Thanks, and also thanks @juliomfreitas for the help digging into this issue! :+1:

PR #295, opened 2016-11-05
WIP: Do treat exceptions in runAsPython as test pass if runAsJava raises the same exception

raphaelm:
As discussed. This probably will make **a lot** of tests fail that need to be fixed. I can't really say how many, as the test suite is still running on my laptop...


eliasdorneles:
So, I believe this was solved by https://github.com/pybee/voc/pull/344, right?
Is there something else I'm missing?

freakboy3742:
I believe you are correct, @eliasdorneles

PR #303, opened 2016-11-12
adding float*list to Java code

mddemarie:


freakboy3742:
@mddemarie I'm not sure what's going on here, but it's nothing you've done - the build is terminating for some reason after 40 minutes. I'll have to investigate further.

PR #304, opened 2016-11-13
[WIP] Attempting to speedup tests by pickling cpython output

eliasdorneles:
This is an initial attempt to try to speed up the tests, by pickling the Python output (which normally doesn't change).

It's enabled by an environment variable USE_PYOUT_CACHE, if this is present it will try to load the cache from a pickle file, and will update the cache file at the end of each test class.

Not sure if this is worth yet because of the hassle of maintaining the cache files, I'll do some more tests.

Running for `tests.datatypes.test_list.ListTests`, the time went from `8s` to `5s` (~40% speedup, encouraging!).
Running for `tests.datatypes.test_list`, the time went from `205s` to `165s` (~20% speedup, not much).

I suspect the latter case gave a lesser speedup because it's updating the whole cache pickle file each time it runs a test case, so I'd like to try next splitting the pickle file per test classes.


eliasdorneles:
Of course, to get the benefits of this in Travis, we'll have to run locally in a similar environment first, at least for most of the tests.

We'd have to find a decent way of keeping these cache files updated.

I'll appreciate any feedback. :)

eliasdorneles:
@freakboy3742 looks reenabling the verbose output still doesn't fix Travis builds timing out. =/

I haven't dug much, but we might try just running separate jobs to test each subset of the tests: https://docs.travis-ci.com/user/speeding-up-the-build/
If I understood correctly, it's okay to have several jobs, as long as they don't take too long.

cflee:
@eliasdorneles going back to verbose output doesn't help due to the overall 50 min limit:

> Travis CI has specific time limits for each job, and will stop the build and and add an error message to the build log in the following situations:

> * A job takes longer than 50 minutes on travis-ci.org

([source](https://docs.travis-ci.com/user/customizing-the-build#Build-Timeouts))

Root cause: this policy seems to be in place as far back as 2014 (https://github.com/travis-ci/travis-ci/issues/2736#issue-41559910) – but rust-lang discovered in Jul 2015 (https://github.com/travis-ci/travis-ci/issues/4521) that the 50 min timeout was not working properly on the container infrastructure which we are on. That explains the 120 min limit we saw previously; I guess Travis recently patched that bug.

I think you're right, splitting into several jobs each under 50 min is the workaround.

/cc @freakboy3742 in case you are looking to invent more new swear words 🙃

eliasdorneles:
I'm going to close this because I don't think the hassle of maintaining the pickle files is worth the little more performance it gives.

freakboy3742:
If I get a moment over the weekend, I'm going to try an experiment that is related to this. Instead of pickling, just use the disk cache: Roll out tests to the file system and check them in, along with the pyc files. 

If you don't see any feedback after this weekend ,feel free to try this experiment yourself :-)

eliasdorneles:
Haha, alright, best of luck :+1:

PR #307, opened 2016-11-14
Implemented endswith java functionality and tests for python.

henriquebraga:
Implemented endswith java functionality and tests for python.

### DIFF
--- a/tests/datatypes/test_str.py
+++ b/tests/datatypes/test_str.py
@@ -10,6 +10,30 @@ class StrTests(TranspileTestCase):
             print('Done.')
             """)
 
+    def test_endswith(self):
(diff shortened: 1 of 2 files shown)


tests/datatypes/test_str.py
```diff
@@ -10,6 +10,30 @@ def test_setattr(self):
             print('Done.')
             """)
 
+    def test_endswith(self):
+        self.assertCodeExecution("""
+            s = "abracadabra"
+            suffix = "abra"
+            print(s.endswith(end))
```
eliasdorneles: I don't think these tests are doing what you think they're doing.

The variable end isn't declared and suffix is never used, I think they're all throwing NameError.

PR #308, opened 2016-11-14
Add workaround to avoid builds timing out b/c of no output

eliasdorneles:
Hey, just trying out to add [the workaround documented here](https://docs.travis-ci.com/user/common-build-problems/#My-builds-are-timing-out) to avoid builds timing out while keeping the output terse.


eliasdorneles:
Looks like it doesn't work, closing. ;-)

PR #310, opened 2016-11-16
Splitting tests to make Travis happy

eliasdorneles:
Let's see if this will work better


eliasdorneles:
The first attempt is still building, but it'll probably fail: https://travis-ci.org/pybee/voc/builds/176356641

I've split datatypes even further, which will turn into a huge matrix, but at least the build should go faster.

eliasdorneles:
good news, it worked!
and finished in under one hour. :)

cflee:
@eliasdorneles I wonder if it might still work if 42c7f17 is reverted (i.e. `-q` is re-enabled)? Then we can see if the 10 min timeout actually requires newlines (or if travis_wait works)!

freakboy3742:
I have some concerns about this one - It's clearly an improvement (because the tests are... you know... passing :-) ) but the mechanism for running the tests is now different between "normal user" and "CI system". That means there is a risk that end-users will add a test block, but it *won't* be added to the CI system, and so the tests won't be run.

As an immediate workaround this is worth doing; but we need a better long-term solution (preferably, one that lets the test suite just run faster in the first place).

eliasdorneles:
Right, I share the same concern.

One way around it could be simply to split the tests in a smarter way.

We can customize further the test running and define parameters for the build matrix like "datatypes_scalar", "datatypes_multiple" that define a hardcoded subset of tests and then have another "datatypes_other" that would run all other datatype tests excluding from the hardcoded subsets.

freakboy3742:
I'm not sure I follow - how are these build matrices being defined and invoked?

eliasdorneles:
Basically, Travis will run the build script once for each combination of python versions and env variables.

So, it will run the build command `python setup.py test -q -s $TEST_SUITE` once for each entry in the `env` block here: https://github.com/pybee/voc/blob/master/.travis.yml#L6-L26 in each Python version.

The way it's split currently merely uses the existing tests module structure, because it was the only way I knew how to run a subset using `python setup.py test`. But we can customize the way the tests run to split in a smarter way.

eliasdorneles:
@freakboy3742 The so-called "build matrix" is just the result of the different combination versions X envs. :point_up_2:

cflee:
I think @eliasdorneles is suggesting to implement a "catchall" test runner that will run all the tests that are not run by the other jobs. It seems that will lead to even more duplication of the list of how Travis is partitioning the tests, unless it will read and parse `.travis.yml`?

---

Of course, the ideal solution is to speed up the test suite entirely, but as it continues to grow we may need to keep this sharding/partitioning approach.

I was looking at the [Knapsack rubygem](https://github.com/ArturT/knapsack) for some inspiration, it is a little overkill but I think we can take the core idea: pass a parameter to the test runner to tell it to run "x/y" _fraction_ of the tests.

* Instead of specifying specific test modules or submodules (not sure what is the correct term for `tests.datatypes.test_bool`), just define say 10 jobs in Travis, and each job has parameters `1 10`, `2 10`, `3 10`... The second test runner instance will determine what is the second 10% of the tests, and be responsible for running it.

  That is a naive approach that avoids the need to commit and update 'cache' or 'index' files, but at the expense of having every test runner job recompute the full list of tests, in order to determine which is its 10% shard of the tests. No issue with missing newly-added tests with this approach.

* If we don't mind having to update some form of cache, then we can store a list of tests, and each test runner job will just read it and find its 10% shard. There will need to be special handling to catchall all the newly added tests which are not in the cache.

* A more advanced and efficient implementation to handle tests of significantly varying durations will be to do what knapsack does, and store the estimated duration of each test, and use that to distribute work across the jobs instead of the number of tests.

At this point I'm hoping someone more familiar with the Python ecosystem chimes in to say that this has already been implemented in a library or tool!

(@freakboy3742 Should we move this discussion to a new issue, since this PR has been merged? It might get lost in ongoing activity if we do not hang on to the email threads.)

freakboy3742:
@eliasdorneles Oh - sure - I understood *that* much. I was wondering how you envisaged the test splitting working (but it sounds like that's the bit where you're `<arms waving>`and we do something`</arms waving>` :-) )

@cflee It has, and it's called py.test, and VOC already supports it :-)

The catch is that it runs multithreaded in a single process, not multiple independent processes. Unfortunately, that doesn't map to what Travis does, which is independent processes.

As for moving this discussion to another ticket - that's probably a good idea. There's at least two underlying tickets here:
1. Make the tests run faster
2. Change how tests are run to support faster parallel builds and burst traffic (such as that seen at sprints)

eliasdorneles:
Right, I should've been more specific.

Yes, @cflee, parsing `.travis.yml` is a good idea!

All we need is a way to say "DON'T run these tests", which is the bit I was handwaving I suppose. =)

We can have a special for value for TEST_SUITE like:

    - TEST_SUITE=all-the-rest

And for the test command, we can use a script like:

```
if [ $TEST_SUITE = "all-the-rest" ]; then
    # here we'd run all the tests, but the ones explicitly declared in .travis.yml
    tests_to_ignore=$(grep TEST_SUITE= .travis.yml | grep -v all-the-rest | sed 's/^.*=//' | paste -d\| -s)
    TEST_COMMAND --ignore $tests_to_ignore
else
    TEST_COMMAND $TEST_SUITE
fi
```

The problem with using `python setup.py test` as TEST_COMMAND is that it doesn't have an option to ignore tests, it only has the option to tell which tests we want to run.

I think we can do this with py.test, but haven't tried it yet.

What do you think?

PR #312, opened 2016-11-17
Added the title method to the string

martini97:


eliasdorneles:
Hey @martini97, I just noticed this implementation is missing to hande the case when the string is already in upper case:

```
>>> 'HELLO WORLD'.title()
'Hello World'
```

There is already a test for this in the method `test_case_changes`, if you uncomment the call to `s.title()` it will show this problem.

PR #313, opened 2016-11-17
Implement bitwise and for Complex class (refs #36)

smilin-desperado:


tests/datatypes/test_complex.py
```diff
@@ -20,24 +20,24 @@ class BinaryComplexOperationTests(BinaryOperationTestCase, TranspileTestCase):
...
-        'test_and_bool',
-        'test_and_bytearray',
-        'test_and_bytes',
+
+
+
+
+
+
+
```
freakboy3742: The blank lines don't need to be preserved - you can purge these.

freakboy3742:
Ah - and I can see that due to my tardiness, someone else has produced a patch that is ready to merge. I can't apologise enough for the delay - I hate it when effort like yours go to waste. I hope I haven't lost a potential contributor because of this.

smilin-desperado:
No problem 😄 Something's always bound to fall through the cracks when you have a lot on your plate. I'm a fan of the project so will definitely contribute in future where possible

PR #314, opened 2016-11-19
Modified test system to use pre-generated python test output.

freakboy3742:


eliasdorneles:
Is it just me or this is actually slower?

eliasdorneles:
Before:

selection_2016-11-19_17 59 14

selection_2016-11-19_17 59 30

After:

selection_2016-11-19_17 59 38

freakboy3742:
Yeah. It's slower :-( 

In small-scale tests on my laptop, it was up to 20% faster; however, I'm guessing we might not be getting some of the caching benefits because it's a clean checkout on every run.

Oh well - chalk this one up as a failed experiment.

eliasdorneles:
Yup, we learned, it was worth a try. :+1: 

What do you think about the [idea of splitting the tests in a "less hardcoded" way](https://github.com/pybee/voc/pull/310#issuecomment-261209869)?

PR #315, opened 2016-11-19
Added the len method to the string

martini97:
Signed-off-by: Alessandro Fernandes Martini <alessandrofmartini>

### DIFF
--- a/python/common/org/python/types/Str.java
+++ b/python/common/org/python/types/Str.java
@@ -185,11 +185,10 @@ public class Str extends org.python.types.Object {
     }
 
     @org.python.Method(
-        __doc__ = ""
+        __doc__ = "len(object)\n\nReturn the number of items of a sequence or collection."
     )
     public org.python.types.Int __len__() {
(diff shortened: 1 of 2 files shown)


python/common/org/python/types/Str.java
```diff
@@ -185,11 +185,10 @@ public boolean __setattr_null(java.lang.String name, org.python.Object value) {
     }
 
     @org.python.Method(
-        __doc__ = ""
+        __doc__ = "len(object)\n\nReturn the number of items of a sequence or collection."
     )
     public org.python.types.Int __len__() {
-        // throw new org.python.exceptions.NotImplementedError("__len__() has not been implemented.");
-        return new org.python.types.Int(this.value.length());
+	return new org.python.types.Int(this.value.length());
```
eliasdorneles: The len method is already implemented, but this is adding a docstring which is nice.

The indentation seems a bit off here.

Also, the test should be renamed (it's saying `test_title`).

PR #321, opened 2016-11-21
Add implementation for many set methods

stummjr:
This implements the following methods for sets:

- `set.copy`
- `set.difference`
- `set.discard`
- `set.intersection`
- `set.remove`
- `set.union`
- `set.difference_update`
- `set.intersection_update`
- `set.update`

### DIFF
--- a/tests/datatypes/test_set.py
+++ b/tests/datatypes/test_set.py
@@ -147,6 +147,14 @@ class SetTests(TranspileTestCase):
             print(x)
             """)
 
+    def test_update(self):
+        self.assertCodeExecution("""
+            x = {1, 2, 3}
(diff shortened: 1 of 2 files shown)


tests/datatypes/test_set.py
```diff
@@ -69,6 +69,92 @@ def test_pop(self):
...
+            y = x.copy()
+            print(x.pop() is y.pop())
+            """)
+
+    def test_difference(self):
+        self.assertCodeExecution("""
+            x = {1, 2, 3}
+            y = {3, 4, 5}
+            z = x.difference(y)
+            print(z)
```
stummjr: Alright, done. 👍

PR #322, opened 2016-11-22
Switch to py.test to split tests without hardcoding all test paths

eliasdorneles:
This makes concrete [my proposal in this comment](https://github.com/pybee/voc/pull/310#issuecomment-262117798).

This PR switches the build to py.test in order to be able to ignore tests explicitly listed in `.travis.yml` file given a special value of TEST_SUITE environment variable.

This allow us to run all tests in the build without having to hardcode all test paths.

I had to update also the TEST_SUITE values to make them point to files instead of modules.

What do you think?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,29 +3,28 @@ python:
   - "3.4.2"
   - "3.4.4"
 env:
-  - TEST_SUITE=tests.builtins
-  - TEST_SUITE=tests.datatypes.test_bool
-  - TEST_SUITE=tests.datatypes.test_bytearray
-  - TEST_SUITE=tests.datatypes.test_bytes
-  - TEST_SUITE=tests.datatypes.test_complex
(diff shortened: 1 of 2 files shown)


freakboy3742:
We've had to move to py.test for Batavia - but the upside is that CircleCI provides parallelisation. As a result, we don't have to explicitly name which tests are split where; we can just use a CircleCI plugin and say "run this 4 fold parallel". 

We should be able to use the CircleCI plugin here, too - if we drop in environment variables to simulate the parallelisation. Or, of course, we could just move everything to CircleCI :-)

eliasdorneles:
Hm, not sure if I follow.

I don't have access to project settings, so I assume you want to do some tests w/ CircleCI first before deciding between approaches, yes?

freakboy3742:
This has been resolved in a different way by #330.

PR #328, opened 2016-12-05
Remove Java-7-only syntax

shaib:
(how come It took 2 months for this to surface?)


freakboy3742:
It took 2 months because we don't (can't?) run CI for android, and the java build just uses whatever Java is installed on the CI box. I think I've go this exact patch in a local branch, where I've been trying to get Toga working on Android - but that's tied up in a bunch of other code, and has been delayed by some distractions in Batavia, conferences, etc. Apologies for the inconvenience.

PR #330, opened 2016-12-10
Enable parallel builds using the CircleCI py.test plugin

freakboy3742:


eliasdorneles:
This is pretty interesting, but I have to say I miss the ability to see test cases are running in which build jobs.

freakboy3742:
Yeah - I agree that it would be nice to have a better split of tasks, and/or a split that gave a better indication of which tests ran where. This should be possible to achieve - it's just a matter of modifying the pytest circleci plugin to collate the tests in a different way. The plugin itself isn't that complicated, and we've [got our own fork of it](https://github.com/pybee/pytest-circleci) for Python3 purposes, so if you want to try your hand at a different scheme for splitting the tests, go right ahead. 

In the interim, the benefit of knowing for sure that we definitely haven't missed any tests outweighs the costs (to my mind, anyway).

PR #331, opened 2016-12-11
Allow use of locals and globals.

freakboy3742:


glasnt:
`E   AssertionError: 'H' format requires 0 <= number <= 65535`

freakboy3742:
Yeah - I was aware of this problem; just wanted to get a feel for how widespread it was. 

The cause of the problem is IF and GOTO jumps in the bytecode that need to jump more than 65535 bytes. This isn't something that happens in normal usage, but our tests that compile lots of very large tests of comparisons etc do it semi-regularly.

There's also the issue of the 50% slowdown in the test execution time. I knew there would be a penalty; but I didn't think it would be *that* bad...

python/common/org/Python.java
```diff
@@ -1815,15 +1815,15 @@ public int compare(org.python.Object o1, org.python.Object o2) {
...
-    @org.python.Method(
-        __doc__ = "vars([object]) -> dictionary" +
-            "\n" +
-            "Without arguments, equivalent to locals().\n" +
-            "With an argument, equivalent to object.__dict__.\n"
-    )
-    public static org.python.types.Dict vars() {
-        throw new org.python.exceptions.NotImplementedError("Builtin function 'vars' not implemented");
-    }
+    // @org.python.Method(
```
glasnt: Did we want to clean up some of these commented out blocks before merging?

voc/python/blocks.py
```diff
@@ -233,10 +258,10 @@ def add_callable(self, function, closure=False):
                 JavaOpcodes.ACONST_NULL(),  # co_code
         )
 
-        # self.add_tuple(function.code.co_consts)
-        self.add_opcodes(
-                JavaOpcodes.ACONST_NULL(),  # co_consts
-        )
+        self.add_tuple(function.code.co_consts)
+        # self.add_opcodes(
```
glasnt: More commenting out comments

freakboy3742: This one is a stray, though.

PR #333, opened 2016-12-17
Threw a dryer sheet into the code.

freakboy3742:
(i.e., removed some code lint)


setup.cfg
```diff
@@ -1,10 +1,11 @@
...
-multi_line_output = 5
-not_skip = __init__.py
-known_standard_library = marshal
-known_first_party = voc
-sections=FUTURE,STDLIB,THIRDPARTY,FIRSTPARTY,LOCALFOLDER
+[flake8]
+# https://flake8.readthedocs.org/en/latest/
+exclude=settings.py,wsgi.py,manage.py,*/migrations/*
+max-complexity = 10
+max-line-length = 119
```
isagalaev: Just wonder, why not a round number? :-)

freakboy3742: Habit, mostly. IIRC, editors in the Good Ole' Days used to show a continuation mark in the 80th column, so you set the line length at 79, and if you ever saw that mark, you knew you had a problem.

setup.cfg
```diff
@@ -1,10 +1,11 @@
...
+[metadata]
+description-file = README.md
+[pep8]
```
Mariatta: `pep8` was renamed into `pycodestyle`
So.. maybe use `pycodestyle` instead? :smiley:
http://pycodestyle.readthedocs.io/en/latest/
I recall reading a discussion where it was proposed that `pep8` be removed from pypi, but I don't know what's the outcome of that.

setup.cfg
isagalaev: Make sense :-) Thanks for the explanation!

PR #340, opened 2017-01-06
Modifications required for Android support

freakboy3742:


glasnt:
Something very bad changed here because I'm seeing so many ` <class 'RuntimeError'> : java.lang.NullPointerException` and even Flake8 errors (which shouldn't be terminating failures?)

freakboy3742:
There's lots of big changes; but it looks like the only failures are in the complex type. 

This is a work in progress; I just wanted to (a) share what I was doing, and (b) get a CI sanity check from multiple Python versions.

I turned on flake8 on a few weeks back; since I was at 100% flake8 passing, I would make sure we stayed there.

glasnt:
:shipit:

PR #341, opened 2017-01-09
Implemented true division on complex class

Felix5721:


freakboy3742:
Fantastic - thanks for the contribution!

Issue #320, opened 2016-11-20
Review str tests that are passing but shouldn't


eliasdorneles:
I've noticed there are some tests in tests/datatypes/test_str that are passing because they're raising the same error, but aren't doing what they're supposing to be doing.

There are also some tests that can be simplified because they have the same setup code (use the same basic string, but several calls to assertCodeExecution).

I'm adding this issue as a reminder for myself to review and cleanup them (can't do it right in this moment).

I think it would be nice to add an optional parameter `expected_output` to `TranspileTestCase.assertCodeExecution` to be checked w/ Python and Java output, to ensure that it's what we expect.

PR #343, opened 2017-01-13
Fix immutability of tuples

### LINKED ISSUE #327: Use Array instead of List in Tuple type
Current `org.python.types.Tuple` uses a `List<Object>` as its internal implementation; takes a List as its argument constructor and stores a reference to it.

This is sub-optimal, because a Python tuple is immutable. An Array would avoid dragging in all the resize mechanisms.

But further, it is incorrect, because a Python tuple should not change when the value used to construct it changes.

cflee:
Return a TypeError for all tuple item assignment attempts, regardless of whether the index used is valid or not.

Fixes #327 indirectly (fixing immutability instead of changing the internal type)

### DIFF
--- a/tests/datatypes/test_tuple.py
+++ b/tests/datatypes/test_tuple.py
@@ -79,9 +79,18 @@ class TupleTests(TranspileTestCase):
     def test_immutable(self):
         self.assertCodeExecution("""
             x = (1, 2)
-            x[0] = 3
-            x[-1] = 3
-            x[2] = 3
+            try:
+                x[0] = 3
+            except TypeError as err:
+                print(err)
+            try:
+                x[-1] = 3
+            except TypeError as err:
+                print(err)
+            try:
+                x[2] = 3
+            except TypeError as err:
+                print(err)
             """, run_in_function=False)
 
 
--- a/tests/datatypes/test_tuple.py
+++ b/tests/datatypes/test_tuple.py
@@ -76,6 +76,14 @@ class TupleTests(TranspileTestCase):
                     print(p1, "<", p2, ":", p1 < p2)
             """, run_in_function=False)
 
+    def test_immutable(self):
+        self.assertCodeExecution("""
+            x = (1, 2)
+            x[0] = 3
+            x[-1] = 3
+            x[2] = 3
+            """, run_in_function=False)
+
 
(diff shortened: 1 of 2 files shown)


tests/datatypes/test_tuple.py
```diff
@@ -76,6 +76,14 @@ def test_lt(self):
                     print(p1, "<", p2, ":", p1 < p2)
             """, run_in_function=False)
 
+    def test_immutable(self):
+        self.assertCodeExecution("""
+            x = (1, 2)
+            x[0] = 3
+            x[-1] = 3
+            x[2] = 3
+            """, run_in_function=False)
```
freakboy3742: This is definitely the right set of things to be testing, and the test will pass, but for the wrong reasons.

The test suite runs the sample code in CPython, and then again in VOC, and verifies that both examples produce the same output. This includes *if the example raises an exception*. So, this code will raise an exception on the `x[0]` assignment... and then *stop running*, because an exception was raised. The `x[-1]` and `x[2]` cases will not be tested.

So - the test needs to be modified to *catch* any anticipated exceptions, print a confirmation message if they *are* raised, and print a *failure* message if they are not. Then, the test will pass/fail depending on whether the confirmation messages were received and printed.

cflee: Right, I was checking Python's behaviour in the interactive console, so it didn't register that execution would stop because an exception is raised.

I've just pushed a commit to add separate try-except blocks for each line, and print the error out to make sure the error text matches too. I'm not sure if there's a more elegant way to do this...

cflee:
I guess the general issue with the test is similar to #320: accidentally getting exceptions (and having execution stop).

One fix could be to make TranspileTestCase append some print statement at the end of all code execution blocks, so that it's clear when any exception is unhandled. Code that expects exceptions would have to catch and preferably print the exception.

There might be a performance hit to the test suite though, since many operation tests raise exceptions and we'd have to make BinaryOperationTestCase etc wrap things in a raise/except block.

freakboy3742:
I've had the same idea - I just haven't got around to implementing it. I was thinking that the test could just append `print("===end of test===")` to the very end of whatever content is in the main test script; and `assertCodeExecution` does a verification that the last line of content is `"===end of test==="`. If that output isn't observed as the last thing, then you can assert that the test terminated early.

PR #344, opened 2017-01-15
Detect broken tests which are passing via unexpected exceptions

### LINKED ISSUE #320: Review str tests that are passing but shouldn't

cflee:
Modify test suite to catch unexpected exceptions and other early exits that cause the test output to terminate earlier than expected. This causes false positives, as tests which raise identical exceptions in Python and Java environments will pass.

The test suite now appends a `print("===end of test===")` statement to all code blocks passed to `assertCodeExecution()`, and checks for that canary value at the end of the Python and Java output. This is done after the `adjust()` stage, so that the print statement is not corrupted by the naive whitespace indentation removal logic.

This can be overridden by the `exits_early=True` parameter, which allows the test to pass when the canary is missing. Conversely, the test will fail if canary value is found unexpectedly.

Refs #343: discussion in comments after merge.
Fixes #320: as part of bulk test fixes in 1f1b7a6

Note: This PR does not cover #320's second point of tests that can be simplified, maybe that should be tracked in a separate issue? I saw something similar in `test_list.ListTests.test_index` which has very many `assertCodeExecution` calls that could be simplified.


tests/builtins/test_abs.py
```diff
@@ -8,7 +8,8 @@ class NotAbsLike:
                 pass
             x = NotAbsLike()
             print(abs(x))
-            """)
+            print('Done.')
+            """, exits_early=True)
```
freakboy3742: This should probably be a change to the test, catching and verifying that the right exception is raised.

tests/builtins/test_bin.py
```diff
@@ -12,7 +12,7 @@ def __int__(self):
 
             x = IntLike(5)
             print(bin(x))
-            """)
+            """, exits_early=True)
```
freakboy3742: As above - this is a flaw in the test.

tests/builtins/test_abs.py
freakboy3742: More generally, the "print('done')" statement that is common to many tests can now be removed, as we have a more reliable canary.

freakboy3742:
The test simplification can definitely be handled as a separate issue. If this PR just covers the protection to make sure the tests we have pass correctly, we can improve on individual tests later.

tests/builtins/test_abs.py
cflee: I can go through all the tests for `print('Done.')` and swap them out for a raise/except block, but is doing `print(err)` to let the test suite match strings sufficient to check that it's the correct exception?

cflee:
I've tried to fix most of the tests, but I'm stumped by these:

1. `tests.structures.test_function.FunctionTests.test_mainline` yields this python code:

```python
if __name__ == '__main__':
    print("Hello, world")

print("===end of test===")
```

The Java output has the canary statement first, before the "Hello, world" while the Python output is, correctly, the reverse.

2. `tests.stdlib.test_sys.SysModuleTests.test_exit`:

```python
import sys
if __name__ == '__main__':
    print(sys.exit())
    print('Done.')

print("===end of test===")
```

Java output has the canary statement while the Python output is correctly silent.

3. `tests.builtins.test_vars.VarsTests.test_simple` was recently marked as no longer expected failure, but I'm not sure whether it's a faulty test, because a KeyError is raised in both Java and Python.

cflee:
I've confirmed that `test_exit` failure is derived from the same underlying issue as `test_mainline`.

To land this large PR earlier, and since the three test failures don't seem to have immediate trivial fixes, I have marked those as `@expectedFailure`, and will track them in a new issue.

PR #345, opened 2017-01-19
Implemented Complex Power

Felix5721:
#36


Felix5721:
The failed tests are due to differences in precion of java and python when computing the power of complex number. The failed test didn't fail on my computer, maybe someone could else could test if even more test are failing on their machine?

freakboy3742:
Failures on other machines don't matter; it's only CI that ultimately matters. If there's an answer that is reliably off on other machines, then we can add that substitution in as well. So as long as you add the two extra substitutions that are needed to make the last CI run pass, this should be good to go.

freakboy3742:
Awesome work - thanks!

PR #351, opened 2017-02-03
Add implementation of with statement (closes #300)

### LINKED ISSUE #300: common patterns like `with` is missing
Seems like `with` pattern is missing/not supported yet, which is very popular and exists in Python since 2.5.
Started this as seperate issue as said by @eliasdorneles in https://github.com/pybee/voc/issues/245

eliasdorneles:
This is still a work in progress, when finished it will add support for the `with` statement fixing #300, I'm creating the PR in advance to benefit from running the tests (well, and report progress).

Things to do:

* [x] Add tests and handle cases of ctx managers with no `__enter__` or no `__exit__`
* [ ] Add test and implement capturing exception information when the body block fails
  * partially done, currently only passing the value -- getting the type and traceback isn't straightforward
* [x] Ensure that `__exit__` is called even if body fails (as long as `__enter__` succeeds)

### DIFF
--- a/voc/python/ast.py
+++ b/voc/python/ast.py
@@ -553,13 +553,35 @@ class Visitor(ast.NodeVisitor):
     def visit_With(self, node):
         # withitem* items, stmt* body):
         #     withitem = (expr context_expr, expr? optional_vars)
+        def opcodes_raise_ifnull(exception, message):
+            return [
+                JavaOpcodes.DUP(),
+                JavaOpcodes.IFNONNULL(14),  # size of IFNONNULL + exception handling opcodes
+                java.New(exception),
+                JavaOpcodes.LDC_W(message),
+                java.Init(exception, 'Ljava/lang/String;'),
+                JavaOpcodes.ATHROW(),
+            ]
+
         for it in node.items:
             self.visit(it.context_expr)
 
-            # get and invoke __enter__
+            # push __exit__ to the stack
             self.context.add_opcodes(
                 JavaOpcodes.DUP(),
-                python.Object.get_attribute('__enter__'),
+                python.Object.get_attribute('__exit__', use_null=True),
+                *opcodes_raise_ifnull('org/python/exceptions/AttributeError', '__exit__')
+            )
+            self.context.add_opcodes(
+                JavaOpcodes.SWAP(),
+            )
+
+            self.context.add_opcodes(
+                python.Object.get_attribute('__enter__', use_null=True),
+                *opcodes_raise_ifnull('org/python/exceptions/AttributeError', '__enter__')
+            )
+
+            self.context.add_opcodes(
                 JavaOpcodes.ACONST_NULL(),
                 JavaOpcodes.ACONST_NULL(),
                 python.Callable.invoke(),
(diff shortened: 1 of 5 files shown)


eliasdorneles:
I think this is ready for a first round of review.

The current implementation is not complete, because `__exit__` should get the exception type and the traceback information, which isn't straightforward to implement just yet. It's probably best postponed to be added some other time in the future, since `with` statement support is already useful even without that.

The tests are passing locally (using both Java 7 and 8), but for some reason in Travis some tests are failing with weird errors (I'd appreciate any help troubleshooting this and also running the tests in `tests/structures/test_with.py` in another machine besides mine to double check I'm not crazy =).

Many thanks to @freakboy3742 for answering my annoying questions and helping me debug bytecode! <3

eliasdorneles:
Updated, Travis tests should pass now.
The problem was that tests expected to raise exceptions were missing the `exits_early=True` introduced in https://github.com/pybee/voc/pull/344.

voc/python/ast.py
```diff
@@ -552,7 +552,121 @@ def visit_If(self, node):
     @node_visitor
     def visit_With(self, node):
         # withitem* items, stmt* body):
-        raise NotImplementedError('No handler for With')
+        #     withitem = (expr context_expr, expr? optional_vars)
+        def opcodes_raise_ifnull(exception, message):
```
freakboy3742: I see what you're doing here; however, this should be the same effect as an `IF()` pseudo-opcode. Instead of using `*opcodes_raise_ifnull()`, you should be able to use:
```
    IF([], IFNONNULL),
        Java.NEW('org/python/exceptions/AttributeError'),
        JavaOpcodes.LDC_W("__exit__"),
        Java.INIT('org/python/exceptions/AttributeError', 'Ljava/lang/String;')
        JavaOpcodes.ATHROW(),
    END_IF(),
```

Better still would be to add code to voc/python/types/java.py that allow for a simple raising of an exception:
```
    IF([], IFNONNULL),
        Java.THROW(
            'org/python/exceptions/AttributeError', 
            ['Ljava/lang/String;', JavaOpcodes.LDC_W("__exit__")],
        )
    END_IF(),
```

voc/python/ast.py
```diff
@@ -552,7 +552,121 @@ def visit_If(self, node):
...
+            """
+            return [
+                JavaOpcodes.IFNONNULL(offset=14),  # size of IFNONNULL + exception handling opcodes
+                java.New(exception),
+                JavaOpcodes.LDC_W(message),
+                java.Init(exception, 'Ljava/lang/String;'),
+                JavaOpcodes.ATHROW(),
+            ]
+
+        def array_of_nones(size):
```
freakboy3742: This could also be pulled out into the language utilities. `Java.Array` currently creates an array; but this could be extended to take a "default" argument that is rolled out as initial values if the default argument is provided:
```
    java.Array(size, default=python.None)
```

voc/python/ast.py
```diff
@@ -552,7 +552,121 @@ def visit_If(self, node):
...
+                    JavaOpcodes.DUP(),
+                    ICONST_val(i),
+                    python.NONE(),
+                    JavaOpcodes.AASTORE(),
+                ]
+            return opcodes
+
+        exit_names = ['#exit-%d-%x' % (i, id(node)) for i, _ in enumerate(node.items)]
+
+        for i, it in enumerate(node.items):
```
freakboy3742: From a code style point of view, I'd prefer `item` to `it` as a variable name.

eliasdorneles:
PR updated, thanks for the review, @freakboy3742 !

I ended up using a different argument name for the new java.Array argument (went with `fill`), because the semantics is a bit different what I'd expect for a default value (as it's not used if it's not given).

Does it look good?

PR #354, opened 2017-02-05
Implement str.center()

j-a-h-i-r:
Implemented center method for string and Fixed an extra blank line

### DIFF
--- a/python/common/org/python/types/Str.java
+++ b/python/common/org/python/types/Str.java
@@ -544,65 +544,52 @@ public class Str extends org.python.types.Object {
 
     @org.python.Method(
         __doc__ = "",
-		args = {"width"},
-		default_args = {"charToFill"}
+        args = {"width"},
+        default_args = {"charToFill"}
     )
     public org.python.Object center(org.python.Object width, org.python.Object charToFill) {
-		if(this instanceof org.python.types.Str)
-		{
-			java.lang.String fillChar = new java.lang.String();
-			if(charToFill instanceof org.python.types.Str)
-			{
-				// fillChar has right type
-				if( ((org.python.types.Str)charToFill).value.length() != 1) {
-					throw new org.python.exceptions.TypeError("The fill character must be exactly one character long");
-				}
-				fillChar = ((org.python.types.Str) charToFill).value;
-			}
-			else if(charToFill == null)
-			{
-				fillChar = " ";
-			}
-			else {
-				throw new org.python.exceptions.TypeError("Fill char must of type String");
-			}
-			
-			if(width instanceof org.python.types.Int)
-			{
-				String str = this.value;
-				int widthVal = (int)((org.python.types.Int)width).value;
-				int strLen = str.length();
-				if(strLen >= widthVal )
-				{
-					return new org.python.types.Str(str);
-				}
-				else {
-					int diff = widthVal - strLen;
-					int lenFirst = (diff)/2;
-					int lenSecond = diff - lenFirst;
-					
-					java.lang.StringBuffer returnString = new java.lang.StringBuffer(widthVal);
-					
-					for(int i=0; i<lenFirst; i++) {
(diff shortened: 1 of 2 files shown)


python/common/org/python/types/Str.java
```diff
@@ -543,10 +543,66 @@ else if (other instanceof org.python.types.Bool) {
     }
 
     @org.python.Method(
-        __doc__ = ""
-    )
-    public org.python.Object center() {
-        throw new org.python.exceptions.NotImplementedError("center() has not been implemented.");
+        __doc__ = "",
+		args = {"width"},
+		default_args = {"charToFill"}
```
freakboy3742: It's a minor thing, but our style guide is 4 character indents, as spaces.

python/common/org/python/types/Str.java
```diff
@@ -543,10 +543,66 @@ else if (other instanceof org.python.types.Bool) {
...
+    )
+    public org.python.Object center(org.python.Object width, org.python.Object charToFill) {
+		if(this instanceof org.python.types.Str)
```
freakboy3742: This branch can't ever happen - `this` will always be a Str.

python/common/org/python/types/Str.java
```diff
@@ -543,10 +543,66 @@ else if (other instanceof org.python.types.Bool) {
...
+		{
+			java.lang.String fillChar = new java.lang.String();
+			if(charToFill instanceof org.python.types.Str)
```
freakboy3742: Another style guide thing - we put a space after the if - `if (` - and use K&R style braces (i.e., `if (...) {`).

python/common/org/python/types/Str.java
```diff
@@ -543,10 +543,66 @@ else if (other instanceof org.python.types.Bool) {
...
+					int lenFirst = (diff)/2;
+					int lenSecond = diff - lenFirst;
+					
+					java.lang.StringBuffer returnString = new java.lang.StringBuffer(widthVal);
+					
+					for(int i=0; i<lenFirst; i++) {
+						returnString.append(fillChar);
+					}
+					returnString.append(str);
+					for(int i=0; i<lenSecond; i++) {
```
freakboy3742: Similar style  issue here with the for; a space after `for`, and spaces around the `=` and `<` operators.

tests/datatypes/test_str.py
```diff
@@ -276,6 +276,39 @@ def test_len(self):
...
+            try:
+                print(s.center(10,"-"))
+            except AttributeError as err:
+                print(err)
+            """)
+        # bool as width
+        self.assertCodeExecution("""
+            s = "abc"
+            print(s.center(True,"-"))
+            """)
```
freakboy3742: This is a really good set of tests; I have a couple of suggestions:

1. There's a small performance boost to be had by using a single assertCodeExecution, instead of multiple. In case of an error, The diff will make it relatively easy to work out which line was the problem (and if it doesn't, you can modify the string output to put helpful markers to help navigate)
2. There's no test of the "default" behavior - where you *don't* specify a fill character
3. There's no test of the math when the input string is empty, or when the size is divisible by 2. Given that we're dealing with halving in the implementation, checking even and odd values is helpful.

python/common/org/python/types/Str.java
j-a-h-i-r: On it.

tests/datatypes/test_str.py
j-a-h-i-r: so there should be only one assertCodeExecution call in the test_center function and I will write test in a single multiline string. did i get it right?

python/common/org/python/types/Str.java
j-a-h-i-r: I thought about this but I have very little knowledge about how it's called internally. So kept the check to be safe

PR #357, opened 2017-02-15
Added support for forelse clause

jootse84:
+ Added temporary variable '#for-orelse-%x' to set in the visit_Break.
+ Generated bytecode for the node.orelse inside visit_For.

### DIFF
--- a/voc/python/ast.py
+++ b/voc/python/ast.py
@@ -15,7 +15,7 @@ from .structures import (
 )
 from .types.primitives import (
     ASTORE_name, ALOAD_name, free_name,
-    ICONST_val,
+    ICONST_val, ISTORE_name, ILOAD_name
 )
 from .types import java, python
 from .debug import (
@@ -463,12 +463,12 @@ class Visitor(ast.NodeVisitor):
         )
 
         loop = START_LOOP()
+
         self.context.add_opcodes(
-            java.New('org/python/types/Bool'),
             JavaOpcodes.ICONST_1(),
-            java.Init('org/python/types/Bool', 'Z'),
-            ASTORE_name('#for-orelse-%x' % id(loop))
+            ISTORE_name('#loop-orelse-%x' % id(loop))
         )
+
         self.context.store_name('#for-iter-%x' % id(node))
         self.context.add_opcodes(
             loop,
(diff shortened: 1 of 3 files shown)


eliasdorneles:
Btw, this closes https://github.com/pybee/voc/issues/171

voc/python/ast.py
```diff
@@ -880,6 +896,12 @@ def visit_Break(self, node):
             if loop.end_op is None:
                 break
         self.context.add_opcodes(
+            java.New('org/python/types/Bool'),
+            JavaOpcodes.ICONST_0(),
+            java.Init('org/python/types/Bool', 'Z'),
+            ASTORE_name('#for-orelse-%x' % id(loop))
+        )
```
freakboy3742: I don't doubt that this works, but it's also going to be a redundant operation anywhere that you use a break that *isn't* part of a For, isn't it? For example, if you break from a while, it certainly won't *hurt* to have this extra bytecode, but it would be nice if it didn't have to run.

voc/python/ast.py
```diff
@@ -463,7 +463,12 @@ def visit_For(self, node):
         )
 
         loop = START_LOOP()
-
+        self.context.add_opcodes(
+            java.New('org/python/types/Bool'),
+            JavaOpcodes.ICONST_1(),
+            java.Init('org/python/types/Bool', 'Z'),
+            ASTORE_name('#for-orelse-%x' % id(loop))
```
freakboy3742: Given that this variable won't ever be accessed by the user, it would be nice to use a primitive, rather than a full object. using ISTORE_0 and ISTORE_1 to store an integer constant instead of a boolean object will be much faster.

voc/python/ast.py
jootse84: Last update from @fedeisas renames the flag to '#loop-orelse-%x' and includes support to the while/else clause.

voc/python/ast.py
jootse84: Hi! Last update from @fedeisas gets rid of the full object and stores an integer instead.

fedeisas:
@freakboy3742 since the last commit (cleanup) one of the tests started failing, but I think it's just a timeout on Travis.

PR #363, opened 2017-02-28
add missing reflected binary operations

leasunhy:


eliasdorneles:
@leasunhy Hi! Can you please add tests for these?

leasunhy:
@eliasdorneles Hi, thank you for reviewing!
Frankly speaking, I don't know where to add the tests, just like there are no tests for `__add__`, but just tests for `operator +`.
In fact, I think the `operator +` should invoke some __operator method__, say `org.python.Object.add()` which in turn tries to call `__add__()` and `__radd__()` to get the sum of the two operands.

Consider the following Python code:
```python
aa = 1
bb = complex(1, 2)
aa + bb  # (2 + 2j)
aa.__add__(bb)  # NotImplemented  NOTE: voc is not consistent here, which throws TypeError
aa.__radd__(bb)  # NotImplemented
bb.__add__(aa)  # (2 + 2j)
bb.__radd__(aa)  # (2 + 2j)
```

In other words, the `operator +` first tries to call `__add__` on its left operand, and if that fails, it tries to call `__radd__` on its right operand.

So I propose to add those __operator methods__ (i.e. `add()`, `mul()`, `sub()`, etc...) to match the behaviour of CPython.

As for the tests, I suggest either use the tests for `operator +` to test both `__add__` and `__radd__`, or separately test `__add__` and `__radd__` (i.e. add tests for both). 

Thank you!

cflee:
@leasunhy See also #213 for work on the mathematical operators, and #122 for a similar work on the comparison operators.

#122 was kinda working but feedback was the logic for selecting between e.g. `__add__` and `__radd__` should be moved into the opcode layer instead of a Java method. I'm still slowly rebasing that branch on latest master, it has moved a lot in one year.

I do think it's possible to combine e.g. `+`, `__add__()` and `__radd__()` into one test, it will avoid increasing the number of tests further for the binary operators.

leasunhy:
@cflee Thank you for the information! Didn't know about #213!

Actually I'm a bit confused about what it means to "do the selection on opcode layer"... Does it mean to select between `__add__` and `__radd__` at compile time? I don't think that's feasible... After all, you don't know which of the two methods should be selected before you actually evaluate them, no?

IMHO, I think a static method e.g. `org.python.types.Object.add(lhs, rhs)` for each operator type would be easy to implement, clear to read and efficient to run.

Thanks!

cflee:
@leasunhy Sorry, I was not clear. That approach means that VOC will generate the bytecode to select between `__add__` and `__radd__` inline wherever the `+` operator is used, so that that bytecode is executed at runtime.

I guess @freakboy3742 will have to decide whether the static method is appropriate :) since there is a runtime cost to one more method invocation.

freakboy3742:
@leasunhy Python's method dispatch is dynamic - by very definition, this means that determining which add implementation to use is a runtime behavior. This *could* be wrapped up into a static method - but you'd be just as well served by putting it into the implementation of the opcode, which means you avoid the function call, too.

leasunhy:
@freakboy3742 Thank you for your reviewing!

Yes, I understand that putting the determination logic in line would avoid a function call for each operator, but please consider that, for each use of operator, we need to put the **try-catch logic** (which should be the only way to determine which of `__add__` and `__radd__` would return a result; see the tail of this post for an example) __in line__. That certainly avoids a function call, but also leads to the [__code bloat__](https://en.wikipedia.org/wiki/Code_bloat) problem as well. And it means the code would be less cache-friendly and would also harm the performance.

Also, IMHO, as for the performance concern, we may later implement a generic (i.e. not only for the operators) optimizer which analyzes the code and if appropriate, in-lines the function.

To sum up, I think the static method approach is more concise and more flexible, and it should cause little harm to the performance (if any).

An example of `__add__` and `__radd__` selection (which would suffice as the body of the `add` static method):
```python
try { return this.__add__(other); }
catch (org.python.exceptions.NotImplementedError e) {}
try { return other.__radd__(this); }
catch (org.python.exceptions.NotImplementedError e) {}
throw new org.python.exceptions.TypeError("unsupported operand type(s) for +: 'int' and '" + other.typeName() + "'");
```

freakboy3742:
I know what you're getting at - but in this case, the code bloat may actually work to the advantage of the JVM - the JIT will be able to optimise code branch paths that is knows won't execute, but AFAIK, it won't be able to do that if the branch is in a function.

There's one easy way to resolve this argument, though: Show me a benchmark. Write up a simple "add lots of numbers" test, and compare performance between an "inline opcode" version and a "utility function" version of the `__radd__` method. Compare performance between "always hits `__add__`, always hits `__radd__`, and alternatively hitting each option. If it turns out the function is faster, we'll go that way.

leasunhy:
@freakboy3742 Working on a benchmark. Found some interesting issue (refs #406) along the way.

leasunhy:
The benchmark is not done because of a strange error arises from the bytecode version :-(
If anyone could help me with that, I would be appreciate...

Nevertheless, now that #122 has taken the static method approach and there is the issue mentioned in #406, I think I should also go that way. :-)

leasunhy:
@freakboy3742 Having started to write the static methods, I noticed that there would be two ways to do this:
1. Write an `add()` for `+`, a `substract()` for `-`, a `multiply()` for `*`.
2. Use the approach same as @cflee has taken in #122.

Generally speaking, if we are to optimize for performance, the first option should be better because it does not need to do `__getattribute__` and `Callable.invoke` things, but it also introduces some code repetition.

Which one do you prefer? 😄

leasunhy:
Just for a note, `Object.__add__` should throw AttributeError as well, just like `Object.__radd__`.

cflee:
I think it's worth noting that #122 ended up on static method due to the demonstrated need (another contributor kindly pointed to the cpython sources) to use that logic elsewhere in the standard library, so we needed to have the static method anyway.

As for the other comment about avoiding `Callable.invoke` – we were discussing this on gitter a few days ago (which led to issue #436 being filed), the whole getattribute and `Callable.invoke` thing _might_ be necessary anyway to ensure python-defined dunder methods are invoked properly. But that's not confirmed yet, we haven't fully investigated that issue.

goanpeca:
@leasunhy could you rebase this PR with the latest master to enable the latest CI checks.

Cheers :-)

phildini:
Hi there! We're closing this at this point due to inactivity, but feel free to re-open if you're interested in continuing it! Thanks!

PR #365, opened 2017-02-28
Accept Java 7 sources

eliasdorneles:
This is needed to build Android with `ant android` for the recent sources.

This is needed because the ancient Android ant build.xml is setting both `java.source` and `java.target` to 1.5.

I think it makes sense to keep the sources compatible to Java 7, since we can simply target the previous JVMs if we need (even though [at some point JDK 8 will probably deprecate the 1.5 target](http://openjdk.java.net/jeps/182)).

### DIFF
--- a/ant.properties
+++ b/ant.properties
@@ -27,3 +27,4 @@ out.rs.obj.absolute.dir=build
 out.rs.libs.absolute.dir=build
 out.aidl.absolute.dir=build
 out.dexed.absolute.dir=build
+java.source=1.7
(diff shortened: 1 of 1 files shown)


freakboy3742:
I take it this is a relatively recent change? At least for Android 4, Java 1.5 is the highest supported version...

eliasdorneles:
You mean for the bytecode, right?

This is only setting `java.source`, the `java.target` is still 1.5.

eliasdorneles:
Nevermind the latest change, javac doesn't even accept that, I didn't know what I was doing.

So, I need to research more to know exactly since which Android version the java 7 stuff has been working.
From some quick web search, it looks like it's KitKat (4.4), but not sure.

PR #375, opened 2017-03-01
Implemented islower() and isupper() for Str (refs #38)

SarthakSuri:
Also added test_iscase() and modified test_case_changes. (refs [#38](https://github.com/pybee/voc/issues/38))

### DIFF
--- a/tests/datatypes/test_str.py
+++ b/tests/datatypes/test_str.py
@@ -139,7 +139,7 @@ class StrTests(TranspileTestCase):
 
     def test_case_changes(self):
         self.assertCodeExecution("""
-            for s in ['hello, world', 'HEllo, WORLD', 'tomo', '']:
+            for s in ['hello, world', 'HEllo, WORLD', 'átomo', '']:
                 print(s.capitalize())
                 print(s.lower())
                 # print(s.swap())
--- a/tests/datatypes/test_str.py
+++ b/tests/datatypes/test_str.py
@@ -11,6 +11,13 @@ class StrTests(TranspileTestCase):
                 print(err)
             """)
 
+    def test_iscase(self):
+        self.assertCodeExecution("""
+            for s in ['hello, World!', 'HELLO, WORLD.', 'ello? world', '']:
+                print(s.islower())
+                print(s.isupper())
+            """)
+
     def test_join(self):
         self.assertCodeExecution("""
             print(','.join(None))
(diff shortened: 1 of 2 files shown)


tests/datatypes/test_str.py
```diff
@@ -132,7 +139,7 @@ def test_slice(self):
 
     def test_case_changes(self):
         self.assertCodeExecution("""
-            for s in ['hello, world', 'HEllo, WORLD', 'átomo', '']:
+            for s in ['hello, world', 'HEllo, WORLD', 'tomo', '']:
```
freakboy3742: The `á` character is in there for a reason - it's to check that non-ascii characters are converted correctly. You can't just delete the character to make the test pass - you have to make the test pass *with* the non-ascii character.

SarthakSuri: Sorry about that. That's what I expected but I thought that method was already implemented. I'll try to fix it.

SarthakSuri:
I took a look at what was throwing the error but it's way out of my depth. Since those methods weren't working should the test be expected to fail? Is there a way to do that without removing test cases?

freakboy3742:
If it was in the test suite originally, the code *was* working previously (or, at least, it was passing the test suite - possibly a false pass). There are ways to mark a test as a known failure, but we only do that when adding new tests, not to modify existing tests that have historically been passing.

SarthakSuri:
The error is from runAsPython()'s return statement `out[0].decode('utf8')` and is a problem in decoding from utf-8.

On a separate note the new methods aren't related to this issue. Should their tests include non-ascii characters too or are they fine? Either way I think that should be a new issue if non-ascii strings need to be supported. Thanks for the explanations.

freakboy3742:
If there's a problem with unicode handling, it's not surprising that it manifests in the `runAsPython()` output handling. However, that code has already had some testing, so if you think something needs to be changed there, you'll need to justify the change, probably with a walkthrough.

Non-ascii strings definitely need to be supported, though. Python 3 uses Unicode, and so VOC has to support Unicode too.

freakboy3742:
I should say - non-ASCII strings need to be supported, but they don't necessarily need to be supported *right now*. A partial implementation of lower() and upper() that works with ASCII-only is better than no implementation at all.

SarthakSuri:
So I got them to work on more characters (including á) but it still isn't perfect. There is a difference in how python and java convert unicode characters so there's still some issues when using u-prefixed strings and code points. (refs [#380](https://github.com/pybee/voc/pull/380))

I'll try to fix case changes next.

Issue #378, opened 2017-03-02
voc is unable to print real and imag part of datatype complex


deep110:
x = complex(1, 1) and x.real or x.imag is giving exception.

> Exception in thread "main" AttributeError: 'complex' object has no attribute 'real'
	at org.python.types.Object.__getattribute__(Object.java:288)
	at python.example.__init__.module$import(example.py:16)
	at python.example.__init__.main(example.py)

eliasdorneles:
Fixed in #379

PR #379, opened 2017-03-02
Fix AttributeError for complex.real and complex.image

deep110:
I have included tests for the issue https://github.com/pybee/voc/issues/378 and marked them as expected failures. I am currently unable to fix this issue. Any help regarding this would great. Thanks.

### DIFF
--- a/tests/datatypes/test_complex.py
+++ b/tests/datatypes/test_complex.py
@@ -1,7 +1,5 @@
 from .. utils import TranspileTestCase, UnaryOperationTestCase, BinaryOperationTestCase, InplaceOperationTestCase
 
-from unittest import expectedFailure
-
 
 class ComplexTests(TranspileTestCase):
     def test_conjugate(self):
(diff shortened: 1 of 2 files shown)


eliasdorneles:
Hi @deep110 !
To fix the `AttributeError` you see after adding this, try adding `@org.python.Attribute` to the real and imag attributes in the `Complex.java` file.
You'll then get another error that's related to the `Complex.toJava()` method not being implemented, and I _think_ you can make `toJava` simply return this.
Give it a try and good luck! :)

deep110:
Thanks @eliasdorneles. It worked like a charm. Now got a bit more understanding of internal code. :)

tests/datatypes/test_complex.py
```diff
@@ -8,6 +8,18 @@ def test_conjugate(self):
...
+        self.assertCodeExecution("""
+            x = complex(1, 1)
+            print(x.real)
+            """)
+
+    def test_imag(self):
+        self.assertCodeExecution("""
+            x = complex(1, 2)
+            print(x.imag)
+            """)
```
eliasdorneles: Could you merge these two tests into one?
They're pretty much related, and it makes the test case run slightly faster (each new test needs compiling and running code a few times).

deep110: @eliasdorneles Merged the tests.

PR #380, opened 2017-03-02
Added isalpha(), isdecimal(), isspace(), and added non-ascii support for lower() and upper(), in Str (refs #38)

SarthakSuri:
Tests pass for non-ascii characters (such as á) on these and on upper() and lower(). (refs [#38](https://github.com/pybee/voc/issues/38))

Also added incomplete implementations of isalnum() and isdigit(). Works for most use cases but need to return true for unicode characters such as "\u00B2" or ². isnumeric() wasn’t implemented as it also needs to return true for characters such as "\u00BD" or ½.  [(reference)](http://stackoverflow.com/a/22789660/)


SarthakSuri:
A few things broke. I think it was somewhere in the tests. I'll fix this on a fresh fork.

Issue #382, opened 2017-03-03
Empty return statement/statements in between the code is not causing any effect


Deepanshu2017:
When I run this code 

```
def foo():
    print("Hello, World!")
    return
    print("Hello, World! Again")


if __name__ == "__main__":
    foo()
```

Which a valid python code with voc I am getting results as,

```
Hello, World!
Hello, World! Again
```

and when I tried below code
```
def add(a, b):
    return a + b

def foo():
    print("Testing")
    return
    c = add(4, 5)
    print("4 + 5 =", c)
    return c


if __name__ == '__main__':
    ret = foo()
    print('return value is:', ret)
```

I got
```
Testing
4 + 5 = 9
return value is: 9
```

Even this is also not working

```
def foo():
    print("test")
    if 1 == 1:
        return
    return
    return
    return
    print("Should not print")


if __name__ == '__main__':
    foo()
```

Output is
```
test
Should not print
```

But surprisingly this code works
```
def foo():
    print("test")
    return None
    # Below works too
    # if 1 == 1:
        # return 1
    print("Should not print")


if __name__ == '__main__':
    foo()
```
Output is
```
test
```

In short whenever we are using any empty return statement in between methods they are not causing any effect in the code, but adding `None` or any other values make them work.

eliasdorneles:
Hey, thanks for isolating the problem and reporting!
I bumped into this issue while making my game but for some reason at the time I didn't seem to be able to reproduce.

Have you noticed some kind of pattern?
From a quick look at the examples, using `return` alone doesn't work but `return None` does -- is that it?

It may be interesting to see what's the difference between the bytecode generated by both cases!

leasunhy:
I would like to take this issue. All the codegen things look interesting to me!

eliasdorneles:
Fixed by #383 -- thank you @Deepanshu2017 and @leasunhy !

PR #383, opened 2017-03-04
Fix empty return stmt

leasunhy:
Refs #382 

* Empty return statement `return` now act as `return None`, in accordance with CPython.
* Modify the original test case for return statements to reflect the problem as described in issue #382


leasunhy:
@eliasdorneles Hi, thank you for reviewing!
Sorry for having made those mistakes :-(
Those two issues are now addressed. Please have a look now :-)
Thanks!

eliasdorneles:
@leasunhy no worries, they were minor concerns!
Good job, and thanks for contributing! :+1:

PR #386, opened 2017-03-05
put flake8 before py.test in travis ci config

leasunhy:
Hi all.

I suggest we put `flake8` before `py.test tests` in `.travis.yml`, because the style checking always completes faster than the unit tests. In this way, if some code has style issues, the travis build will be able to detect the style issues more quickly and "fail fast" so that it won't bother to run the lengthy unit testing process, and we save time by this way.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,5 +16,5 @@ install:
   - "pip install git+https://github.com/pybee/pytest-circleci"
   - "pip install flake8"
 script:
-  - "py.test tests"
   - flake8
+  - "py.test tests"


cflee:
Hmm, this doesn't seem to make it fail fast :(

See:
https://travis-ci.org/cflee/voc/builds/207890647 (py.test still runs after flake8 error)
https://github.com/travis-ci/travis-ci/issues/1066 (open issue. suggested workaround: `set -e`)

eliasdorneles:
oh, weird, I thought `set -e` was the default. :/
will try it out and send a PR soon, thanks @cflee !

leasunhy:
Oh, I thought `set -e` was the default too... Thanks @cflee @eliasdorneles !

PR #388, opened 2017-03-05
Add --version to voc cmdline

eliasdorneles:
This adds `--version` argument to the `voc` entrypoint script and adds a suffix `-dev` to the `voc.__version__` for the code in master.

This helps to quickly check which VOC version is being used and it's something we can ask to be included in bug reports.

Does this look good?

### DIFF
--- a/tests/utils.py
+++ b/tests/utils.py
@@ -60,7 +60,7 @@ def setUpSuite():
     )
 
     try:
-        out, err = proc.communicate(timeout=15)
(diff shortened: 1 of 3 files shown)


eliasdorneles:
I'm sneaking in a little change to increase the timeout (from 15s to 30s) of the pipe opened for `ant java` when starting the test suite, because I've noticed a bunch of build jobs failing just because of that.

PR #391, opened 2017-03-06
Implements some methods for dict

deep110:
Implemented methods are:-

- copy()
- iter()
- fromkeys()


deep110:
@eliasdorneles @freakboy3742 I found the reason for why TravisCI failed
> -<iterator object at 0xXXXXXXXX>
+<dict_keyiterator object at 0xXXXXXXXX>

I am returning iterator object but Cpython returns **dict_keyiterator** which I think in voc is **DictView**. Can you give me tips on how to implement it. But I think its implementation should involve mainly that DictView will hold iterator object.
Thanks!!

tests/datatypes/test_dict.py
```diff
@@ -229,6 +229,66 @@ def test_method_get(self):
...
+            print(y)
+            """)
+
+        self.assertCodeExecution("""
+            x = {42: 'Babel'}
+            y = x.copy()
+            print(x == y)
+            """)
+
+        self.assertCodeExecution("""
```
cflee: This test and the two above can be combined into a single `self.assertCodeExecution()` call – since they don't don't make any changes to x and y, they shouldn't interfere with each other.

tests/datatypes/test_dict.py
```diff
@@ -229,6 +229,66 @@ def test_method_get(self):
...
+            x = {42: 'Babel'}
+            y = x.copy()
+            print(x is not y)
+            """)
+
+        self.assertCodeExecution("""
+            x = {42: 'Babel'}
+            y = x.copy()
+            y['42'] = 'Babel'
```
cflee: I think it'd useful to modify this test (or add a new test) to change the value for an existing key in `x`, then check that the change is not present in `y`, and that the dicts aren't the same.

This could be followed by another line to similarly test adding a new key to one of the dicts.

tests/datatypes/test_dict.py
```diff
@@ -229,6 +229,66 @@ def test_method_get(self):
...
+        self.assertCodeExecution("""
+            keys = [1, 2]
+            value = 2
+            print({}.fromkeys(keys, value))
+            """)
+
+        self.assertCodeExecution("""
+            keys = (1, 2)
+            value = 5
+            print({}.fromkeys(keys, 5))
```
cflee: It looks like `value` is not being used here :) 

That said, I think this and the above two cases can be combined into one `self.assertCodeExecution()` – this current approach would work, and we do want the tests to be correct more than we want them to be fast, but each `self.assertCodeExecution()` call takes time to run so we prefer to reduce those where possible.

(I realise that some of the existing tests similarly have multiple `self.assertCodeExecution()` that aren't fully necessary, but we're working on gradually combining those where appropriate!)

tests/builtins/test_iter.py
```diff
@@ -12,9 +12,9 @@ class BuiltinIterFunctionTests(BuiltinFunctionTestCase, TranspileTestCase):
         'test_bytearray',
         'test_bytes',
         'test_class',
-        'test_dict',
         'test_frozenset',
         'test_list',
+        'test_dict',
```
cflee: This line seems to have been moved down with no real effect – could you undo this change?

tests/datatypes/test_dict.py
```diff
@@ -229,6 +229,66 @@ def test_method_get(self):
...
+            """)
+
+        # non-hashable error on key test
+        self.assertCodeExecution("""
+            keys = [[1], 2]
+            try:
+                print({}.fromkeys(keys))
+            except TypeError as err:
+                print(err)
+            """)
```
cflee: Just two small things – could you add a few lines to ensure that `dict.fromkeys()` doesn't affect an existing dict, and also that it can handle empty iterables?

python/common/org/python/types/Dict.java
```diff
@@ -357,10 +357,12 @@ public void __delitem__(org.python.Object item) {
 
     @org.python.Method(
             __doc__ = "",
-            args = {"iterable"},
-            default_args = {"value"}
+            default_args = {"iterable", "value"}
```
cflee: I see that the `iterable` arg has been made a default arg, with accompanying `iterable == null` check – is this exercised by any test? It looks like it will raise a TypeError exception, but I _think_ the only TypeError being caught in the tests is for a non-iterable, not for a missing iterable.

deep110: @cflee yes, it is because when I added iterable to args, and passed no params, it raised TypeError but error message doesn't match exactly with CPython's though they meant same, so I thought to handle `iterable=null` seperately. Should I add add test for it, or convert `iterable` back to args?

cflee: @deep110 This looks tricky. This format of TypeError message seems to come from [cpython's getargs.c unpack_stack()](https://github.com/python/cpython/blob/6f0eb93183519024cb360162bdd81b9faec97ba6/Python/getargs.c#L2365), and it looks like it can be triggered elsewhere too, not just from Dict objects.

I think what we can do is to move the specific test that fails only due to the TypeError message to a separate test like `test_fromkeys_missing_iterable`, mark it as an expected failure, and file a new issue. After that, change the implementation of `fromkeys()` back to having `iterable` as required arg, so that it's a correct implementation apart from the message for the missing `iterable` arg.

This ensures that the contributor who fixes that issue later will easily discover that it fixes this case (due to test case unexpected success), and we definitely won't have the workaround left here in `fromkeys()` even when it's not needed.

deep110: @cflee yes, this is nice solution. Fixing it soon.

Issue #186, opened 2016-07-08
Adding a linter to travis


soummyaah:
A linter needs to be added to travis, one that if possible, reports non-compliance the same way that a test failure reports - "tests passed", but "Python lint failed on file:<>"

freakboy3742:
For context: This came out of a conversation with @soummyaah on a ticket. Ideally the result wouldn't come up *as* a test failure, but as a separate pre-commit report in GitHub.

jambonrose:
Hi Russ,
I'm planning to take a crack at this (if someone is already working on this, please stop me!). I plan to:

- add [flake8](https://pypi.python.org/pypi/flake8) to `requirements/test.txt`
- fix any linting problems (the hard part)
- add a command to `.travis.yml` which runs flake8, and which exits with an error in the event the linter fails

I would also like to add [isort](https://pypi.python.org/pypi/isort).

NB: My understanding of Travis means that this will prevent any tests from running. If we want to run linting and tests, I believe we will need another service such as [code climate](https://codeclimate.com/) or [codacy](https://www.codacy.com/).

jambonrose:
Scratch the post above.

After a conversation between @freakboy3742 and @ldanilov the answer appears to be to use [pybee/beefore](https://github.com/pybee/beefore). Details forthcoming.

freakboy3742:
As @jambonrose indicated, I've been working on pybee/beefore to address this. Beefore needs a few more checks, though:
 * a java linter
 * isort
 * a spell checker for docs
 * a check that docs compile without error

There is also the task of working through all the *existing* code to get it to a lint-less state.

cflee:
Closing this issue as flake8 and checkstyle have since been added to the Travis build.

PR #396, opened 2017-03-07
Add Java code linting using checkstyle

cflee:
It's rather annoying for contributors to get their Java code nitpicked over for style bugs. We should have an automatic linter for Java code that fails the build, just like `flake8` for Python code.

This PR:
- Integrates the pretty standard [checkstyle](http://checkstyle.sourceforge.net/) tool as an ant target. v6.19 is used to allow running the tool with JDK 7, since there hasn't been any decision to require a minimum of JDK 8.

Refs #186, it seems that we now directly use `flake8` instead of going through Beefore, so I guess it's ok to integrate this separate tool into the build as well.

WIP: there are currently 700+ errors after switching off many irrelevant checks, as well as the potentially-useful LineLength and OperatorWrap checks. Most of the errors are whitespace and `{}` related items which look reasonable to enforce.


freakboy3742:
YESYESYESYEYESYSEYSEYSEYSEYSEYSYEYESYYEYSEYS. THIS!

freakboy3742:
So yes - the goal is to eventually move to BeeFore; the problem has been finding a way to host it so that it can run. But as an interim measure, this is fantastic - thanks for taking on this (not insignificant) project!

cflee:
I tried giving it a go with IntelliJ IDEA's Reformat Code function. Unfortunately, their [continuation indent setting](https://youtrack.jetbrains.com/issue/IDEA-120057) is not quite configurable enough, so it seems to be making more changes than necessary...

Some options: (if the build passes)
- We could just live with this one-off nuclear option and rely on checkstyle moving forwards.
- I can see if Eclipse or some other IDE is a little smarter about this.
- I can manually stage chunks of the proposed changes from IntelliJ IDEA, but I need to find a better git gui client first – Sourcetree for Mac has a [nasty UI bug](https://jira.atlassian.com/browse/SRCTREE-4184) that's making it rather difficult.

freakboy3742:
My inclination is the nuclear option; yes, it will be a PITA to get right - but it's also something that can be easily promoted as a first time contributor project and shared

freakboy3742:
Of course, if we can work out a .editorconfig that makes it easy for Eclipse/IDEA/other IDEs to conform to a style guide, that's all the better - but from my experience, most of the people coming to VOC aren't using an IDE - they're using a text editor like Sublime or Atom.

cflee:
On further thought, my main objection to how IDEA was formatting it is actually not correct. I was concerned about these changes:

```diff
     @org.python.Method(
 -        __doc__ = "Return repr(self)."
 +            __doc__ = "Return repr(self)."
      )
```

It looks unnecessary as the closing parentheses is on the same level as the opening line. However, the Sun/Oracle Java code conventions say [just indent 8 spaces instead](http://www.oracle.com/technetwork/java/javase/documentation/codeconventions-136091.html#248), which is admittedly easier to explain/enforce than having some special case to make these annotations look nicer.

Hence, I think this autoformatted output _should_ generally be ok.

Re: .editorconfig – that only gets us to settings like 4 spaces for indents, eof newline, trim trailing whitespace... I think there's no harm but the impact on code style isn't that large, since it can't pickup things like `for(int i=0;i<this.value.length();i++) {`.

At least with this PR there will be `ant checkstyle` that contributors can run locally, just like `flake8`.

PR #399, opened 2017-03-07
Added implementation for builtin function ascii() with tests

gEt-rIgHt-jR:
Implemented `ascii()` function.

It doesn't give correct output where ever `__repr__()` doesn't give correct output.

### DIFF
--- a/docs/internals/contributing.rst
+++ b/docs/internals/contributing.rst
@@ -100,6 +100,25 @@ Or, to run all the datatypes tests:
     $ python setup.py test -s tests.datatypes
 
 
+Running the code style checks
+-----------------------------
+
+Before sending your pull request for review, you may want to run the style checks locally.
+
+These checks also run automatically in Travis, but you will avoid unnecessary
+waiting time if you do this beforehand and fix your code to follow the style
+rules.
+
+In order to do that, first you need to install flake8::
+
+    pip install flake8
+
+Then, whenever you want to run the checks, run the following command inside the
+project's directory::
+
+    flake8 && ant checkstyle
+
+
 Working with code for Java bytecode
 -----------------------------------
 


python/common/org/Python.java
```diff
@@ -395,10 +395,30 @@ public static void initializeModule(java.lang.Class cls, java.util.Map<java.lang
...
+            "to that returned by repr() in Python 2.\n",
+        args = {"object"}
     )
-    public static org.python.types.Str ascii() {
-        throw new org.python.exceptions.NotImplementedError("Builtin function 'ascii' not implemented");
+    public static org.python.types.Str ascii(org.python.Object object) {
+        java.lang.String s = ((org.python.types.Str) object.__repr__()).value;
+        StringBuilder resultString = new StringBuilder();
+
+        for(java.lang.Character c : s.toCharArray()){
```
cflee: Minor style point as our Java code linter isn't up yet: for the `for` and `if` constructs, there should always be a space between the keyword and the left parentheses, and between the right parentheses and left curly brace. Instead of `for(...){` it should be like `for (...) {`.

tests/builtins/test_ascii.py
```diff
@@ -2,7 +2,33 @@
 
 
 class AsciiTests(TranspileTestCase):
-    pass
+    def test_ascii_string_without_unicode(self):
+        self.assertCodeExecution("""
+            print(ascii("hello world"))
+            print(ascii(chr(130)))
+            print(ascii(chr(270)))
```
cflee: I think it would be nice to at least test all the 'boundary' values – so based on the code above, we have to make sure that it changes around 127/128 and 255/256 and so on, so the test could check those values. I'm curious if it prints some of the low (`< \x20`) non-printable ASCII characters as well.

Another approach might be to just generate a long string that includes every single character from 0 to n, and then let the test suite compare the resulting `ascii(string)` output. That should ensure that every character is encoded to the appropriate format.

python/common/org/Python.java
gEt-rIgHt-jR: I'll keep this in mind, thanks for telling.

tests/builtins/test_ascii.py
gEt-rIgHt-jR: I'll add more tests, thanks for the help.

deep110:
@gEt-rIgHt-jR I think you need to rebase your branch, instead of merging.

gEt-rIgHt-jR:
Implementation and tests changed.

Issue #362, opened 2017-02-25
Setting instance attributes is failing to overwrite class attributes


eliasdorneles:
For the following program:

```
class MyClass:
    someattr = 1


print(MyClass.someattr)

inst = MyClass()
print(inst.someattr)

inst.someattr = 123
print(inst.someattr)
print(MyClass.someattr)
```

... the expected behavior in Python is:

```
$ python overwrite_class_attrs.py
1
1
123
1
```

What you get with VOC:

```
+ java -cp ../voc/dist/python-java-support.jar:. python.overwrite_class_attrs.__init__
1
1
1
1
```

It seems to be always using the class attribute and disregarding the reassignment.

eliasdorneles:
I should mention that this might also be a bug in the code that reads the attribute.

sohit-nayak:
Hi, this actually looks quite interesting.. Could anyone advice on how to start??

tbliu:
Hi, I would also like to work on this bug and am looking for advice/pointers! Thanks!

PR #400, opened 2017-03-09
fix issue#362 Setting instance attributes

deep110:
This PR fixes https://github.com/pybee/voc/issues/362. 

The main problem was in setting attributes ([here](https://github.com/pybee/voc/blob/master/python/common/org/python/types/Object.java#L368)), Local data attribute that matches with class attribute was never saved in the `Object.__dict__` if modified directly.

### DIFF
--- a/python/common/org/python/types/Object.java
+++ b/python/common/org/python/types/Object.java
@@ -366,8 +366,13 @@ public class Object extends java.lang.RuntimeException implements org.python.Obj
         if (attr == null) {
             this.__dict__.put(name, value);
         } else {
+            // if attribute is not a descriptor add it to local instance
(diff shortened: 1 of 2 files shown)


python/common/org/python/types/Object.java
```diff
@@ -362,12 +362,17 @@ public boolean __setattr_null(java.lang.String name, org.python.Object value) {
                 return false;
             }
         }
-
+
```
cflee: Minor code style point: could you remove the added leading whitespace here? And below, the `if (...){` should have a space before the left curly brace, so that it's `if (...) {`.

deep110: @cflee Updated the changes, Thanks for pointing it out.

Issue #369, opened 2017-03-01
Some expressions in list comprehensions causes runtime crash


eliasdorneles:
I'm running into an interesting issue with list comprehensions, here's the test case:

```
$ cat testrange.py 

# this works:
print([i+1 for i in range(3)])

# this also works:
print(list(i for i in range(3)))

# this crashes:
print([i for i in range(3)])
```

Running this with Python gives the expected output:

```
$ python testrange.py 
[1, 2, 3]
[0, 1, 2]
[0, 1, 2]
```

Running with VOC (java and javac 1.8, Python 3.4), I get a "tuple index out of range" error:

```
$ java -cp ../voc/dist/python-java-support.jar:. python.testrange.__init__
[1, 2, 3]
[0, 1, 2]
Exception in thread "main" IndexError: tuple index out of range
	at org.python.types.Tuple.__getitem__(Tuple.java:334)
	at org.python.types.Function.populateAttrs(Function.java:41)
	at org.python.types.Function.<init>(Function.java:153)
	at python.testrange.__init__.module$import(testrange.py:9)
	at python.testrange.__init__.main(testrange.py)
```

gEt-rIgHt-jR:
Hi, I would like to solve this issue. I want some help.
From what I have understood, the reason is that the size of [`co_const`](https://github.com/pybee/voc/blob/master/python/common/org/python/types/Function.java#L39) is zero.
Can someone tell me where it is initialized and how can I solve this?

deep110:
@gEt-rIgHt-jR I think _co_const_ is initialized [here](https://github.com/pybee/voc/blob/master/python/common/org/python/types/Function.java#L111).

gEt-rIgHt-jR:
@deep110 I think [this](https://github.com/pybee/voc/blob/master/python/common/org/python/types/Function.java#L130) constructor is used, but I don't know when it is called.

PR #401, opened 2017-03-09
fix issue list comprehensions causes runtime crash

deep110:
This PR fixes https://github.com/pybee/voc/issues/369

The problem was that, `consts` tuple has length zero and we were trying to get first index ([here](https://github.com/pybee/voc/blob/master/python/common/org/python/types/Function.java#L41)), so I added check for count.


eliasdorneles:
@deep110 can you please add a test for this in `tests/structures/test_list_comprehension.py` ?

deep110:
@eliasdorneles I have added the tests.

PR #402, opened 2017-03-10
Partial implementation of `max` builtin function. Refs #76

avikantz:
Passing for simple lists, `max([1, 2, 13, 8, 5, 3]))`; 
simple varargs, `max(1, 5, 3, 2, 4, 9, 12)`, `max("Kirk", "Scotty", "Bones", "Spock", "Sulu")`; 
tuples `max(("ackbar", "balrog"), ("klingon", "ewoks"), ("zalthor", "zathura"))`.

Failing for single strings, `max('Han Solo')`: Expected `Han Solo`, Actual: `o`.
tuples, `max((1, 2, 3, 4, 5, 6, 7, 8, 9, 10))`: Expected `10`, Actual: `(1, 2, 3, 4, 5, 6, 7, 8, 9, 10)`
string args, `max("Three args: %s | %s | %s")`: Expected: ?, Actual: `|`


eliasdorneles:
Hi @avikantz!

Have you considered using [`Collections.max`](https://docs.oracle.com/javase/7/docs/api/java/util/Collections.html#max(java.util.Collection)) ?


> Failing for single strings, max('Han Solo'): Expected `Han Solo`, Actual: `o`.

Hm, it's the other way around, `o` is the expected output, but the code is returning the entire string.

```
$ python
[GCC 4.8.4] on linux
>>> max('Han Solo')
'o'
```

python/common/org/Python.java
```diff
@@ -915,10 +915,60 @@ public static void initializeModule(java.lang.Class cls, java.util.Map<java.lang
...
     )
-    public static org.python.Object max() {
-        throw new org.python.exceptions.NotImplementedError("Builtin function 'max' not implemented");
+    public static org.python.Object max(org.python.Object first, org.python.types.Tuple rest, org.python.types.Dict kwargs) {
+        org.python.Object val = null;
+        if (first == null) {
+            throw new org.python.exceptions.TypeError("max expected 1 arguments, got 0");
+        } else {
+            org.python.Iterable iter;
+            if (first instanceof org.python.types.List) {
```
eliasdorneles: You shouldn't do type-checking, the max builtin function should work on any iterable object (namely, any object that has an `__iter__` method that returns an iterator), not just on lists/tuples/dicts/etc.

Example:

```
>>> class Deck:
...     cards = ['A'] + [str(i) for i in range(1, 10)] + ['J', 'Q', 'K']
...     def __iter__(self):
...         return iter(self.cards)
... 
>>> max(Deck())
'Q'
```

[Look here for an example of iterating over an iterable](https://github.com/pybee/voc/blob/master/python/common/org/Python.java#L1395-L1403).

avikantz: Ok, I tried iterating through the params, it's breaking at some places:
```python
max("Alpha", "Scotty", "Bones", "Spock", "Kirk")
- p
+ Spock
```

But working for similar cases:
```python
max("foo", "bar", "other")
other
```

I'm also not able to find out where a `java.util.NoSuchElementException` is produced, when an empty array is passed to as first.

eliasdorneles:
The [build failed because of a style check](https://travis-ci.org/pybee/voc/jobs/209840557#L180), can you have a look please?
Also, can you please sync the code with the current master and solve the conflicts? (either merge or rebase, whatever you're more comfortable)

avikantz:
Corrected the style check, and resolved the conflicts. Could you have a look. Thanks.

eliasdorneles:
@avikantz can you check those test failures?

avikantz:
I checked, I'm not quite sure why are they not working for some specific versions of python.

cflee:
@avikantz They seem to be the same four or five test failures repeated across the three Python versions, so I don't think it's a python version thing. You'll have to look at the diff that comes with each test failure to see where it's going wrong.

For example, in `tests.builtins.test_max.MaxTests.test_max`, the statement `print(max("Pen", "Pineapple", "Apple", "Pen"))` is printing `n` (actual output) instead of `Pineapple` (expected output).

eliasdorneles:
> I checked, I'm not quite sure why are they not working for some specific versions of python.

Note that the build is split in many jobs, so having a few green jobs doesn't mean that it passes for a Python version, it just means that a subset of the tests are passing. ;)

avikantz:
Oh, alright. I shall look into it. I labeled it partial for the same reason.

One of the issues is I'm not able to distinguish between `RuntimeError: java.util.NoSuchElementException` with `ValueError: max() arg is an empty sequence` for empty input.

goanpeca:
@avikantz could you rebase this PR with the latest master to enable the latest CI checks.

Did you manage to solve some of the errors you were having? Need some help?

Cheers :-)

cflee:
Closing, as this was implemented in #504.

PR #403, opened 2017-03-10
Update for list test_slices for VOC

flacode:
Adding tests for list slicing operations not implemented in Java.

1. test_slice_with_zero_step() to test if a value error is raised when step=0 is used for slicing and implementation in java.

2. test_slice_in_reverse() to test the use of steps less than 0 during slicing.

### DIFF
--- a/python/common/org/python/types/Slice.java
+++ b/python/common/org/python/types/Slice.java
@@ -35,8 +35,13 @@ public class Slice extends org.python.types.Object {
         }
 
         if (step instanceof org.python.types.Int) {
-            this.__dict__.put("step", step);
             this.step = (org.python.types.Int) step;
+            if (this.step.value == 0 ){
+                throw new org.python.exceptions.ValueError("slice step cannot be zero");
+            }
+            else{
+                this.__dict__.put("step", step);   
+            }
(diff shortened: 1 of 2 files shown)


python/common/org/python/types/Slice.java
```diff
@@ -35,8 +35,13 @@ public Slice(org.python.Object start, org.python.Object stop, org.python.Object
         }
 
         if (step instanceof org.python.types.Int) {
-            this.__dict__.put("step", step);
             this.step = (org.python.types.Int) step;
+            if (this.step.value == 0 ){
```
freakboy3742: There's some odd spacing around this closing brace - the space should be *after* the `)`, not before.

python/common/org/python/types/Slice.java
```diff
@@ -35,8 +35,13 @@ public Slice(org.python.Object start, org.python.Object stop, org.python.Object
...
+                throw new org.python.exceptions.ValueError("slice step cannot be zero");
+            }
+            else{
```
freakboy3742: Another spacing issue here; the else should be on the same line as the previous bracket, and there should be a space before and after else - i.e., `} else {`

PR #405, opened 2017-03-11
Added the builtin function partition() to the String type (Refs #38)

gilmouta:


tests/datatypes/test_str.py
```diff
@@ -356,6 +356,19 @@ def test_ljust(self):
                 print(err)
                 """)
 
+    def test_partition(self):
+        self.assertCodeExecution("""
+            s = "foobar"
+            print(s.partition("ob"))
+            print(s.partition("o"))
+            print(s.partition("f"))
+            print(s.partition("r"))
```
freakboy3742: This looks really good - but it's missing one obvious test - partitioning on a substring that doesn't exist.

gilmouta: Good point, thanks. Fixed.

PR #408, opened 2017-03-12
Enumerate

Saurabh--Kumar:
Hi BeeWare community. I am new to open source. This is my first pull request ever. Suggestions and comments are welcomed. 

This PR contains implementation of builtin function  enumerate(). Improvements pointed out, are made.

Saurabh Kumar


eliasdorneles:
Hello and welcome, @Saurabh--Kumar !

Please check `git diff` before you commit stuff, and make sure it's only adding the files you want to add.

This PR is introducing a bunch of `hs_err_pidXXXX.log` files that shouldn't be there -- can you please remove them?

python/common/org/Python.java
```diff
@@ -602,12 +602,45 @@ public static void initializeModule(java.lang.Class cls, java.util.Map<java.lang
...
-    public static org.python.Iterable enumerate() {
-        throw new org.python.exceptions.NotImplementedError("Builtin function 'enumerate' not implemented");
+    //public static org.python.Iterable enumerate(org.python.Iterable items, org.python.types.Int start) {
+   public static org.python.Iterable enumerate(org.python.Object items,/*org.python.types.List items,*/ org.python.Object start) {
+    	try{
+    		if(start==null)
+    			start=new org.python.types.Int(0);
+    		else {
+    			if(!(start instanceof org.python.types.Int))
+    				throw new org.python.exceptions.TypeError("integer argument expected as 'start' , got " + start.typeName() + "");
```
eliasdorneles: Note that these errors should match the errors in CPython, this looks different than what I get on Python 3.4:

```
>>> list(enumerate([], start=None))
TypeError: 'NoneType' object cannot be interpreted as an integer
```

Can you add some tests that exercise this code?

Something like:

```
def test_enumerate_invalid_start_args(self):
       self.assertCodeExecution("""
            print(list(enumerate(['a','b','c'], start=None)))
            """, exits_early=True)
       self.assertCodeExecution("""
            print(list(enumerate(['a','b','c'], start=1.5)))
            """, exits_early=True)
       self.assertCodeExecution("""
            print(list(enumerate(['a','b','c'], start=None)))
            """, exits_early=True)
```

Using `instanceof` here is probably not be the best, try adding a test for passing a boolean:

    print(list(enumerate(['a','b','c'], start=True))) # in Python, True has integer value of 1

You may try simply casting for now and then throw the proper error if casting fails.

tests/builtins/test_enumerate.py
```diff
@@ -2,7 +2,28 @@
 
 
 class EnumerateTests(TranspileTestCase):
-    pass
+      def test_enumerate(self):
+        self.assertCodeExecution("""
```
Saurabh--Kumar: apart from indentation, are these test cases written in the correct way?

Saurabh--Kumar:
hi @eliasdorneles . I have made the changes that you had suggested. Please review this updated PR and see if it is fit for merging. 

Thanks.

python/common/org/Python.java
```diff
@@ -602,12 +602,44 @@ public static void initializeModule(java.lang.Class cls, java.util.Map<java.lang
...
+            args = {"items"},
+            default_args = {"start"}
     )
-    public static org.python.Iterable enumerate() {
-        throw new org.python.exceptions.NotImplementedError("Builtin function 'enumerate' not implemented");
+    public static org.python.Iterable enumerate(org.python.Object items, org.python.Object start) {
+        try {
+            org.python.Object index = new org.python.types.Int(0);
+            if (start == null) {
+                start = new org.python.types.Int(0);
```
cflee: This statement doesn't seem very necessary because `start` isn't used after this – `index` is, and that's already initialised to 0 a few lines above this, or in the below `else` block.

Perhaps this line should be removed, and the if conditional changed to `start != null`?

python/common/org/Python.java
```diff
@@ -602,12 +602,44 @@ public static void initializeModule(java.lang.Class cls, java.util.Map<java.lang
...
                     "enumerate is useful for obtaining an indexed list:\n" +
-                    "       (0, seq[0]), (1, seq[1]), (2, seq[2]), ...\n"
+                    "(0, seq[0]), (1, seq[1]), (2, seq[2]), ...\n",
+            args = {"items"},
+            default_args = {"start"}
     )
-    public static org.python.Iterable enumerate() {
-        throw new org.python.exceptions.NotImplementedError("Builtin function 'enumerate' not implemented");
+    public static org.python.Iterable enumerate(org.python.Object items, org.python.Object start) {
+        try {
```
cflee: This `try` block is quite large, and it's not immediately obvious where the `AttributeError` could come from. I'm guessing it's for `org.Python.iter()`, so to make the code clearer by minimizing the coverage, only that and everything that depends the `iter` should be inside.

I think one way of doing this would be to move the initial index/start handling code out to before the try block, that should be the minimum.

python/common/org/Python.java
```diff
@@ -602,12 +602,44 @@ public static void initializeModule(java.lang.Class cls, java.util.Map<java.lang
...
+                    break;
+                }
+            }
+            return org.Python.iter(new org.python.types.List(enumList));
+        } catch (org.python.exceptions.AttributeError e) {
+            throw new org.python.exceptions.TypeError("'" + items.typeName() + "' object is not iterable");
+        }
     }
 
+
```
cflee: This empty newline seems unnecessary, it's usually 1 line between Java methods.

tests/builtins/test_enumerate.py
```diff
@@ -2,7 +2,55 @@
 
 
 class EnumerateTests(TranspileTestCase):
-    pass
+    def test_enumerate(self):
```
cflee: Could you add some tests here for (i) missing items arg, (ii) empty list/iterable?

Also, for this and the subsequent tests, you should merge the separate `self.assertCodeExecution()` blocks into one, such that there's one for `test_enumerate()`, one for `test_enumerate_invalid_start_args()`, etc.

Saurabh--Kumar:
@eliasdorneles Could you please suggest required changes to make this PR fit for merge?

tests/builtins/test_enumerate.py
cflee: It looks like there's now a test for missing `items` arg. Is there one for empty list/iterable?

Saurabh--Kumar: @cflee Added test for empty iterable. Regarding lazy iterable, I talked to @eliasdorneles regarding this issue, he asked me to implement the list version for the time being. I see this PR has been marked "not-quite-right". could you suggest changes to make it fit for merge?

PR #413, opened 2017-03-13
Implements the isalpha() String method

ashutoshsaboo:

This PR implements the [isalpha()](https://docs.python.org/3/library/stdtypes.html#str.isalpha) method for strings. 

Please review this PR. ~~Also, are there any tests needed before merging this PR?~~ 
Update: Added test cases for isalpha() too. @freakboy3742 @eliasdorneles  It'd great if you could review this PR. :smiley: 

Does it seem ready to merge? Thanks! :smile:

### DIFF
--- a/python/common/org/python/types/Str.java
+++ b/python/common/org/python/types/Str.java
@@ -719,12 +719,11 @@ public class Str extends org.python.types.Object {
     )
     public org.python.Object isalpha() {
         int c = 0;
-        java.lang.String checkString = this.value.toUpperCase();
+        java.lang.String checkString = this.value;
         char currentCharacter;
         while (c < (checkString.length()-1)) {
             currentCharacter = checkString.charAt(c);
-            int ascii = (int) currentCharacter;
-            if (ascii<65 || ascii>90) {
+            if (!Character.isLetter(currentCharacter)) {
                 return new org.python.types.Bool(false);
             }
             c++;
--- a/python/common/org/python/types/Str.java
+++ b/python/common/org/python/types/Str.java
@@ -721,12 +721,17 @@ public class Str extends org.python.types.Object {
         int c = 0;
         java.lang.String checkString = this.value;
         char currentCharacter;
-        while (c < (checkString.length())) {
-            currentCharacter = checkString.charAt(c);
-            if (!Character.isLetter(currentCharacter)) {
-                return new org.python.types.Bool(false);
+
+        if (checkString.length() == 0) {
+            return new org.python.types.Bool(false);
+        } else {
+            while (c < (checkString.length())) {
+                currentCharacter = checkString.charAt(c);
+                if (!Character.isLetter(currentCharacter)) {
+                    return new org.python.types.Bool(false);
+                }
+                c++;
             }
-            c++;
         }
         return new org.python.types.Bool(true);
     }
--- a/python/common/org/python/types/Str.java
+++ b/python/common/org/python/types/Str.java
@@ -721,7 +721,7 @@ public class Str extends org.python.types.Object {
         int c = 0;
         java.lang.String checkString = this.value;
         char currentCharacter;
-        while (c < (checkString.length()-1)) {
+        while (c < (checkString.length() - 1)) {
             currentCharacter = checkString.charAt(c);
             if (!Character.isLetter(currentCharacter)) {
                 return new org.python.types.Bool(false);
--- a/python/common/org/python/types/Str.java
+++ b/python/common/org/python/types/Str.java
@@ -718,7 +718,18 @@ public class Str extends org.python.types.Object {
             __doc__ = ""
     )
     public org.python.Object isalpha() {
-        return new org.python.types.Bool(this.value.matches("[a-zA-Z]+"));
+        int c = 0;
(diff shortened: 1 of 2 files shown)


python/common/org/python/types/Str.java
```diff
@@ -718,7 +718,7 @@ private long toPositiveIndex(long index) {
             __doc__ = ""
     )
     public org.python.Object isalpha() {
-        throw new org.python.exceptions.NotImplementedError("isalpha() has not been implemented.");
+        return new org.python.types.Bool(this.value.matches("[a-zA-Z]+"));
```
eliasdorneles: Can you add a test for `print('átomo'.isalpha())` ?
This prints True in CPython, and I don't think the `[a-zA-Z]+` will match for this case.
Not sure what's the proper way of doing this, but an option could be to use [Character.isLetter()](https://docs.oracle.com/javase/7/docs/api/java/lang/Character.html#isLetter(char)).

It would be nice also to have a test for `'CasesLikeTheseWithoutSpaces'`, though it should pass with the current code.

ashutoshsaboo: @eliasdorneles Thanks a lot! I'll change it, and re-push the PR.

ashutoshsaboo:
@eliasdorneles The latest commit, resolves the issue that you pointed out. Does this seem ready for merging?

eliasdorneles:
Hi @ashutoshsaboo, please run the new tests in your machine before claiming that it's fixed -- the new test is failing.

The issue I mentioned is supporting non-ASCII characters (like CPython does), you should NOT hardcode ASCII table numbers.
This is the behavior we expect:

```
>>> print('átomo'.isalpha())
True
```

Check my previous comment about using `Character.isLetter`, have you tried that?

ashutoshsaboo:
@eliasdorneles Ahh! I tried the isLetter method only first, If you'll check `print('átomo'.isalpha())` in python2.7 it prints false, and in python3 it prints true. That's the reason actually why I changed it from isLetter() to the ascii values. Mistakingly, I checked it on python2.7 before pushing, and that let to the edit. :disappointed: 
```
saboo@systemslab312-OptiPlex-3040:~$ python3
Python 3.4.3 (default, Nov 17 2016, 01:08:31) 
[GCC 4.8.4] on linux
>>> print('átomo'.isalpha())
True
>>> 
saboo@systemslab312-OptiPlex-3040:~$ python3
Python 3.4.3 (default, Nov 17 2016, 01:08:31) 
[GCC 4.8.4] on linux
>>> 
saboo@systemslab312-OptiPlex-3040:~$ python2.7
Python 2.7.6 (default, Jun 22 2015, 17:58:13) 
[GCC 4.8.2] on linux2
>>> print('átomo'.isalpha())
False
>>> 
```
Nvm, I'll re-push the commit. Thanks! :smile:

ashutoshsaboo:
@eliasdorneles Re-pushed the commit, with the isLetter() method. :+1: Thanks!

eliasdorneles:
There is a style check failing: https://travis-ci.org/pybee/voc/jobs/210690066#L180

You can run those locally before commiting, if you want to speed things up.
First, install flake8 in the virtualenv:

    pip install flake8

Then, run the style checks:

    flake8 && ant checkstyle

ashutoshsaboo:
@eliasdorneles Thanks for the resource that you mentioned. It'll help me a lot. The latest commit fixes the style error issue. :smile:

tests/datatypes/test_str.py
```diff
@@ -30,6 +30,13 @@ def test_isspace(self):
                 print(s.isspace())
             """)
 
+    def test_isalpha(self):
+        self.assertCodeExecution("""
+            for s in ['Hello World', 'hello wORLd.', 'Hello world.', '',
```
eliasdorneles: can you add a test for `hello1` ?
I think this code also has a off-by-one issue.

ashutoshsaboo: @eliasdorneles The latest commit adds this test as well. Thanks! :smile:

ashutoshsaboo:
@eliasdorneles Why do the tests fail now? :( Could you help?

tests/datatypes/test_str.py
```diff
@@ -32,7 +32,7 @@ def test_isspace(self):
 
     def test_isalpha(self):
         self.assertCodeExecution("""
-            for s in ['Hello World', 'hello wORLd.', 'Hello world.', '',
+            for s in ['Hello World', 'hello wORLd.', 'Hello world.', '', 'hello1'
```
smilin-desperado: @ashutoshsaboo did you mean to remove the comma after the hello1 string? Without it, 'hello1' and 'this' will be implicitly concatenated into 'hello1this' and I'm not sure that is what you want to test.

ashutoshsaboo: Resolved in the latest commit. Thanks for helping out! :smile:

ashutoshsaboo:
@eliasdorneles Finally this passes all the tests. Ready for merge now?

eliasdorneles:
@ashutoshsaboo yes, looks good now, thanks!

Issue #352, opened 2017-02-05
Exceptions usage fail in unexpected ways


eliasdorneles:
Here are some test cases that don't work as expected (tested creating a file `testraise.py`, putting these snippets on it then compiling w/ VOC and running):

    raise ValueError("test")

results in:

```
Exception in thread "main" java.lang.NoSuchMethodError: org.python.exceptions.ValueError.<init>(Lorg/python/Object;)V
	at python.testraise.__init__.module$import(testraise.py:1)
	at python.testraise.__init__.main(testraise.py)
```

However, if you do `raise ValueError()` (no args), the exception is raised normally.

The inverse happens for `KeyError`: if you don't give it an argument, it fails with the same error as above, but it works if you give it an argument.

A few other exceptions I tested (TypeError, AttributeError and KeyboardInterrupt) also have the same failure pattern as ValueError, there are probably more of this.

~~Another thing to note (possibly related) is that this also fails:~~ (EDIT: removed separate error, which was duplicate of https://github.com/pybee/voc/issues/350)

Deepanshu2017:
I have fixed this issue. I will create a PR tomorrow after some code clean up.

eliasdorneles:
Hey there!
Thanks for contributing.
Since it’s been a couple days and this PR is incomplete, we’re going to close it for now.
If you’d like to re-open it and fix it up, we’ll be happy to look at it again.

eliasdorneles:
Oops, closed by mistake there! :point_up_2: 
Sorry for the noise

eliasdorneles:
Quoting from [a previous comment](https://github.com/pybee/voc/pull/415#issuecomment-286221329), a fix for this issue should pass the following test:

```
self.assertCodeExecution("""
    for exc in [KeyError, ValueError, TypeError]:
        try:
            raise exc("one")
        except Exception as e:
            print(e, e.args)
        try:
            raise exc("one", 2)
        except Exception as e:
            print(e, e.args)
""")
```

dibyadas:
This problem arose due to wrong function declaration. 
In ValueError and KeyError , the constructor required an python object array,
```public ValueError(org.python.Object[] args, java.util.Map<java.lang.String, org.python.Object> kwargs)```
But in ast.py , visit_Raise() ,
```java.Init(exception, *(['Lorg/python/Object;'] * len(args)))```
which translated to this constructor call :- 
[...]

PR #415, opened 2017-03-13
Resolves Issue #352 | Corrects Exception Implementation, similar to CPython

ashutoshsaboo:

Hi @freakboy3742 @cflee @eliasdorneles .

This PR resolves Issue #352 . The major issue, that was, creating the problem, was that kwargs is not getting used itself. Hence removing the same resolves the issue. 

Please review the PR! Thanks! :smile: 

Do I need to add anything else? Or does this seem ready to merge?

### DIFF
--- a/python/common/org/python/exceptions/BaseException.java
+++ b/python/common/org/python/exceptions/BaseException.java
@@ -9,6 +9,14 @@ public class BaseException extends org.python.types.Object {
         // System.out.println("EX: " + this);
     }
 
+    public BaseException(org.python.types.Tuple tuple) {
+        super();
+    }
+
+    public BaseException(org.python.Object args) {
+        super(args.toString());
+    }    
+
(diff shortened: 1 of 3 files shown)


eliasdorneles:
Hi @ashutoshsaboo !
So, this doesn't really fix the problem in #352 -- the fix for that should pass the following test case:

```
self.assertCodeExecution("""
    for exc in [KeyError, ValueError, TypeError]:
        try:
            raise exc("one")
        except Exception as e:
            print(e, e.args)
        try:
            raise exc("one", 2)
        except Exception as e:
            print(e, e.args)
""")
```

eliasdorneles:
Hey there!
Thanks for contributing.
Since it’s been a couple days and this PR is incomplete, we’re going to close it for now.
If you’d like to re-open it and fix it up, we’re happy to look at it again.

PR #416, opened 2017-03-13
Implement FrozenSet constructors, __len__

ASP1234:
**Changes**

- Implemented FrozenSet Constructor as per the requirement

- Added __len__ method to finish the builtin len() fn

**Testing**

- Pass all tests for len()

- Also added sample data to frozenset test in utils.py . New implementation passes the test data successfully

### DIFF
--- a/python/common/org/python/types/FrozenSet.java
+++ b/python/common/org/python/types/FrozenSet.java
@@ -7,7 +7,8 @@ public class FrozenSet extends org.python.types.Object {
             __doc__ = "frozenset() -> empty frozenset object" +
                     "frozenset(iterable) -> frozenset object\n" +
                     "\n" +
-                    "Build an immutable unordered collection of unique elements.\n"
+                    "Build an immutable unordered collection of unique elements.\n",
+            default_args = {"iterable"}        
     )
     public FrozenSet(org.python.Object[] args, java.util.Map<java.lang.String, org.python.Object> kwargs) {
         if (args[0] == null) {
@@ -19,15 +20,15 @@ public class FrozenSet extends org.python.types.Object {
                 );
             } else if (args[0] instanceof org.python.types.List) {
                 this.value = java.util.Collections.unmodifiableSet(
-                		new java.util.HashSet<org.python.Object>(
+                        new java.util.HashSet<org.python.Object>(
                         ((org.python.types.List) args[0]).value)
                 );
             } else if (args[0] instanceof org.python.types.Tuple) {
                 this.value = java.util.Collections.unmodifiableSet(
-                		new java.util.HashSet<org.python.Object>(
+                        new java.util.HashSet<org.python.Object>(
                         ((org.python.types.Tuple) args[0]).value)
                 );
-            } else {
+            } else if (org.Python.iter(args[0]) instanceof org.python.Iterable) {
                 org.python.Iterable iterator = org.Python.iter(args[0]);
                 java.util.Set<org.python.Object> generated = new java.util.HashSet<org.python.Object>();
                 try {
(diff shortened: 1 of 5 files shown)


tests/utils.py
```diff
@@ -647,6 +647,9 @@ def wrapper(*args, **kwargs):
     'frozenset': [
             'frozenset()',
             'frozenset({1, 2.3456, "another"})',
+            'frozenset([1, 2.3456, "another"])',
+            'frozenset((1, 2.3456, "another"))',
+            'frozenset(iter([1, 2.3456, "another"]))',
```
cflee: It looks like you're trying to test the frozenset constructor here. This is not the best place to do it, as the `SAMPLE_DATA` is meant to provide a variety of values, not a bunch of the same values.

Instead, these tests should go into `tests/datatypes/test_frozenset.py`, under the 'FrozensetTests' test case.

ASP1234:
**Addressing PR comments**

- Fixed codestyle errors

- Moved tests to FrozenSetTests

- Added behavior for non-iterable i/p

tests/datatypes/test_frozenset.py
```diff
@@ -2,7 +2,33 @@
...
+
+        self.assertCodeExecution("""
+            x = frozenset((1, 2))
+            print(x)
+            """)
+
+        self.assertCodeExecution("""
+            x = frozenset('a')
+            print(x)
+            """)
```
cflee: Please combine all these tests into one large `self.assertCodeExecution()`, it's ok to keep assigning the value of x and then accessing it.

tests/datatypes/test_frozenset.py
```diff
@@ -2,7 +2,33 @@
...
+
+        self.assertCodeExecution("""
+            x = frozenset(['a'])
+            print(x)
+            """)
+
+        self.assertCodeExecution("""
+            x = frozenset((1, 2))
```
cflee: This has tested list and tuple, should there be a test for a set (and frozenset) as input too? We should also test whether changing the source set/array/tuple affects the newly created frozenset.

python/common/org/python/types/FrozenSet.java
```diff
@@ -1,13 +1,82 @@
...
+
+                throw new org.python.exceptions.TypeError(buffer.toString());
+            }
+        }
+    }
+
+    @org.python.Method(
+            __doc__ = ""
+    )
+    public org.python.types.Str __repr__() {
```
cflee: Since this method has been implemented, the `tests.builtins.test_repr.BuiltinReprFunctionTests.test_frozenset` test should be passing as well, could you check why it's still failing?

python/common/org/python/types/FrozenSet.java
```diff
@@ -1,13 +1,82 @@
...
+                    }
+                } catch (org.python.exceptions.StopIteration si) {
+                }
+                this.value = java.util.Collections.unmodifiableSet(generated);
+            } else {
+                java.lang.StringBuilder buffer = new java.lang.StringBuilder("'");
+                buffer.append(args[0].typeName());
+                buffer.append("' object is not iterable");
+
+                throw new org.python.exceptions.TypeError(buffer.toString());
```
cflee: Is this `else` case necessary, since `org.Python.iter()` will throw TypeError with this exact message if the parameter is not iterable?

python/common/org/python/types/FrozenSet.java
ASP1234: It's not failing. Just needed to remove from not implemented list :)

python/common/org/python/types/FrozenSet.java
```diff
@@ -1,13 +1,82 @@
...
+
     @org.python.Method(
             __doc__ = "frozenset() -> empty frozenset object" +
                     "frozenset(iterable) -> frozenset object\n" +
                     "\n" +
-                    "Build an immutable unordered collection of unique elements.\n"
-    )
+                    "Build an immutable unordered collection of unique elements.\n",
+            default_args = {"iterable"}
+            )
```
ASP1234: Fixed!

python/common/org/python/types/FrozenSet.java
ASP1234: Didn't know about this earlier. Removed the redundant else part

ASP1234:
**Changes**

- Moved creation tests into a single tests

- Tested by changing source of frozenset, expected o/p should not change with source changes

- Implemented __repr__, thus tests pass for frozenset __repr__ 😄 

- Removed else part from constructors for non iterable objects. TypeError takes care of it it seems!

ASP1234:
Checkstyle fix

PR #417, opened 2017-03-13
Implements the isnumeric() method in Python for strings

ashutoshsaboo:

Implements the [isnumeric()](https://docs.python.org/3/library/stdtypes.html#str.isnumeric) method. 

@eliasdorneles @cflee Kindly review the PR! Thanks! :smile:

### DIFF
--- a/python/common/org/python/types/Str.java
+++ b/python/common/org/python/types/Str.java
@@ -767,7 +767,7 @@ public class Str extends org.python.types.Object {
         int c = 0;
         java.lang.String checkString = this.value;
         char currentCharacter;
-        while (c < (checkString.length()-1)) {
+        while (c < (checkString.length() - 1)) {
             currentCharacter = checkString.charAt(c);
             if (!Character.isDigit(currentCharacter)) {
                 return new org.python.types.Bool(false);
--- a/python/common/org/python/types/Str.java
+++ b/python/common/org/python/types/Str.java
@@ -767,7 +767,7 @@ public class Str extends org.python.types.Object {
         int c = 0;
         java.lang.String checkString = this.value;
         char currentCharacter;
-        while (c < (checkString.length() - 1)) {
+        while (c < (checkString.length())) {
             currentCharacter = checkString.charAt(c);
             if (!Character.isDigit(currentCharacter)) {
                 return new org.python.types.Bool(false);
--- a/python/common/org/python/types/Str.java
+++ b/python/common/org/python/types/Str.java
@@ -767,6 +767,9 @@ public class Str extends org.python.types.Object {
         int c = 0;
         java.lang.String checkString = this.value;
         char currentCharacter;
+        if (checkString.length() == 0) {
+            return new org.python.types.Bool(false);
+        }
         while (c < (checkString.length())) {
             currentCharacter = checkString.charAt(c);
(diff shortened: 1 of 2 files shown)


python/common/org/python/types/Str.java
```diff
@@ -764,7 +764,17 @@ private long toPositiveIndex(long index) {
             __doc__ = ""
     )
     public org.python.Object isnumeric() {
-        throw new org.python.exceptions.NotImplementedError("isnumeric() has not been implemented.");
+        int c = 0;
+        java.lang.String checkString = this.value;
+        char currentCharacter;
+        while (c < (checkString.length()-1)) {
```
eliasdorneles: There are some style checks failing in the build: https://travis-ci.org/pybee/voc/jobs/210692217#L187-L188

Also, this looks like there is an off-by-one error, you should add a test for a string like `3.14M`.

ashutoshsaboo:
@eliasdorneles Pushed the latest commit, that resolves the style problem! :smile:

deep110:
@ashutoshsaboo I think empty string should also be added to test.

tests/datatypes/test_str.py
```diff
@@ -24,6 +24,12 @@ def test_isdigit(self):
                 print(s.isdigit())
             """)
 
+    def test_isnumeric(self):
+        self.assertCodeExecution("""
+            for s in [u'112358132134', u'3.14159', u'12312344df', '', 'this2009', '23443434', '3.14', '314.', 3.14M']:
```
deep110: there is syntax error in last element, `3,14M'` should be `'3.14M'`.

ashutoshsaboo: Resolved in the latest commit. Thanks. :smile:

python/common/org/python/types/Str.java
```diff
@@ -764,7 +764,20 @@ private long toPositiveIndex(long index) {
...
     )
     public org.python.Object isnumeric() {
-        throw new org.python.exceptions.NotImplementedError("isnumeric() has not been implemented.");
+        int c = 0;
+        java.lang.String checkString = this.value;
+        char currentCharacter;
+        if (checkString.length() == 0) {
+            return new org.python.types.Bool(false);
+        }
+        while (c < (checkString.length())) {
```
cflee: It might be clearer to have this as a `for (int c = 0; c < checkString.length(); c++)` loop instead, since you aren't using the value of `c` after and outside the loop.

python/common/org/python/types/Str.java
```diff
@@ -764,7 +764,20 @@ private long toPositiveIndex(long index) {
             __doc__ = ""
     )
     public org.python.Object isnumeric() {
-        throw new org.python.exceptions.NotImplementedError("isnumeric() has not been implemented.");
+        int c = 0;
+        java.lang.String checkString = this.value;
+        char currentCharacter;
```
cflee: It's best to only declare variables in the scope that they're needed – in this case this is only used within the loop, so it's not necessary to declare this outside. Similarly for `c`, but that's moot from the later comment about a for loop.

cflee:
Closing this as `Str.isnumeric()` has just been implemented in #528.

PR #418, opened 2017-03-13
[WIP] Add test for extending java class

eliasdorneles:
I've come up with a test for the issue I'm having while I try to build my next Android app using VOC.

I need to extend the class `android.view.View` to write a custom view, and I keep getting an `IllegalAccessError` like this:

```
java.lang.IllegalAccessError: Method 'void android.view.View.<init>()' is inaccessible to class 'org.eliasdorneles.drawingapp.app.SomeViewGroup' (declaration of 'org.eliasdorneles.drawingapp.app.SomeViewGroup' appears in /data/app/org.eliasdorneles-2/base.apk)
```

I've not been able to reproduce the exact same error outside Android environment, but the tests here already show part of the problem (I can't extend the Java class without providing an alternate constructor that does the right thing).

I have to dig a bit more to see how to do this, any pointers are appreciated. :)

### DIFF
--- a/tests/test_interoperability.py
+++ b/tests/test_interoperability.py
@@ -1,0 +1,56 @@
+from .utils import TranspileTestCase
+
+
+class InteroperabilityTest(TranspileTestCase):
+    def test_extend_java_class_with_simple_constructor(self):
+        self.assertJavaExecution(
+            """
+            class MyPythonClass(extends=com.example.MyClass):
+                pass
+
+            MyPythonClass()
+            print("Done.")
+            """,
+            java={
+                'com/example/MyClass': """
+                package com.example;
+
+                public class MyClass {
+                    public MyClass() {
+                        System.out.println("MyClass started");
+                    }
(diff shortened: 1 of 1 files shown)


eliasdorneles:
After digging a bit more in the code, I confirmed that [VOC currently only supports extending Java classes that have a default constructor](https://github.com/pybee/voc/blob/master/voc/python/methods.py#L445-L450).

I'm going to work-around this by doing the inheritance in Java side, just to be able to build my app -- it'll some time to figure out how to fix this in VOC.

freakboy3742:
I've just hit this exact problem trying to subclass `android.view.ViewGroup` - I've got some ideas on how to fix it, though. Grab me on gitter if you want to play with some ideas.

eliasdorneles:
This was fixed in: https://github.com/pybee/voc/commit/2b4e6de0c3d67bd46fd46f0c8b5ce4768c181d2c

Issue #406, opened 2017-03-12
voc is not capable to cope with very long python scripts


leasunhy:
For benchmarking purpose (refs #363), I created a python script that has thousands of lines. The script looks like this (which can be run with CPython):

```python
class A:
    def __init__(self, val):
        self.val = val

    def __add__(self, other):
        return self.val + other.val

class B(A):
    def __init__(self, val):
        self.val = val

    def __radd__(self, other):
        return self.val + other.val + 1

def test():
    a = A(1)
    b = B(2)
    print(a + b)
    ##### `print(a + b)` repeats 2000 times
    print(a + b)

if __name__ == "__main__":
        test()
```

When compiling the script, voc throws:
```
Traceback (most recent call last):
  File "/home/leasunhy/voc/env/bin/voc", line 11, in <module>
    load_entry_point('voc', 'console_scripts', 'voc')()
  File "/home/leasunhy/voc/voc-github/voc/__main__.py", line 46, in main
    verbosity=args.verbosity
  File "/home/leasunhy/voc/voc-github/voc/transpiler.py", line 33, in transpile
    transpiler.write(outdir)
  File "/home/leasunhy/voc/voc-github/voc/transpiler.py", line 62, in write
    javaclassfile.write(out)
  File "/home/leasunhy/voc/voc-github/voc/java/klass.py", line 438, in write
    method.write(writer)
  File "/home/leasunhy/voc/voc-github/voc/java/methods.py", line 209, in write
    attribute.write(writer)
  File "/home/leasunhy/voc/voc-github/voc/java/attributes.py", line 48, in write
    self.write_info(writer)
  File "/home/leasunhy/voc/voc-github/voc/java/attributes.py", line 458, in write_info
    attribute.write(writer)
  File "/home/leasunhy/voc/voc-github/voc/java/attributes.py", line 48, in write
    self.write_info(writer)
  File "/home/leasunhy/voc/voc-github/voc/java/attributes.py", line 1840, in write_info
    writer.write_u2(start_pc)
  File "/home/leasunhy/voc/voc-github/voc/java/klass.py", line 73, in write_u2
    self._outfile.write(struct.pack('>H', u2))
struct.error: 'H' format requires 0 <= number <= 65535
```

After some playing with this, I noticed that this issue is not because the number of lines of the python script is too big, but because the generated Java bytecode is too long.

[...]

PR #420, opened 2017-03-14
add error message for too large code

leasunhy:
refs #406

### DIFF
--- a/tests/test_large_code.py
+++ b/tests/test_large_code.py
@@ -1,0 +1,22 @@
+import unittest
+
+from voc.transpiler import Transpiler
+from voc.python.blocks import BlockCodeTooLarge
+from voc.python.methods import MethodCodeTooLarge
+
+
+class LargeCodeTest(unittest.TestCase):
+
+    def test_block_large_code(self):
+        "Test exception for large code of an anonymous block."
+        large_code = 'print(1 + 2)\n' * 3000
+        transpiler = Transpiler(verbosity=0)
+        self.assertRaises(BlockCodeTooLarge, transpiler.transpile_string,
+                          "test.py", large_code)
+
+    def test_method_large_code(self):
+        "Test exception for large code of a method."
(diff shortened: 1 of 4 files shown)


cflee:
I think we need tests to make sure that this is working as expected.

Also, some quick thoughts:
- Since this relies on the start position of the instruction, and instructions are variable length, it might let a slightly-too-long method through.
- The limit should be <= 65534 bytes, as stated in the [JVM 7 spec](http://docs.oracle.com/javase/specs/jvms/se7/html/jvms-4.html#jvms-4.7.3):
  > limiting the maximum size of the generated Java Virtual Machine code for any method, instance initialization method, or static initializer (the size of any code array) to 65534 bytes.
- Also from the above, it needs to check the length of resulting method, instance initialization method, and static initializers. I haven't dug into the code to check this, but definitely these need to be tested to demonstrate that the check correctly.
- Would it be more accurate and clearer to check the Code object's `code_length` property elsewhere, maybe even in the constructor if possible? It might be less surprising to see this error being raised from Code than from LineNumberTable, since it's actually the total size of the Code object's opcodes that is the problem, not LineNumberTable size.
- The message could be more informative and briefly refer to the ~64KiB/65534 bytes method size limit.

leasunhy:
@cflee Thank you for reviewing!
Frankly speaking, I've thought of adding a test for this... But I wondered where I should add the test to (because it seems that all other tests are the `compare-output-with-CPython` type of tests)... I didn't know whether it's appropriate for me to add a `code_length_test.py` only for this purpose... So I just tested by hand.. :-(

All of your points are valid! Please allow me some time to figure out how to improve this PR :-)

Thank you for your advice :-)

freakboy3742:
I'll also echo @cflee's comments - and add that the best place I can think of to deal with the offset problem is in `voc/python/blocks.py`, in `transpile_code()` - that's where the code offsets are evaluated.

leasunhy:
Ready for review here :-)

PR #422, opened 2017-03-14
Added implementation of replace method for String type

sipah00:


sipah00:
Thanks for the suggestion..working on the changes

PR #423, opened 2017-03-14
Added implementation of replace method for String type

sipah00:


bitdumper1:
The conditions should be like this
```java
if (condition) {
    statement;
}
```
The styles are different, doing a 
```
ant checkstyle
```
in home directory will help.

sipah00:
thanks

sipah00:
thanks..working on changes

eliasdorneles:
@sipah00 have you considered the idea I mentioned earlier of using `Pattern.quote` and `String.replaceFirst` for implementing the replace method?
Please consider it, I'd like to know if it's really the case that we can't use it, because this code is quite complicated and it will be much better for everyone if this can be done in a simpler way.

You should also use `Pattern.quote` in the other methods -- anytime you want to use a regular string meant to be literal inside a regular expression you have to quote it, otherwise it will fail when user provides a regex metacharacter as an argument.

Example case that fails with the code in this PR:

```
>>> '[oi]'.lstrip('[')
Exception in thread "main" RuntimeError: Unclosed character class near index 4
^[[]+
    ^
	at python._tempfile.__init__.module$import(_tempfile.py:1)
	at python._tempfile.__init__.main(_tempfile.py)
```

Please be _very thorough_ in your tests when you use regular expressions.

sipah00:
@eliasdorneles ,thank you very much..for review,I had considered replaceFirst method but i was facing problem in some case ..so i thought to do on my own way..but i will again try to implement replace method by your suggestion(replaceFirst)..

eliasdorneles:
Can you please make this pull request only about the `replace` method.
Rstrip and Lstrip were implemented on https://github.com/pybee/voc/pull/439
An implementation of `strip()` should be equivalent to `s.rstrip().lstrip()`, but it's best done in a separate PR.

Regarding the idea of using `replaceFirst`, you can drop it -- I've realized it wouldn't work for the following case:

    >>> 'abccabccabcc'.replace('abc', 'ab', 2)
    'abcabcabcc'

An implementation w/ replaceFirst would render a result equivalent to:

    >>> 'abccabccabcc'.replace('abc', 'ab', 1).replace('abc', 'ab', 1)
    'ababccabcc'

It's a good idea to add a test for both above cases.

PR #426, opened 2017-03-15
Added issubset() ,issuperset() and isdisjoint() 

bittikarwasara:


bittikarwasara:
Refs #36

tests/datatypes/test_set.py
```diff
@@ -157,12 +157,25 @@ def test_difference_update(self):
             print(x)
             """)
 
-    def test_intersection_update(self):
+    def test_issuperset(self):
         self.assertCodeExecution("""
             x = {1, 2, 3}
-            y = {3, 4, 5}
-            x.intersection_update(y)
-            print(x)
```
eliasdorneles: This looks like an unintended change, it is deleting the intersection_update test -- can you bring it back please?

bittikarwasara: ohh my bad , thanks for pointing it out  :)
Doing it right now !

python/common/org/python/types/Set.java
```diff
@@ -430,24 +430,41 @@ public boolean __setattr_null(java.lang.String name, org.python.Object value) {
...
+            __doc__ = "state whether other set is a subset",
+            args = {"other"}
     )
     public org.python.Object issubset(org.python.Object other) {
-        throw new org.python.exceptions.NotImplementedError("issubset() has not been implemented.");
+        if (other instanceof org.python.types.Set) {
+            org.python.types.Set otherSet = (org.python.types.Set) other;
+            return new org.python.types.Bool(otherSet.value.containsAll(this.value));
+        } else if (other instanceof org.python.types.Int) {
+            throw new org.python.exceptions.TypeError("'int' object is not iterable");
```
cflee: It seems quite unlikely that `issubset()` and `issuperset()` only throw a TypeError for int objects. Per the [python stdtypes docs](https://docs.python.org/3/library/stdtypes.html#set): 
> Note, the non-operator versions of union(), intersection(), difference(), and symmetric_difference(), issubset(), and issuperset() methods will accept any iterable as an argument.

Perhaps the logic here (and in `issuperset()`) should handling sets/frozensets, then attempting to call `org.Python.iter(other)`, which either throws the appropriate TypeError or returns the Iterable, which is then used to perform the `issubset()` check.

All these should be exercised in the test cases. Also, if there's any case where NotImplemented is returned, that should be tested as well.

bittikarwasara: @cflee can you take a look at the changes now .

cflee:
@bittikarwasara It looks like the tests are failing on various edge cases. Could you check them out? If you click on the red cross next to the commit, you'll be able to view the Travis build logs, which will show you which tests are failing. You can [run those tests on your own machine](https://voc.readthedocs.io/en/latest/internals/contributing.html#running-the-test-suite) before you commit/push to make sure they're fixed.

Also, you should be able to combine most of the `self.assertCodeExecution()` calls under each test (e.g. `test_isdisjoint`) into one large block, it's not necessary to have them separated into so many `assertCodeExecution()` calls.

goanpeca:
@bittikarwasara could you rebase this with the latest master to get the new CI checks running?

Do you still have some time to finish up the needed tests on this PR? Do you need help?

Cheers :-)

cflee:
Closing, as this functionality was implemented in #515.

PR #429, opened 2017-03-16
Added implementation of builtin function ascii().

gEt-rIgHt-jR:
Created new branch | [Original pull request](https://github.com/pybee/voc/pull/425) 
* Removed some tests from `not_implemented` list
* `test_frozenset` fails because it is not implemented
* `test_class` , `test_complex` , `test_slice` fail because `__repr__()` doesn't give correct output
* `test_str` fails because, for strings like `hôvèrçràft` , `__repr__()` doesn't work (in CPython also)

### DIFF
--- a/python/common/org/Python.java
+++ b/python/common/org/Python.java
@@ -394,10 +394,65 @@ public class Python {
                     "As repr(), return a string containing a printable representation of an\n" +
                     "object, but escape the non-ASCII characters in the string returned by\n" +
                     "repr() using \\x, \\u or \\U escapes.  This generates a string similar\n" +
-                    "to that returned by repr() in Python 2.\n"
+                    "to that returned by repr() in Python 2.\n",
+            args = {"object"}
     )
-    public static org.python.types.Str ascii() {
-        throw new org.python.exceptions.NotImplementedError("Builtin function 'ascii' not implemented");
+    public static org.python.types.Str ascii(org.python.Object object) {
+        boolean isString = false;
+        if (object instanceof  org.python.types.Str) {
+            isString = true;
+        }
+        java.lang.String s = ((org.python.types.Str) object.__repr__()).value;
+        StringBuilder resultString = new StringBuilder();
+
+        if (s.length() == 3) {
+            if (s.charAt(0) == 39 && s.charAt(1) == 39 && s.charAt(2) == 39) {
+                return new org.python.types.Str("\"'\"");
+            }
+        }
+
+        for (int i = 0; i < s.length(); i++) {
+            java.lang.Character c = s.charAt(i);
+            if (c < 32) {
+                if (c == 9) {
+                    resultString.append("\\t");
+                    continue;
+                }
+                if (c == 10) {
+                    resultString.append("\\n");
+                    continue;
+                }
+                if (c == 13) {
+                    resultString.append("\\r");
+                    continue;
+                }
+                resultString.append("\\x");
+                resultString.append(java.lang.String.format("%02x", java.lang.Integer.valueOf(c)));
+                continue;
+            }
+            if (c < 127) {
+                if (isString && (c == 39 || c == 92)) {
+                    if (i == 0 || i == s.length() - 1) {
+                    } else {
+                        resultString.append("\\").append(c);
+                        continue;
+                    }
+                }
+                resultString.append(c);
+                continue;
+            }
+            if (c < 256) {
+                resultString.append("\\x");
+                resultString.append(java.lang.Integer.toHexString(c));
+                continue;
+            }
+            if (c <= java.lang.Character.MAX_VALUE) {
+                resultString.append("\\u");
+                resultString.append(java.lang.String.format("%04x", java.lang.Integer.valueOf(c)));
+            }
+        }
+
+        return new org.python.types.Str(resultString.toString());
     }
 
     @org.python.Method(
--- a/python/common/org/Python.java
+++ b/python/common/org/Python.java
@@ -399,7 +399,7 @@ public class Python {
     )
     public static org.python.types.Str ascii(org.python.Object object) {
         boolean isString = false;
-        if (object instanceof  org.python.types.Str) {
+        if (object instanceof org.python.types.Str) {
             isString = true;
         }
         java.lang.String s = ((org.python.types.Str) object.__repr__()).value;
@@ -412,7 +412,7 @@ public class Python {
         }
 
         for (int i = 0; i < s.length(); i++) {
-            java.lang.Character c = s.charAt(i);
+            char c = s.charAt(i);
             if (c < 32) {
                 if (c == 9) {
                     resultString.append("\\t");
@@ -427,13 +427,12 @@ public class Python {
                     continue;
                 }
                 resultString.append("\\x");
-                resultString.append(java.lang.String.format("%02x", java.lang.Integer.valueOf(c)));
+                resultString.append(java.lang.String.format("%02x", (int) c));
                 continue;
             }
             if (c < 127) {
                 if (isString && (c == 39 || c == 92)) {
-                    if (i == 0 || i == s.length() - 1) {
-                    } else {
+                    if (i != 0 && i != s.length() - 1) {
                         resultString.append("\\").append(c);
                         continue;
                     }
@@ -448,7 +447,7 @@ public class Python {
             }
             if (c <= java.lang.Character.MAX_VALUE) {
                 resultString.append("\\u");
-                resultString.append(java.lang.String.format("%04x", java.lang.Integer.valueOf(c)));
+                resultString.append(java.lang.String.format("%04x", (int) c));
             }
         }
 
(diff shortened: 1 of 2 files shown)


tests/builtins/test_ascii.py
```diff
@@ -2,28 +2,45 @@
...
+            print(ascii('\u0041'))
+
+            string = ''
+            for i in range(100):
+                print(ascii(chr(i)))
+                string += chr(i)
+            print(ascii(string))
+            """)
+
+    def tset_ascii_list(self):
```
cflee: This test might not be run right now, the method name should start with `test_`.

python/common/org/Python.java
```diff
@@ -394,10 +394,65 @@ public static void initializeModule(java.lang.Class cls, java.util.Map<java.lang
...
                     "repr() using \\x, \\u or \\U escapes.  This generates a string similar\n" +
-                    "to that returned by repr() in Python 2.\n"
+                    "to that returned by repr() in Python 2.\n",
+            args = {"object"}
     )
-    public static org.python.types.Str ascii() {
-        throw new org.python.exceptions.NotImplementedError("Builtin function 'ascii' not implemented");
+    public static org.python.types.Str ascii(org.python.Object object) {
+        boolean isString = false;
+        if (object instanceof  org.python.types.Str) {
```
cflee: There should be a single space here after `instanceof` – I'll adjust checkstyle to flag this.

python/common/org/Python.java
```diff
@@ -394,10 +394,65 @@ public static void initializeModule(java.lang.Class cls, java.util.Map<java.lang
...
+        StringBuilder resultString = new StringBuilder();
+
+        if (s.length() == 3) {
+            if (s.charAt(0) == 39 && s.charAt(1) == 39 && s.charAt(2) == 39) {
+                return new org.python.types.Str("\"'\"");
+            }
+        }
+
+        for (int i = 0; i < s.length(); i++) {
+            java.lang.Character c = s.charAt(i);
```
cflee: `java.lang.String.charAt()` actually returns a primitive `char` – this is autoboxing it to a `Character` object, which isn't necessary. All your subsequent usage of `c` seems to be relying on the autounboxing to get a primitive `char` (which is often then implicitly cast to `int`).

python/common/org/Python.java
```diff
@@ -394,10 +394,65 @@ public static void initializeModule(java.lang.Class cls, java.util.Map<java.lang
...
+                if (c == 10) {
+                    resultString.append("\\n");
+                    continue;
+                }
+                if (c == 13) {
+                    resultString.append("\\r");
+                    continue;
+                }
+                resultString.append("\\x");
+                resultString.append(java.lang.String.format("%02x", java.lang.Integer.valueOf(c)));
```
cflee: If c is a primitive `char`, you can just cast it to `(int)` here, instead of using `java.lang.Integer.valueOf()`.

python/common/org/Python.java
```diff
@@ -394,10 +394,65 @@ public static void initializeModule(java.lang.Class cls, java.util.Map<java.lang
...
+                continue;
+            }
+            if (c < 127) {
+                if (isString && (c == 39 || c == 92)) {
+                    if (i == 0 || i == s.length() - 1) {
+                    } else {
```
cflee: Try to avoid having an empty `if` body and everything in the `else` block – the conditional should be inverted instead. In this case it'd be `i != 0 && i != s.length() - 1`.

tests/builtins/test_ascii.py
```diff
@@ -2,28 +2,45 @@
...
+    def test_ascii_defs(self):
+        self.assertCodeExecution("""
+            class TestClass:
+                def __init__(self, value):
+                    self.value = value;
+
+                def get_incremented_value(self, by):
+                    return self.value + by
+
+            print(ascii(TestClass(5).get_incremented_value(5)))
```
cflee: I'm not sure I understand what this test is testing – could you explain please? :)

tests/builtins/test_ascii.py
```diff
@@ -2,28 +2,45 @@
...
 
 class AsciiTests(TranspileTestCase):
-    pass
+    def test_ascii_without_unicode(self):
+        self.assertCodeExecution("""
+            print(ascii("Hello World!"))
+            print(ascii('\u0041'))
+
+            string = ''
+            for i in range(100):
```
cflee: It looks like going to 100 won't be sufficient to fully exercise your implementation of `ascii()` – it seems to have different logic for 32, 127, 256 and more?

python/common/org/Python.java
```diff
@@ -394,10 +394,65 @@ public static void initializeModule(java.lang.Class cls, java.util.Map<java.lang
...
+    public static org.python.types.Str ascii(org.python.Object object) {
+        boolean isString = false;
+        if (object instanceof  org.python.types.Str) {
+            isString = true;
+        }
+        java.lang.String s = ((org.python.types.Str) object.__repr__()).value;
+        StringBuilder resultString = new StringBuilder();
+
+        if (s.length() == 3) {
+            if (s.charAt(0) == 39 && s.charAt(1) == 39 && s.charAt(2) == 39) {
```
cflee: Hmm, could you explain why this special case for `'''` is necessary?

cflee (changes requested):
Great to see that the rebase/merge issues have been sorted out! It's definitely best to work on PRs on a branch (as this PR is) instead of master, so that you can keep your fork's master branch up to date with the upstream pybee/voc repo.

Some questions and small pointers in the review below.

(The old PR comments are no longer available due to the commits being removed, so I hope I'm not duplicating any previous comments...)

tests/builtins/test_ascii.py
gEt-rIgHt-jR: Sorry, it was supposed to be 300

python/common/org/Python.java
gEt-rIgHt-jR: Thanx for pointing it out

python/common/org/Python.java
gEt-rIgHt-jR: The output should be `"'"` and not `'''`
If I just check `charAt(1) == 39` then it won't work for `b''`

python/common/org/Python.java
gEt-rIgHt-jR: I thought we dont use direct `int` and `char`

gEt-rIgHt-jR:
@cflee @freakboy3742 Updated the PR. Any changes?

gEt-rIgHt-jR:
Learning a lot from this PR :)

python/common/org/Python.java
cflee: It's fine, usually the issue is with not using the Python types when necessary, then everything falls apart :) Also for this specific case, `c` is only used within the method, so it's not visible from outside the method anyway.

python/common/org/Python.java
cflee: I see. If that's the case, there definitely should be a test for it.

What happens if the input is just one or two `'`? Or some other number of `'`?

cflee:
@gEt-rIgHt-jR It looks like most of the review comments have been addressed, but I think the single quote issue needs to be explored further. In particular there needs to be more tests to ensure (and demonstrate) that the behaviour is correct, especially around single quote and backslash as the special cases for those seem quite complex.

There's no real need to rebase unless there's a merge conflict stopping us from merging the PR.

gEt-rIgHt-jR:
@cflee Fixed the issue of Quotes for strings.
But the issue is still there for `lists`.
For example, `['hello\'']` gives incorrect answer.
How should I tackle `lists`?

gEt-rIgHt-jR:
@cflee The main problem is `__repr__()` doesn't give correct string
Should I change the `__repr__()` method of datatypes?

cflee:
@gEt-rIgHt-jR It looks like list correctly delegates to the values' `__repr__()`. So yes, you might have to fix `__repr__()` for Str at least. If you're doing that, make sure to add appropriate tests.

gEt-rIgHt-jR:
@cflee  `List`'s `__repr__()` also has quotes problems
For example, `l = ["hel'lo"];print(repr(l))` gives `['hel'lo']` but,
it should give `["hel'lo"]`

cflee:
@gEt-rIgHt-jR Yes, your example illustrates the problem with `Str.__repr__()`.

As you can see in [`List.__repr__()`](https://github.com/pybee/voc/blob/51339b510986c70bb58a3a30f4211e689cf2b95e/python/common/org/python/types/List.java#L136-L149), it is only providing `[]` and `, `. Everything else calls from invoking each individual object's `.__repr__()` method.

Hence, for your example that prints `['hel'lo']`, everything apart from `[]` are from `Str.__repr__()`. You will have to fix that method (with tests for both `Str.__repr__()` and add an case to `test_ascii_list`) to ensure that this `ascii()` method works correctly.

Also, the test is failing due to the triple quotes prematurely ending the code block in the test, please look into that as well.

goanpeca:
@gEt-rIgHt-jR could you rebase the PR with the latest master to trigger the new CI ?

Do you still have some time to finish this PR?

Cheers :-)

phildini:

PR #431, opened 2017-03-17
Adjust checkstyle config

cflee:
After a couple of PR reviews, found that some options should be added:

- [SingleSpaceSeparator](http://checkstyle.sourceforge.net/config_whitespace.html#SingleSpaceSeparator): "Checks that non-whitespace characters are separated by no more than one whitespace."
  This prevent things like `x instanceof  y` where there are two spaces after `instanceof` (and which GFM eats in comments...). Side-effect: it also forbids horizontal alignment outside of comments, which seems like a reasonable tradeoff.
- [EmptyBlock](http://checkstyle.sourceforge.net/config_blocks.html#EmptyBlock): "Checks for empty blocks."
  This prevents things like `if (conditional) {} else { //code }`, where the conditional should be inverted.

I'm deliberately pushing a commit with only the checkstyle config change first to demonstrate that the build fails, and will follow up with the code fixes.

### DIFF
--- a/checkstyle.xml
+++ b/checkstyle.xml
@@ -130,6 +130,7 @@
         <module name="Indentation">
             <property name="lineWrappingIndentation" value="8"/>
         </module>
+        <module name="SingleSpaceSeparator"/>
 
         <!-- Modifier Checks                                    -->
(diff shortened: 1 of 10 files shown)


cflee:
This is now ready for review.

PR #438, opened 2017-03-17
Added str.casefold()

encore0117:
Hi! I used the casefold mapping from ftp://ftp.unicode.org/Public/UNIDATA/CaseFolding.txt. This implementation only covers "Full" casefolding. 

I did not think I should be putting the mapping in a file, but simply putting it inline will result in lengthy code. 

Please let me know how I can make it better.

Thanks~


python/common/org/python/types/CaseFolding.txt
```diff
@@ -0,0 +1,104 @@
+00DF; F; 0073 0073; # LATIN SMALL LETTER SHARP S
```
freakboy3742: What is the source of this data? It strikes me as unusual that this (a) is all the entries required in a case folding table, and (b) that Java doesn't have such a table internally.

encore0117:
Hi! Thanks for getting back to me!

From the documentation of Python [https://docs.python.org/3/library/stdtypes.html#str.casefold](url), it specifies that it adopts the default casefolding algorithm mentioned in [http://www.unicode.org/versions/Unicode9.0.0/ch03.pdf](url) (there is a section called "Default Case Folding"). It uses "full case folding", which means the resulted string length may change, e.g. casefolded "ß" is "ss". That's where the "F" in the mapping file comes from. (I truncated the original mapping file [ftp://ftp.unicode.org/Public/UNIDATA/CaseFolding.txt](url) to only include the "F" part.)

For the special cases, turkic i, from what I have read in [https://groups.google.com/forum/#!topic/python-ideas/JZZRwU2B3kc](url), I do not think the one implemented this method in Python mutated locale to make this work, and what they have done was to only implement full casefolding. I changed the test case and verified my implementation has the same behaviour as Python's.

To be honest, I do not know if Java has the equivalent method as the closest method I could find in Java was str.toLowerCase() and I'm sure it has different behaviour as Python's str.casefold() in dealing with characters like german b.

encore0117:
Also, I think using a relative path in code might be an issue, but I don't know if there is better solution. Please let me know. Thanks~

python/common/org/python/types/CaseFolding.txt
encore0117: Hi! I've made my comment below.

freakboy3742:
Closing this in preference to #480, which is a lot simpler, but seems to work. If you can identify a test case where the approach in #480 fails, please open a ticket.

PR #440, opened 2017-03-18
Change C:\Python35\python.exe to C:\Python34\python.exe in installation doc

soandrew:
I believe there's a typo in the installation doc for setting up the virtual environment on Windows. 
As it is right now, it is going to use Python 3.5 to set up the environment instead of Python 3.4

Throughout the docs, it says that VOC won’t build on Python3.5+ and that a separate Python 3.4+ installation is required, so I am pretty sure this is just a typo.

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -6,11 +6,11 @@ VOC
 
 **VOC is an early alpha project. If it breaks, you get to keep all the shiny pieces.**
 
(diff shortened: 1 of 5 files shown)


soandrew:
Ah I see, is it only up to Python 3.5 for now? Or is Python 3.6 supported as well?

freakboy3742:
We run CI on 3.4, 3.4 and 3.6, so all three should be OK.

PR #441, opened 2017-03-18
Refs #361 Fix

ASP1234:
**Changes**

- Change exception from Runtime to ClassNotFound in importPythonModule. This is because the method itself throws ClassNotFound exception rather than Runtime

- Removed the try catch block, This is because if a module doesn't exist in python too then why are we catching the ClassNotFoundException rather than simply throwing it. Main reason of issue#361

**Testing**

- test_import successfully completed

- Expected o/p for issue #361

### DIFF
--- a/python/common/org/python/ImportLib.java
+++ b/python/common/org/python/ImportLib.java
@@ -120,16 +120,8 @@ public class ImportLib {
                             python_module = importNativeModule(mod_name);
                         } else {
                             // org.Python.debug("handle IMPORT", mod_name);
-                            try {
-                                // try to import unknown native module as python..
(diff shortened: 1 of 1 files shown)


freakboy3742 (changes requested):
Changes without tests are *very* suspicious - and it's not clear why these changes fix the problem you've described.

ASP1234:
- Import tests are passing with these changes

- It produces the expected #361 behavior

- We need to throw ClassNotFound Exception rather than throwing runtime exception here

goanpeca:
@ASP1234 could you perform a rebase with the latest master?

Cheers

phildini:

PR #442, opened 2017-03-18
Added strip method for String type

sipah00:


python/common/org/python/types/Str.java
```diff
@@ -1148,10 +1149,21 @@ private long toPositiveIndex(long index) {
...
 
     @org.python.Method(
-            __doc__ = ""
+            __doc__ = "strip",
+            default_args = {"charsToBeRm"}
     )
-    public org.python.Object strip() {
-        throw new org.python.exceptions.NotImplementedError("strip() has not been implemented.");
+    public org.python.Object strip(org.python.Object charsToBeRm) {
+        String strim = "";
```
eliasdorneles: As I mentioned in the previous pull request, this is best implemented by calling rstrip and lstrip, like:

```
    public org.python.Object strip(org.python.Object chars) {
        return ((Str ) this.rstrip(chars)).lstrip(chars);
    }
```

sipah00: @eliasdorneles ,I got an issue regarding rstrip function ..it fails in case of print('ABCBBAA'.rstrip('AB')) gives ABCBBAA , but it should be ABC.
but in my  implementation of rstrip it gives correct result. Should i submit PR.

cflee:
Closing this as #484 has been merged.

PR #443, opened 2017-03-18
Added replace method for String Type

sipah00:


sipah00:
@freakboy3742 , I have added tests for replace method.

cflee:
@sipah00 The tests are failing due to failing the Python code style checks – you may want to run `flake8 && ant checkstyle` locally after fixing those lines, just to make sure that it will pass the Java code style checks as well.

PR #446, opened 2017-03-19
Added isalnum() and isdecimal() functions for Str type

gEt-rIgHt-jR:


tests/datatypes/test_str.py
```diff
@@ -30,13 +30,27 @@ def test_isspace(self):
                 print(s.isspace())
             """)
 
+    def test_isalnum(self):
+        self.assertCodeExecution("""
+            for word in ["", "12", "abc", "abc12", "\u00c4", "\x41", "a@g", "äÆ",
+            "12.2", "'hi'", "hello!!", "gEt_rIgHt_jR."]:
```
cflee: Please use strings that specifically exercise various edge cases :) I'm not sure what the last one is testing for, compared to the rest.

python/common/org/python/types/Str.java
```diff
@@ -708,10 +708,21 @@ private long toPositiveIndex(long index) {
...
     @org.python.Method(
-            __doc__ = ""
+            __doc__ = "string.isalnum() -> bool\n" +
+                    "\n" +
+                    "Returns false when the string has atleast ONE character which is\n" +
+                    "not an alphanumeric or if the string is empty.\n"
     )
     public org.python.Object isalnum() {
-        throw new org.python.exceptions.NotImplementedError("isalnum() has not been implemented.");
+        if (!this.value.isEmpty()) {
```
cflee: For both methods, can you shift the logic flow such that if it's empty, it returns false first? This will let you 'flatten' the rest of the logic so that it's clearer what the method is doing. Right now there are these three return statements and you have to kind of read it very carefully to see what's going on.

i.e.
```
if (empty) {
    return false
}
for (...) {
    if (...) {
        return false
    }
}
return true
```

(Some of the other methods have these huge if/elseif/elseif blocks that more or less encompass the whole method, but that's usually clearer when we need to handle different parameter types separately.)

python/common/org/python/types/Str.java
```diff
@@ -708,10 +708,21 @@ private long toPositiveIndex(long index) {
     }
 
     @org.python.Method(
-            __doc__ = ""
+            __doc__ = "string.isalnum() -> bool\n" +
+                    "\n" +
+                    "Returns false when the string has atleast ONE character which is\n" +
```
cflee: I'm just curious, where are these docstrings coming from?

tests/datatypes/test_str.py
gEt-rIgHt-jR: Sorry :P

python/common/org/python/types/Str.java
gEt-rIgHt-jR: I also wanted to do it like that but as mentioned in above comment, I took `isalpha()` as a reference (before refactoring).
I'll change it to this logic flow.
Thank you for reviewing my PR :)

python/common/org/python/types/Str.java
gEt-rIgHt-jR: I took [`isalpha()` ](https://github.com/pybee/voc/blob/master/python/common/org/python/types/Str.java#L727) as a reference. Should I keep it or remove?

gEt-rIgHt-jR:
@cflee Changes made.

python/common/org/python/types/Str.java
cflee: Ah, I see. We don't currently have a way of automatically verifying methods' docstrings against what Python has, but you should use the actual ones where possible! For this particular method you can get it by `str.isalnum.__doc__` in a Python console.

(Please do this for isdecimal below as well.)

python/common/org/python/types/Str.java
cflee: Sure, I understand. It's certainly a very good idea to think twice about whether there's some reason behind what may initially appear to be strange/weird code!

Sometimes what I do when I'm not sure is regardless of whether I follow or change it to a different way, I highlight it in the message/comments when creating a PR, so that anyone else can also take a closer look and agree/disagree with the action. It could be an opportunity to improve both the current PR and existing code!

python/common/org/python/types/Str.java
```diff
@@ -708,10 +708,21 @@ private long toPositiveIndex(long index) {
     }
 
     @org.python.Method(
-            __doc__ = ""
+            __doc__ = "string.isalnum() -> bool\n" +
```
cflee: As mentioned in the other comment – the docstrings for this `isdecimal()` should match the ones in CPython.

tests/datatypes/test_str.py
```diff
@@ -30,13 +30,27 @@ def test_isspace(self):
                 print(s.isspace())
             """)
 
+    def test_isalnum(self):
+        self.assertCodeExecution("""
+            for word in ["", "12", "abc", "abc12", "\u00c4", "\x41", "a@g", "äÆ",
```
cflee: Since `alnum()` seems to be effectively some combination of `isalpha()` and `isdecimal()`, probably should include some strings with uppercase letters here, to ensure that uppercase letters don't make strings non-`alnum` when they should be.

python/common/org/python/types/Str.java
gEt-rIgHt-jR: Thank you so much for this beautiful advice :)

gEt-rIgHt-jR:
@cflee Changed made

PR #448, opened 2017-03-19
Str.rjust() implementation

Saurabh--Kumar:
Str method rjust() implemented . Refs #38


tests/datatypes/test_str.py
```diff
@@ -517,6 +517,33 @@ def test_rindex(self):
...
+            print(st.rjust(100, 'X'))
+            print(st.rjust(10, 'X'))
+
+            """)
+
+        # test cases to generate exception
+        self.assertCodeExecution("""
+            st="a good cook could cook good"
+            try:
+                print(st.rjust(-20, 'cook'))
```
cflee: Does this trigger the edge case for negative width or for too-long fill char? It might be better to have those as separate tests, to make sure both cases get exercised.

python/common/org/python/types/Str.java
```diff
@@ -1026,10 +1026,40 @@ private long toPositiveIndex(long index) {
...
+            }
+            java.lang.StringBuffer str = new java.lang.StringBuffer(w);
+            int balance = w - this.value.length();
+            for (int i = 0; i < balance; i++) {
+                str.append(ch);
+            }
+            str.append(this.value);
+            return new org.python.types.Str(str.toString());
+        }
+        return new org.python.exceptions.TypeError("integer argument expected, got " + Python.typeName(width.getClass()));
```
cflee: If you move this TypeError up to the top, in a "else if width not instanceof Int" block right after `width == null`, then you can flatten and remove this `if (width instanceof org.python.types.Int)` check. In other words, get the more essential missing args/wrong type args handling out of the way at the top, followed by the handling for the optional args like fillChar, then it's clear where the logic for this particular `rjust()` method is.

Perhaps it probably should be a `throw` new exception instead of a `return` as well?

Saurabh--Kumar: @cflee made changes.

cflee:
Looks great, thanks!

PR #449, opened 2017-03-19
Implemented __iter__() for Set.

bitdumper1:
Added __iter__() for Bool, Complex, Float and Int.
Added Set_Iterator and implemented __iter__() for Set.


bitdumper1:
Before the __iter__() method
```
Exception in thread "main" AttributeError: 'int' object has no attribute '__iter__
```
but it should give
```
TypeError: 'int' object is not iterable
```
So I changed to give correct errors.

cflee:
This is what I get on Python for Int:

```
>>> x = 5
>>> x.__iter__()
AttributeError: 'int' object has no attribute '__iter__'
>>> iter(x)
TypeError: 'int' object is not iterable
```

You can verify this result for the other datatypes. It looks like `__iter__()` should return `AttributeError` instead of `TypeError`.

Also, [`org.Python.iter()`](https://github.com/pybee/voc/blob/69e910e8f5fd6ae1be3b26703a4b395ec326a21c/python/common/org/Python.java#L905-L916) seems like it's already implementing the conversion from `AttributeError` to `TypeError`.

PR #451, opened 2017-03-19
Remove .class file merged in #445

cflee:
- This was committed in 1f3a46e via #445, together with a ton of other strange files.


freakboy3742:
Oops. Someone better take my keys away from me. :-)

cflee:
@freakboy3742 It was admittedly hard to spot since it's collapsed in the Github web diffs as a binary file – I only noticed by accident when looking through PRs for something else...

PR #453, opened 2017-03-19
Implement str strip(), and fix lstrip() and rstrip() implementations

soandrew:
I implemented the str method strip()
I also found problems with the lstrip() and rstrip() implementations when dealing with whitespace, None, and stripping the entire string. I included test cases for those as well as fixed implementation code to pass these tests. 
Refs #36


python/common/org/python/types/Str.java
```diff
@@ -1147,10 +1140,32 @@ private long toPositiveIndex(long index) {
...
+                    "The chars argument is not a prefix or suffix; rather, all combinations of its values are stripped.",
+            default_args = {"chars"}
     )
-    public org.python.Object strip() {
-        throw new org.python.exceptions.NotImplementedError("strip() has not been implemented.");
+    public org.python.Object strip(org.python.Object chars) {
+        if (chars == null || chars instanceof org.python.types.NoneType) {
+            return new org.python.types.Str(this.value.trim());
+        } else if (chars instanceof org.python.types.Str) {
+            org.python.types.Str chars_str = (org.python.types.Str) chars;
```
eliasdorneles: Isn't it best to just call rstrip and lstrip here?

cflee:
@soandrew Hmm, it looks like #484 has been merged.

I do think the tests on this PR are more comprehensive – if you could port/rebase those onto the latest master, we might discover some bugs (like the handling of `None` I've just mentioned on that PR), that could be useful to discover and fix.

soandrew:
I merged my tests with the latest tests in master and found that there were still issues with handling of whitespace and None in the existing implementation. I've included my fix of these algorithms in the updated PR.

tests/datatypes/test_str.py
```diff
@@ -447,16 +447,21 @@ def test_partition(self):
...
-            str = "gggfoo"
-            print(str.lstrip('g'))
+            str = "\t\t   gggfoo "
             print(str.lstrip('h'))
-            str = "   foo"
             print(str.lstrip())
-            print("foot".lstrip("foobar"))
+            print(str.lstrip(None))
+            print(str.lstrip(''))
+            print(str.lstrip(\t ))
```
soandrew: Ah yes, good eye. Thanks!

cflee:
Thanks @soandrew!

PR #455, opened 2017-03-20
Implemented floor_divide on Object

AdilsonTorres:
Refs #36

### DIFF
--- a/python/common/org/python/types/Str.java
+++ b/python/common/org/python/types/Str.java
@@ -384,7 +384,7 @@ public class Str extends org.python.types.Object {
     @org.python.Method(
             __doc__ = ""
     )
-    public org.python.Object __floordiv__ (org.python.Object other){
+    public org.python.Object __floordiv__(org.python.Object other) {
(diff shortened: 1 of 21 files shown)


python/common/org/python/types/Str.java
```diff
@@ -384,6 +384,27 @@ public boolean __setattr_null(java.lang.String name, org.python.Object value) {
     @org.python.Method(
             __doc__ = ""
     )
+    public org.python.Object __floordiv__(org.python.Object other) {
+        if (other instanceof org.python.types.Complex) {
```
eliasdorneles: I believe this is not a behavior specific to strings, see:

```
>>> 2 // complex(1, 0)
TypeError: can't take floor of complex number.
>>>
>>> object() // complex(1, 0)
TypeError: can't take floor of complex number.
```

Have you tried using this code in the `Object` parent class instead?

AdilsonTorres: Hum... I see, so instead for me doing this only in Str, I should do on Object. I'll try do this and commit when I get any success!!!

PR #459, opened 2017-03-22
Update README.rst

kaushalpaulite:


freakboy3742:
This isn't correct. Once upon a time, VOC started with Python bytecode; but now it uses the AST.

kaushalpaulite:
Ok I got it now.
By the way I was browsing through the link which you had sent and I saw the
type_str.py file and we are supposed to choose any type and work in it.
But I did not get as how exactly we will be running our code and if you
could provide me with a example it would be really helpful.
Thank you.

PR #460, opened 2017-03-22
Add checkstyle rule for MultipleVariableDeclarations

cflee:
In certain programming styles (particularly old C-style code), there is a strong practice of placing all the variable declarations upfront at the top of a method/block. This eventually leads to (i) variables being declared far from their first usage, and (ii) variables with unnecessary scope, all of which contribute towards making code more difficult to read.

* Enabling the MultipleVariableDeclarations rule for checkstyle enforces this rule on CI builds.

The few code chunks fixed in current master to comply with this rule aren't particularly confusing, but I think it's a good preventive measure, especially where more complex logic and methods are involved.


cflee:
Restarting the entire build so that it's run against latest master with the rjust fix.

eliasdorneles:
Well, I'm not really convinced that declaring variables closer to its usage is always best for readability.
Note that the [Java code conventions about declaration placement](http://www.oracle.com/technetwork/java/javase/documentation/codeconventions-141270.html) say:

> Don't wait to declare variables until their first use; it can confuse the unwary programmer and hamper code portability within the scope

But I don't have any strong feelings about this, I'll let @freakboy3742 decide. :)

cflee:
@eliasdorneles 

Agreed, code style is ultimately very subjective! Do you have any particular opinion on declaring/initialising multiple variables on a single line though?

The code conventions definitely state that, but as a contrast, Bloch's _Effective Java_ argues for declaring variables close/where it is first used as a means to achieve "Item 45: Minimize the scope of local variables". ([Google Books link](https://books.google.com.sg/books?id=ka2VUBqHiWkC&pg=PA209&lpg=PA209&dq=java+declare+variables+first+use&source=bl&ots=yZLfKmq3N_&sig=ZjJXxC_M_qzL_nwaTgL8uiMeFU0&hl=en&sa=X&ved=0ahUKEwjvvYyL--zSAhVIN48KHac2DsgQ6AEISTAI#v=onepage&q=java%20declare%20variables%20first%20use&f=false))

It's an appeal to authority, but I guess that's all I can put forward for this argument. :)

freakboy3742:
I agree with Bloch's comment in principle, but I tend to interpret that as "put the declaration at the start of the *block* where it is used , rather than allowing definitions at any point in code, or putting all definitions at the start of the method. That includes, occasionally, a "conceptual block" if a method is especially long for some reason.

My gut tells me that this is something that isn't *that* big of an issue, and that automation is going to cause more complications and annoyance than it will fix/prevent.

cflee:
Sure. Closing this PR!

PR #463, opened 2017-03-22
Minor fix in README

dibyadas:
Lambda feature has been implemented. So removed it from the README


README.rst
```diff
@@ -35,7 +35,7 @@ What it does:
   Python code on a JVM (including Android's VM).
 
 It isn't a *completely* compliant Python 3.4 implementation - there are some
-language features (like lambdas and some builtin functions) that still need to be
```
eliasdorneles: it could mention `yield from` instead. :)

dibyadas: let's hope that we implement it as soon as possible  :smile:

PR #465, opened 2017-03-22
implement str.splitlines

sngguojie:
Refs #36

### DIFF
--- a/python/common/org/python/types/Str.java
+++ b/python/common/org/python/types/Str.java
@@ -1154,22 +1154,24 @@ public class Str extends org.python.types.Object {
         } else if (keepends instanceof org.python.types.Int) {
             parseKeepEnd = ((org.python.types.Int) keepends).value != 0;
         } else if (keepends instanceof org.python.types.Float) {
-            throw new org.python.exceptions.TypeError("integer argument expected, got "+keepends.typeName());
+            throw new org.python.exceptions.TypeError("integer argument expected, got " + keepends.typeName());
         } else if (keepends instanceof org.python.types.Complex) {
-        throw new org.python.exceptions.TypeError("can't convert "+keepends.typeName()+" to int");
+            throw new org.python.exceptions.TypeError("can't convert " + keepends.typeName() + " to int");
         } else {
-            throw new org.python.exceptions.TypeError("an integer is required (got type "+keepends.typeName()+")");
+            throw new org.python.exceptions.TypeError("an integer is required (got type " + keepends.typeName() + ")");
         }
 
         org.python.types.List result_list = new org.python.types.List();
-        
-        if (this.value.isEmpty()) return result_list;
+
+        if (this.value.isEmpty()) {
+            return result_list;
+        }
 
         java.lang.String[] result;
         // List of line boundaries from https://docs.python.org/3.4/library/stdtypes.html#str.splitlines
         java.lang.String lineBoundaries = "\\r\\n|\\n|\\r|\\v|\\x0b|\\f|\\x0c|\\x1c|\\x1d|\\x1e|\\x85|\\u2028|\\u2029";
-        java.lang.String[] lineBoundariesList = {"\\r\\n","\\n","\\r","\\v","\\x0b","\\f","\\x0c","\\x1c","\\x1d","\\x1e","\\x85","\\u2028","\\u2029"};
-        
+        java.lang.String[] lineBoundariesList = {"\\r\\n", "\\n", "\\r", "\\v", "\\x0b", "\\f", "\\x0c", "\\x1c", "\\x1d", "\\x1e", "\\x85", "\\u2028", "\\u2029"};
+
         if (!parseKeepEnd) {
             result = this.value.toString().split(lineBoundaries);
             for (java.lang.String w : result) {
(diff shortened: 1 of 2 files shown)


python/common/org/python/types/Str.java
```diff
@@ -1296,14 +1296,78 @@ private long toPositiveIndex(long index) {
...
+    )
+    public org.python.Object splitlines(org.python.Object keepends) {
+        boolean parseKeepEnd;
+        if (keepends == null) {
+            parseKeepEnd = false;
+        } else if (keepends instanceof org.python.types.Bool) {
+            parseKeepEnd = ((org.python.types.Bool) keepends).value;
+        } else if (keepends instanceof org.python.types.Int) {
+            parseKeepEnd = ((org.python.types.Int) keepends).value != 0;
+        } else if (keepends instanceof org.python.types.Float) {
```
cflee: It seems that Int-like objects are accepted: (but not Bool-like ones)

```
class IntLike:
    def __int__(self):
        return 5

x = "abc\ndef"
y = IntLike()
x.splitlines(y)
x.splitlines(0)
```

It might be simpler to call `__int__()` on `keepends`, then test for `!= 0`. I think that will take care of the Bool, Int, Float, and Complex cases.

python/common/org/python/types/Str.java
```diff
@@ -1296,14 +1296,78 @@ private long toPositiveIndex(long index) {
...
+            return result_list;
+        }
+
+        java.lang.String[] result;
+        // List of line boundaries from https://docs.python.org/3.4/library/stdtypes.html#str.splitlines
+        java.lang.String lineBoundaries = "\\r\\n|\\n|\\r|\\v|\\x0b|\\f|\\x0c|\\x1c|\\x1d|\\x1e|\\x85|\\u2028|\\u2029";
+        java.lang.String[] lineBoundariesList = {"\\r\\n", "\\n", "\\r", "\\v", "\\x0b", "\\f", "\\x0c", "\\x1c", "\\x1d", "\\x1e", "\\x85", "\\u2028", "\\u2029"};
+
+        if (!parseKeepEnd) {
+            result = this.value.toString().split(lineBoundaries);
```
cflee: Hmm, is `this.value` already a `java.lang.String`?

tests/datatypes/test_str.py
```diff
@@ -240,6 +240,27 @@ def test_split(self):
                     print(err)
             """)
 
+    def test_splitlines(self):
+        self.assertCodeExecution("""
+            s = 'ab c\\nde fg\\rkl\\r\\nabc'
```
cflee: I think more test cases are necessary to fully exercise the function – things like having two line boundaries in a row, having one or more line boundaries at the end of the string (which the example in the python docs had), empty string, and any other edge cases you can think of.

python/common/org/python/types/Str.java
sngguojie: Thanks for the tip! I'll make that change!

python/common/org/python/types/Str.java
sngguojie: Oh yea, I will remove that. So sorry.

tests/datatypes/test_str.py
sngguojie: Okay! Ill will think of more test cases

sngguojie:
@cflee Regarding this:
```
I think another approach could be to iterate over the string (as a char array?) and check each byte to see whether it's a line ending. There's a special case if it's a \r, then check the next byte to see if it's \n. Then you can compute the start/end indexes (as seems to be done with i and j now) and substring the original string to add to the result list.
```
I attempted your suggestion locally and a major setback I faced is that the python representation of the line boundaries `\x0b` and the java representation of the same line boundary `\u000B`. If keeplines is true, I need to print the `\x0b` character, but if I use the java representation, it will print a new line, which is incorrect. I will research more on this as I think I do not fully understand how to deal with this yet. 

Alternatively, the regex approach seems to pass and I will also try to test more thoroughly.

Thanks again for the feedback!! :)

sngguojie:
@cflee @freakboy3742 
Hi Chiang Fong Lee and Fede Isas, I followed the cpython implementation as suggested and also provided more test cases. Do let me know if it is okay, and I hope to be able to have my first ever contribution to open source soon! - Melvyn

cflee:
@sngguojie The build is failing due to some small Java code style issues – please look into that so we can check the full test output.

> I attempted your suggestion locally and a major setback I faced is that the python representation of the line boundaries `\x0b` and the java representation of the same line boundary `\u000B`. If keeplines is true, I need to print the `\x0b` character, but if I use the java representation, it will print a new line, which is incorrect.

> To solve the above isse, I used a HashMap that takes the java line break character as the key and the string representation in python as the value. When the java line break character is detected, then I will append the python string representation given in the value.

I'm not quite sure I follow this – how is the Python representation involved here? It seems like all we are doing here is manipulating the strings in Java.

I'm quite curious about the choice of using a HashMap to get the String as well – could that be entirely avoided by adjusting the index used to substring from the char array? That way, there's no need to explicitly save/retrieve and append the linebreak char(s), and we wouldn't need those two representations (?) being put side-by-side.

I'm also a little split between having switching the HashMap/HashSet to a static object (since it doesn't change between invocations) or just having a large switch block.

I'll have another look at this after we can confirm that the newly added tests are passing.

sngguojie:
@cflee Thanks for the quick feedback. I have fixed the style errors, sorry about that.

> I'm not quite sure I follow this – how is the Python representation involved here? It seems like all we are doing here is manipulating the strings in Java.

Oh, I ran the tests to see what the outcome should be and manually added the strings in Java.

> I'm quite curious about the choice of using a HashMap to get the String as well – could that be entirely avoided by adjusting the index used to substring from the char array? That way, there's no need to explicitly save/retrieve and append the linebreak char(s), and we wouldn't need those two representations (?) being put side-by-side.

For example, when I have the python test 
```
s1 = 'a\nb'
print(s1.splitlines(True))
```
The expected result should be
```
['a\n', 'b']
```
However, when I attempted the implementation you mentioned (adjusting the index of the substring), the result is
```
['a
', b]
```
Which fails the test (Hence my alternative implementation). I hope my explanation on my choice to not use the indexing is clear. I'm not sure if my method is the best way, and I'm definitely open to learning how to improve!

> I'm also a little split between having switching the HashMap/HashSet to a static object (since it doesn't change between invocations) or just having a large switch block.

Yup, I agree a static object or a large switch block would be more optimized than creating a new HashMap/HashSet every time the function is invoked. But this would depend on whether we establish that the indexing method you mentioned is possible. 

Looking forward! :)

cflee:
@sngguojie I don't have time to dig into this properly right at this moment, but some thoughts to try and unblock this...

My gut feel is that the issues around `['a\x0b', 'b]` and `['a\n', 'b']` in the output are due to a faulty implementation of `Str.__repr__()`, which `print()` on the list invokes. I suspect the adjusting index approach is producing the correct string, but the output is wrong due to `Str.__repr__()`. Maybe you can verify whether this is (or is not!) the case.

You could check the length of the output strings as well, like `'a\n'` must be two bytes long, not three. Is this the case for current implementation?

I think once #429 lands with `ascii()` builtin that could be used in the test as well, since it should escape non-ASCII characters.

sngguojie:
@cflee Okay! Thanks for the direction! I will look into it!

cflee:
`Str.splitlines()` has just been implemented in #566.

However, there's two things in this PR that would still be great contributions:
* the improved `Str.__repr__()` – though it's still missing the quotes support per #568, and we probably should use a StringBuilder instead of a String, it's still an improvement over the existing implementation)
* the understanding of handling all the non-`\r\n` newlines – the current implementation doesn't have this logic yet.

The tests in this PR are also failing on the newly-merged implementation, so some parts could be used a starting point for additional tests for the above two points too.

Since these pieces would be better handled in separate/new PRs, I'm closing this PR first. Thanks!

eliasdorneles:
Ohh, I'm so sorry, I missed this PR while going through the list yesterday !
Thanks for bringing it up @cflee, I agree, +1 !

PR #466, opened 2017-03-23
Add getenv

Nasreen123:
Hi 
This adds implementation for the os.getenv function.  
Thanks @eliasdorneles for pairing with me on this.
Does this look ok?

### DIFF
--- a/tests/stdlib/test_os.py
+++ b/tests/stdlib/test_os.py
@@ -20,4 +20,8 @@ class OsModuleTests(TranspileTestCase):
                 print(os.getenv(1))
             except Exception as e:
                 print(type(e), e)
+            try:
+                print(os.getenv())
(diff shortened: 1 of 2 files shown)


tests/stdlib/test_os.py
```diff
@@ -0,0 +1,23 @@
...
+        self.assertCodeExecution("""
+            import os
+            print(os.getenv('USER'))
+            print(os.getenv('NOT_HERE', default='something'))
+            print(os.getenv('NOT_HERE', 'something'))
+            print(os.getenv('NOT_HERE'))
+            """)
+        self.assertCodeExecution("""
+            import os
+            try:
```
cflee: Could you just add one test for `os.getenv()` (required `key` arg missing)?

PR #469, opened 2017-03-24
Modulo on Str with float operand

TheGamer007:
Refs #36 
Most of the code was directly taken from `java.util.Formatter`, the class that is used when `String.format()` is called, and then modified to suit our purposes. I initially tried extending or creating a wrapper class, but the format specifier 'conversions' are in a private class within the final class. Since Python string formatting does not have a specific Date/Time section ( that is handled by `strftime` and some other methods), the pattern was changed accordingly.

The current working process is to take the string given by the user, convert the flags not in Java into their Java counterparts on a case-by-case basis, and assign the string to a local variable `format_string`. Then, by matching the argument to the flag it is corresponding to, implicit conversions have been handled. Currently, only `float` type is implemented. 

The final string formatting is still done by passing the `format_string` along with `format_args` to `String.format()`


goanpeca:
@TheGamer007 have you find some time to work on this? You will need to rebase this with the latest master and relaunch the CI checks.

TheGamer007:
@goanpeca I definitely plan to rebase and continue as and when time permits, but my work on this will be sporadic due to some other commitments. If that will be an issue, then I am fine with dropping this for now, so that others may work on it, and taking it up again when my schedule is free.

TheGamer007:
I don't think I'll be able to continue work on this. If anyone is interested in taking this up, feel free to continue.

cflee:
Thanks @TheGamer007, closing this PR.

PR #470, opened 2017-03-24
[WIP] Exceptions now working with multiple arguments 

### LINKED ISSUE #352: Exceptions usage fail in unexpected ways

dibyadas:
This PR partially fixes #352. 
Now ValueError and KeyError can be raised with or without arguments.
But this works for only a single argument passed to them. Multiple arguments fail.
I have identified the issue but I need some pointers to solve it.

### DIFF
--- a/python/common/org/python/exceptions/KeyError.java
+++ b/python/common/org/python/exceptions/KeyError.java
@@ -5,7 +5,9 @@ public class KeyError extends org.python.exceptions.LookupError {
 	public KeyError() {
         super();
     }
-
+	public KeyError(org.python.Object[] args) {
+       super(args[0].toString());
+    }
     public KeyError(org.python.Object[] args, java.util.Map<java.lang.String, org.python.Object> kwargs) {
         super(args[0].__repr__().toString());
     }
--- a/python/common/org/python/exceptions/KeyError.java
+++ b/python/common/org/python/exceptions/KeyError.java
@@ -1,6 +1,11 @@
 package org.python.exceptions;
 
 public class KeyError extends org.python.exceptions.LookupError {
+
+	public KeyError() {
+        super();
+    }
+
     public KeyError(org.python.Object[] args, java.util.Map<java.lang.String, org.python.Object> kwargs) {
         super(args[0].__repr__().toString());
     }
--- a/voc/python/ast.py
+++ b/voc/python/ast.py
@@ -617,8 +617,25 @@ class Visitor(ast.NodeVisitor):
             for arg in args:
                 self.visit(arg)
 
-            self.context.add_opcodes(
-                java.Init(exception, *(['Lorg/python/Object;'] * len(args)))
+            for i in args:       #store the arguments passed in local variable #name
+                self.context.add_opcodes(
+                ASTORE_name(str(i)),
+                )
+
+            self.context.add_opcodes(    #declaring array of python objects 
+                ICONST_val(len(args)), 
+                JavaOpcodes.ANEWARRAY('org/python/types/Object'),
+            )
+            for i in args:      #storing the passed objects in the array
+                self.context.add_opcodes(  
+                  JavaOpcodes.DUP(),
+                  ICONST_val(len(args)-args.index(i)-1),
+                  ALOAD_name(str(i)),
+                 JavaOpcodes.AASTORE(),
+                )
+
+            self.context.add_opcodes(     
+                java.Init(exception, '[Lorg/python/Object;')
             )
 
         self.context.add_opcodes(
--- a/python/common/org/python/exceptions/ValueError.java
+++ b/python/common/org/python/exceptions/ValueError.java
@@ -9,6 +9,10 @@ public class ValueError extends org.python.exceptions.Exception {
         super(msg);
     }
 
+	public ValueError(org.python.Object arg) {
+       super(arg.__repr__().toString());
+    }
+
     public ValueError(org.python.Object[] args, java.util.Map<java.lang.String, org.python.Object> kwargs) {
         super(args, kwargs);
     }
--- a/python/common/org/python/exceptions/ValueError.java
+++ b/python/common/org/python/exceptions/ValueError.java
@@ -9,8 +9,8 @@ public class ValueError extends org.python.exceptions.Exception {
         super(msg);
     }
 
-	public ValueError(org.python.Object arg) {
-       super(arg.__repr__().toString());
+	public ValueError(org.python.Object[] args) {
+       super(args[0].toString());
     }
 
     public ValueError(org.python.Object[] args, java.util.Map<java.lang.String, org.python.Object> kwargs) {


dibyadas:
I have almost fixed the issue. Now it works for mutiple arguments. 
This PR needs reviews.

dibyadas:
@eliasdorneles  Thank you for reviewing the PR. Apologies for not being descriptive with the title of the 
PR.Henceforth I will add tests for the issue i am solving.
 This was a test fix I made to check if I am following the right direction. 
As for the test case you mentioned, it fails when you try to access the `e.args` . It returns `e` has no attribute `args` , else leaving the e.args part it works.
With a conversation with @freakboy3742 , he mentioned a case where my way of fixing the issue might run into problems. The case was when a user defines his own exception class , and overrides the \_\_init\_\_() method, my solution might not work.

eliasdorneles:
@dibyadas Well, this is still doesn't fix the multiple arguments issue for exceptions because it's ignoring all but the first argument.
I mean, it doesn't fix this:

```
>>> e = ValueError(3, 4, 5)
>>> str(e)
3
```

A proper solution must handle multiple arguments stored and printed as a tuple, as CPython does:

```
$ python3
Python 3.5.2 (default, Nov 17 2016, 17:05:23) 
[GCC 5.4.0 20160609] on linux
>>> e = ValueError("oi")  # single argument
>>> str(e)
'oi'
>>> e = ValueError("oi", 4)  # multiple arguments
>>> str(e)
"('oi', 4)"
```
That's why I think it makes sense to add an `org.python.types.Tuple args` attribute to BaseException.

dibyadas:
@eliasdorneles Thanks for your views on this. I had temporarily stopped working on this issue because I was working on the Python math module implementation in VOC. I will update the PR within two days fixing the issue along with adding the tests cases for this.

dibyadas:
@eliasdorneles I added the `org.python.types.Tuple args` attribute to BaseException class along with the decorator `@org.python.Attribute` , but I do not know why VOC throws an attribute error when I try to access `args`.  I was trying this with KeyError which inherits from LookupError and it inherits from Exception and then finally from BaseException. Since it did not work in BaseException , I tried adding it in Exception and got the same error. I did the same with LookupError and it too gave that error. Finally when I added it to KeyError , it worked. 
I will resume working on this after a few days because of my end semesters examination.

dibyadas:
Hi @eliasdorneles. Now that my exams are over and my semester too. I can resume the work :slightly_smiling_face: 
I have posted the problem I was facing while fixing this. I would be great if I can get some help in this.

eliasdorneles:
Hm, that's weird that it doesn't work in BaseException, I would expect that to work.
What error did you get? I wonder if it's something that needs fixing.
I'll try to look at this with more patience soon, but please keep digging!

eliasdorneles:
So, I'm giving this a second look, and I'm able to reproduce the problem with trying to access the args attribute.

I tried adding the fix to call using the array of `python..Object`, and then adding this field to different exception classes:

```
    @org.python.Attribute()
    org.python.types.Tuple args = new org.python.types.Tuple();
```
... and finally tried to access it like:
```
try:
    raise ValueError("bla1", "bla2")
except Exception as e:
    print(e.args)
```
If I add the field to `BaseException` or `Exception`, it seems to be completely ignored, throwing an AttributeError:

```
Exception in thread "main" AttributeError: 'ValueError' object has no attribute 'args'
	at org.python.types.Object.__getattribute__(Object.java:310)
	at python.exceptionstest.module$import(exceptionstest.py:4)
	at python.exceptionstest.main(exceptionstest.py)
```
However, if I add the field directly to the children exception, I get an illegal access error:

```
Exception in thread "main" RuntimeError: Illegal access to native field args
	at org.python.java.Field.__get__(Field.java:34)
	at org.python.types.Object.__getattribute_null(Object.java:342)
	at org.python.types.Object.__getattribute__(Object.java:308)
	at python.exceptionstest.module$import(exceptionstest.py:4)
	at python.exceptionstest.main(exceptionstest.py)
```

So, perhaps we can split the work for fixing this, by making this PR about just accepting multiple arguments and building a String representation of the arguments for the `String message` constructor, and then create an issue for the problem of accessing attributes with an isolated test case (since the attribute access probably happens for other kinds of classes as well).

python/common/org/python/exceptions/KeyError.java
```diff
@@ -1,6 +1,13 @@
 package org.python.exceptions;
 
 public class KeyError extends org.python.exceptions.LookupError {
+
+	public KeyError() {
+        super();
+    }
+	public KeyError(org.python.Object[] args) {
+       super(args[0].toString());
```
eliasdorneles: So, assuming we'll fix the `args` attribute access later on, here's what I think we should do:

* add the Tuple args attribute to BaseException
* add a `getMessage()` method to BaseException that checks the size of the Tuple, and builds the proper String message
* change the constructors to set the args tuple attribute

voc/python/ast.py
```diff
@@ -617,8 +617,25 @@ def visit_Raise(self, node):
...
+            for i in args:      #storing the passed objects in the array
+                self.context.add_opcodes(  
+                  JavaOpcodes.DUP(),
+                  ICONST_val(len(args)-args.index(i)-1),
+                  ALOAD_name(str(i)),
+                 JavaOpcodes.AASTORE(),
+                )
+
+            self.context.add_opcodes(     
+                java.Init(exception, '[Lorg/python/Object;')
```
eliasdorneles: This looks like it would work, but instead of creating variable names, it will be better to update the `Array()` helper class in `voc/python/types/java.py` to support creating arrays loading the values from the stack.

That class supports an argument `fill` that is used to fill the array.
We can make it a special value `'STACK'` that would load the values from the stack instead.

Here is what the code would look like:

```
        if self.fill:
            if self.fill == 'STACK':
                for i in range(self.size):
                    context.add_opcodes(
                         # the next two instructions modify the stack like: 1 2 -> 2 2 1
                        JavaOpcodes.DUP_X1(),
                        JavaOpcodes.SWAP(),

                        ICONST_val(i),
                        JavaOpcodes.SWAP(),
                        JavaOpcodes.AASTORE(),
                    )
            else:
                for i in range(self.size):
                    context.add_opcodes(
                        JavaOpcodes.DUP(),
                        ICONST_val(i),
                        self.fill,
                        JavaOpcodes.AASTORE(),
                    )
```

Then we can use this function to build the array to pass to `java.Init`.

Does this make sense?

Feel free to ask any questions, and thanks for getting this far! :)

dibyadas:
I apologize for the late reply. I am travelling back to my home for the summers. It was really difficult for me to find time amidst the rush of packing things and boarding train. 

>  it will be better to update the Array() helper class in voc/python/types/java.py to support creating arrays loading the values from the stack.

I had not really thought about this until now. This seems really good. 

> add a getMessage() method to BaseException that checks the size of the Tuple, and builds the proper String message

`getMessage()`  will act as private helper method? Can this not be done in the constructors?

> Then we can use the Array() helper here to build the array to pass to java.Init.

This will be great in many ways and also avoiding the clutter in `ast.py` file.

> However, if I add the field directly to the children exception, I get an illegal access error:

I got something a bit different. 
I added the args in KeyException like this :- 
```
@org.python.Attribute
	public org.python.Object args;

```
and here is one of the constructors of KeyError :-
```
public KeyError(org.python.Object[] args) {
        super(args);
	System.out.println("in KeyError args");  // debug calls
    }
```
I have put debug calls in LookupError, Exception, BaseException as well. It does tell me that that all necessary constructors are being called. So the `args` should have been initialized , but when I try to access that , I get `None` as its value
Here is my test code :- 
```
try:
	raise KeyError("sdf","sdgsg","sdfds")`
except KeyError as e:
	print(e.args)
```
which outputs `None`.

I will continue digging into this further once I reach my home safely.  Thanks for your help! :)

goanpeca:
@dibyadas have you manage to find some time to work on this :-) ?

dibyadas:
@goanpeca  My apologies for not updating this PR. I got busy with my end semesters examinations. Then I did two internships this summer. Our 3rd sem has started and I'm getting more occupied by the day. I would definitely like to continue this. But I guess I need some time.  Apologies again for any inconvenience.

eliasdorneles:
Hello @dibyadas -- this was recently fixed in #702.
Thank you so much for getting the ball rolling, I referred back a few times to our discussion in here while working on the fix. :)

PR #472, opened 2017-03-24
added rpartition

sohit-nayak:
added rpartition and tests.

### DIFF
--- a/python/common/org/python/types/Str.java
+++ b/python/common/org/python/types/Str.java
@@ -1190,10 +1190,34 @@ public class Str extends org.python.types.Object {
                     "\n" +
                     "Search for the separator sep in S, starting at the end of S, and return\n" +
                     "the part before it, the separator itself, and the part after it.  If the\n" +
-                    "separator is not found, return two empty strings and S.\n"
+                    "separator is not found, return two empty strings and S.\n",
+            args = {"sep"}
     )
-    public org.python.Object rpartition() {
-        throw new org.python.exceptions.NotImplementedError("rpartition() has not been implemented.");
+    public org.python.Object rpartition(org.python.types.Object sep) {
+        java.lang.String sepStr = ((org.python.types.Str) sep).value;
+        java.lang.String str_one = new java.lang.String();
+        java.util.List<org.python.Object> tuple = new java.util.ArrayList<org.python.Object>();
+        if (sepStr.equals("")) {
+            throw new org.python.exceptions.ValueError("empty separator");
+        }
+        java.lang.String[] split = this.value.split(sepStr);
+        if (split.length == 1) {
+            tuple.add(new org.python.types.Str(""));
+            tuple.add(new org.python.types.Str(""));
+            tuple.add(new org.python.types.Str(this.value));
+            return new org.python.types.Tuple(tuple);
+        }
+        for (int i = 0; i < split.length - 1; i++) {
+            if (i == split.length - 2) {
+                str_one += split[i];
+                break;
+            }
+            str_one += split[i] + sepStr;
+        }
+        tuple.add(new org.python.types.Str(str_one));
+        tuple.add(new org.python.types.Str(sepStr));
+        tuple.add(new org.python.types.Str(split[split.length - 1]));
+        return new org.python.types.Tuple(tuple);
     }
 
     @org.python.Method(
--- a/python/common/org/python/types/Str.java
+++ b/python/common/org/python/types/Str.java
@@ -1201,7 +1201,19 @@ public class Str extends org.python.types.Object {
             throw new org.python.exceptions.ValueError("empty separator");
         }
         java.lang.String[] split = this.value.split(sepStr);
+        if(this.value.length() < sepStr.length()) {
+            tuple.add(new org.python.types.Str(""));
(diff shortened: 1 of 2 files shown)


tests/datatypes/test_str.py
```diff
@@ -574,6 +574,18 @@ def test_rjust(self):
                 print(err)
             """)
 
+    def test_rpartition(self):
+        self.assertCodeExecution("""
+            Stri = "Hello World!"
```
freakboy3742: flake8 won't catch the problems with test code, but your test code should be flake8 compliant, too. Variable names shouldn't start with a capital letter.

tests/datatypes/test_str.py
```diff
@@ -574,6 +574,18 @@ def test_rjust(self):
...
+            print(Stri.rpartition("H"))
+            print(Stri.rpartition(" "))
+            print(Stri.rpartition("l"))
+            try:
+                print(Stri.rpartition(""))
+            except ValueError as err:
+                print(err)
```
freakboy3742: This is a good start to testing, but there's a bunch of other tests needed - most notably, partitioning on a string that doesn't exist, and partitioning on a multi-character string.

sohit-nayak:
@freakboy3742 I added some more test cases, and found out that the tailing cases were'nt handled properly. I have modified the code, and guess its good to go now! :smile:

tests/datatypes/test_str.py
```diff
@@ -574,6 +574,23 @@ def test_rjust(self):
...
+            print(st.rpartition("H"))
+            print(st.rpartition(" "))
+            print(st.rpartition("l"))
+            print(st.rpartition("q"))
+            print(st.rpartition("lo"))
+            print(st.rpartition("!"))
+            print(st.rpartition("ld!"))
+            print("voc".rpartition("foobar"))
+            try:
+                print(st.rpartition(""))
```
cflee: It's great that an empty string is being tested here, but could you also test the cases of no parameter being passed (`rpartition()`), and passing in some non-string types?

It might also be worth testing whether it accepts an object that implements `__str__()`, to be thorough.

Also, you should test the case where `st = ""` as well.

python/common/org/python/types/Str.java
```diff
@@ -1190,10 +1190,46 @@ private long toPositiveIndex(long index) {
...
-    public org.python.Object rpartition() {
-        throw new org.python.exceptions.NotImplementedError("rpartition() has not been implemented.");
+    public org.python.Object rpartition(org.python.types.Object sep) {
+        java.lang.String sepStr = ((org.python.types.Str) sep).value;
+        java.lang.String str_one = new java.lang.String();
+        java.util.List<org.python.Object> tuple = new java.util.ArrayList<org.python.Object>();
+        if (sepStr.equals("")) {
+            throw new org.python.exceptions.ValueError("empty separator");
+        }
+        java.lang.String[] split = this.value.split(sepStr);
```
cflee: What happens if you pass in some separator string that can be interpreted as a regex? `java.lang.String.split()` will interpret it as a regex, but will python's `Str.rpartition()`?

If it shouldn't be interpreted as a regex, could you use something like `java.lang.String.lastIndexOf()` instead?

python/common/org/python/types/Str.java
```diff
@@ -1190,10 +1190,46 @@ private long toPositiveIndex(long index) {
...
+            tuple.add(new org.python.types.Str(""));
+            tuple.add(new org.python.types.Str(this.value));
+            return new org.python.types.Tuple(tuple);
+        }
+        for (int i = 0; i < split.length - 1; i++) {
+            if (i == split.length - 2) {
+                str_one += split[i];
+                break;
+            }
+            str_one += split[i] + sepStr;
```
cflee: This loop seems rather inefficient. If the string is split into many parts (due to having many separator inside), each part needs to be individually concatenated to form the string. One way of making it more efficient would be to find the length of the part-after-the-separator and just do a `java.lang.String.substring()`.

python/common/org/python/types/Str.java
```diff
@@ -1190,10 +1190,46 @@ private long toPositiveIndex(long index) {
...
+        }
+        java.lang.String[] split = this.value.split(sepStr);
+        if (this.value.length() < sepStr.length()) {
+            tuple.add(new org.python.types.Str(""));
+            tuple.add(new org.python.types.Str(""));
+            tuple.add(new org.python.types.Str(this.value));
+            return new org.python.types.Tuple(tuple);
+        }
+        if (split.length == 1) {
+            if ((this.value.substring(this.value.length() - sepStr.length())).equals(sepStr)) {
```
cflee: This condition seems a bit strange. In which case will it be true? It seems like you're taking the string of the last `sepStr.length()` chars and comparing it with `sepStr` itself?

python/common/org/python/types/Str.java
sohit-nayak: hmm, I too think use of split is wrong here since it accepts regex, but rpartition doesnt.

sohit-nayak:
@cflee I have opened a fresh PR #512 with the requested changes. Please have a look. Thanks!

PR #484, opened 2017-03-27
Implemented l/rstrip, strip function in Str.java , edited test_str accordingly , meets stylecheck norms

gautamworah96:
Fixed the buggy rstrip , lstrip function (had to completely rewrite it) , also implemented strip() for strings


cflee:
@gautamworah96 Please note that all three methods fail when `None` is passed as the chars parameter.

PR #488, opened 2017-03-28
Miscellaneous fixes for Toga on Android

freakboy3742:


voc/python/ast.py
```diff
@@ -986,27 +995,71 @@ def _create_function(self, node, func_name, decorator_list):
...
         # Now actually define the function.
         for decorator in decorator_list:
-            self.visit(decorator)
-            self.context.add_opcodes(
-                JavaOpcodes.CHECKCAST('org/python/Callable'),
-                java.Array(1),
-                JavaOpcodes.DUP(),
-                JavaOpcodes.ICONST_0(),
-            )
+            # The @super decorator on __init__() is a special case. It tells us how
```
eliasdorneles: Wouldn't it be better to move this code to `voc/python/klass.py` ?
Would it make sense to create a `self.context.add_constructor_with_args()` there that we could call here? (That was the approach I was going for this morning).

docs/internals/release_process.rst
```diff
@@ -0,0 +1,49 @@
+Release Process
```
eliasdorneles: This is great! :heart:

eliasdorneles:
So, I tried to use this with briefcase for the following app:

```
import android
from android.util import Log
from android.widget import LinearLayout
from android.widget import Button
from android.widget import TextView
from android.view import Gravity
from android.view import MotionEvent
from android.graphics import Paint, Path, Bitmap, Canvas
import android.view


class CustomView(extends=android.view.View):
    @setup({context: android.content.Context})
    def __init__(self, context):
        print('hello from init')


class MyApp:
    def __init__(self):
        self._activity = None

    def link(self, activity):
        self._activity = activity
        print('linked activity', self._activity)

    def onCreate(self):
        print('onCreate, activity is', self._activity)
        parent_layout = LinearLayout(self._activity)
        self.drawing = CustomView(self._activity)
        parent_layout.addView(self.drawing)
        self._activity.setContentView(parent_layout)


print('starting app...')
app = MyApp()
activity = android.PythonActivity.setListener(app)
app.link(activity)
```

and I'm getting the following error:

```
E/AndroidRuntime( 3742): Caused by: AttributeError: 'org.python.java.Type' object has no attribute 'setup'
E/AndroidRuntime( 3742): 	at python.drawingapp.app.CustomView.class$init(app.py:13)
E/AndroidRuntime( 3742): 	at python.drawingapp.app.__init__.module$import(app.py:12)
E/AndroidRuntime( 3742): 	at python.drawingapp.app.__init__.main(app.py)
E/AndroidRuntime( 3742): 	at java.lang.reflect.Method.invoke(Native Method)
E/AndroidRuntime( 3742): 	at java.lang.reflect.Method.invoke(Method.java:372)
E/AndroidRuntime( 3742): 	at android.PythonActivity.onCreate(PythonActivity.java:37)
```

Is this reproducible for you?

freakboy3742:
@eliasdorneles Yup - completely reproducible. And completely fixable if you use `@super` instead of `@setup` :-)

Which raises the fact that non-existent decorators aren't raising the right error, but that's a whole other problem.

voc/python/ast.py
freakboy3742: Makes sense. The code won't be re-used anywhere, but it would certainly help with readability.

freakboy3742: On closer inspection, I'm not sure it does help. The call to push/pop context happens quite deep in this block, and has to happen *after* the `InitMethod` is constructed. This means you have to pass the visitor to the klass instance, which would be first place in the code this happens, and would also introduce a circular dependency. 

I've added a minor change to improve error reporting,  but unless you can see a better way to factor this out, I'm not sure it's possible to move the logic to `klass.py`

eliasdorneles:
OMG, haha!
I'm sorry for the false alarm!

PR #490, opened 2017-03-28
Implemented keys,values for dict and radd,rsub for bool

gautamworah96:
Keys and values needs to return a dict_view ..Currently I have returned a new list based on the previous implementation of items() ..radd,rsub also implemented for bool..Test cases not written


gautamworah96:
@freakboy3742 I cannot think of any test case for radd , rsub for bool ..Could you give a small example?

gautamworah96:
Thanks for the feedback , will work on it.

cflee:
Closing this PR for now due to inactivity – if you would like us to take a look at this again, please re-open it and address the earlier comments!

Also, `str.replace()` was implemented in #501, but we would welcome a contribution (with tests) for the other methods!

PR #494, opened 2017-03-29
Python math module implementation complete

dibyadas:
Except fsum , all other functions of math module added. 
I will implement fsum and add tests within two days.

### DIFF
--- a/python/common/python/math/__init__.java
+++ b/python/common/python/math/__init__.java
@@ -1,6 +1,5 @@
 package python.math;
 
-import java.util.*;
 import java.math.BigInteger;
 import org.python.types.Int;
 import org.python.types.Float;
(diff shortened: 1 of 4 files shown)


dibyadas:
Python math module implementation complete. Ready for review.

dibyadas:
@cflee Thanks a lot for reviewing my PR. 
I apologize for the size of this PR and I will surely keep this in mind in the future.

- About the code style.
  - I chose to import the classes because then it would be less cumbersome to write the code. And I thought this made the code more readable and clean?
  - I will change the annotations to agree with the rest of the code.
  - For the tests, I followed the existing tests suite for `str` and others. I do agree that the test for a single function can be inside one `assertCodeExecution()` block but instead I chose to follow the existing format. I shall definitely change this if this helps in faster execution of the test suite.
  - It seems I have not properly used whitespaces. I will update the PR after adding those properly.
- The `erf()` and `gamma()` implementation are from the Jython modules.I have attached the files here.
  - I tried different test cases and found the exceptions raised by python and implemented them, hence the same exceptions are raised in VOC as in python.
- `tofloatvalue()` is indeed a helper method that is supposed to be `private` . I did not notice this until now. I shall change this and bring consistency in the naming of helper methods. 
- I will update the test cases for `factorial()` and other such methods(if any) to check for non-error output as well.
- `ceil(x)`,`floor(x)` & `trunc(x)` are the methods in python math module that delegate to `x.__ceil__()`  and so on respectively.  I will update these methods too.

The code for `erf()` and `gamma()` from jython modules:-
[math_gamma.java.zip](https://github.com/pybee/voc/files/908516/math_gamma.java.zip)
[math_erf.java.zip](https://github.com/pybee/voc/files/908517/math_erf.java.zip)

cflee:
@dibyadas 

> The code for erf() and gamma() from jython modules

Ok, the size of this PR makes a lot more sense since much/most (?) of the `math` module is ported from Jython. Assuming the porting went correctly, I think the main issues will be with the tests – we need to make sure the method implementations _are_ correct, and _stay_ correct. I'm not sure if the test cases can be derived from Jython as well?

> I chose to import the classes because then it would be less cumbersome to write the code. And I thought this made the code more readable and clean?

Yes, I agree that it _usually_ makes things more readable.

However, for cases like `Object`, most folks who are familiar with Java will expect that to be `java.lang.Object`, not `org.python.types.Object`. Or for `List`, there's a fair amount of VOC code that interacts with both `java.util.List` (when making tuples) and `org.python.types.List`, and I think being non-ambiguous is very helpful.

> For the tests, I followed the existing tests suite for str and others.

Yup, that's the case, but I'm trying to slowly change this by asking new contributions to have fewer `assertCodeExecution()` where possible, for each `test_abc()` part.

@freakboy3742 Is there anything we need to do to comply with making a derivative work from Jython under the PSF License? As far as I can tell [the only requirement for derivative works](http://www.jython.org/license.html) is "a brief summary of the changes made".

freakboy3742:
@cflee Your analysis matches my understanding. We could use Jython code as the basis for functionality in VOC, as long as we give credit and provide a "brief summary" of any changes we make. That particular turn of phrase is delightfully vague, but shouldn't be too hard to comply with in both letter and spirit.

dibyadas:
@cflee I have made changes according to your suggestions. Interestingly, the factorial implementation is somewhat faulty. This implementation works only for calculating factorial of small numbers, up-to 12 or something. It returns 0 for numbers greater than that. While CPython can go much more than that. This went completely unnoticed. Probably, this happened because I missed the non-error test cases for factorial functions. I did a bit of research and I found some good algorithms like [Poor man's algorithm](http://www.luschny.de/math/factorial/csharp/FactorialPoorMans.cs.html) and [this](http://www.geeksforgeeks.org/factorial-large-number/) which can be used to calculate factorial of large numbers in Java. I guess the PR for implementing this will be a bit long. So I think it will be better if this PR is merged before that is opened.
This PR is ready for another review.

dibyadas:
Hi @cflee. This PR has been lying around for a long time. The implementation is almost complete and is ready for another review. I will implement the `factorial` method as soon as possible and make another PR.

python/common/python/math/__init__.java
```diff
@@ -0,0 +1,840 @@
...
+            throw new org.python.exceptions.ValueError("math domain error");
+        } else if (!isIntegral(v)) {
+            throw new org.python.exceptions.ValueError("math domain error");
+        } else {
+            long value = (long) v;
+            BigInteger bi = new BigInteger(Long.toString(value));
+            for (long l = value - 1; l > 1; l--) {         // this implementation doesn't work for large numbers
+                bi = bi.multiply(new BigInteger(Long.toString(l)));  // this needs work
+            }
+            return new org.python.types.Int(bi.intValue());
```
cflee: It seems that this should be using `BigInteger.longValue()` instead of `intValue()`, since the Python Int type actually wraps a Java `long`. Would this solve the factorial 'big number' issue?

python/common/python/math/__init__.java
```diff
@@ -0,0 +1,840 @@
+package python.math;
+
+import java.math.BigInteger;
+import java.math.BigDecimal;
+import python.errorfunction.math_erf;
+import python.gammafunction.math_gamma;
+
```
cflee: As discussed, it's necessary to insert a credit to the original Jython source, since it's definitely derived from it. I think it'd be best to both have an attribution statement, and also either a reference or a direct link to their repo of the specific version/file that is the source.

This should be done for the erf/gamma files as well.

python/common/python/math/__init__.java
```diff
@@ -0,0 +1,840 @@
+package python.math;
```
cflee: This file and package needs to be re-organised per #517, since there are no sub-modules.

python/common/python/math/math_gamma.java
```diff
@@ -0,0 +1,242 @@
+package python.gammafunction;
```
cflee: Should this gamma and the erf class be a separate package? I'm not sure I understand why it's done this way.

python/common/python/math/__init__.java
```diff
@@ -0,0 +1,840 @@
...
+        }
+    }
+
+    /*@org.python.Method(
+            __doc__ = "",
+            args = {"arg"}
+    )
+    public static org.python.Object trunc(org.python.Object arg) {
+        return arg.__trunc__();
+    }*/
```
cflee: Is this method supposed to be implemented or a WIP?

Also, too many empty lines after this block.

python/common/python/math/__init__.java
```diff
@@ -0,0 +1,840 @@
...
+    public static org.python.Object factorial(org.python.Object arg) {
+        double v = 0.0;
+        if (arg instanceof org.python.types.Float) {
+            v = ((org.python.types.Float) arg).value;
+        } else if (arg instanceof org.python.types.Int) {
+            v = (new org.python.types.Float(((org.python.types.Int) arg).value)).value;
+        } else {
+            throw new org.python.exceptions.TypeError("an integer is required (got type " + arg.typeName() + ")");
+        }
+        if ((v - ((int) v)) != 0.0) {
```
cflee: Is this effectively the same test as `isIntegral()`? If not, can you ensure that both this and the `isIntegral()` code paths are exercised in the test cases?

python/common/python/math/__init__.java
```diff
@@ -0,0 +1,840 @@
...
+            throw new org.python.exceptions.ValueError("factorial() only accepts integral values");
+        } else if (v <= 0.0) {
```
cflee: This seems to be redundant since it overlaps with both `v == 0.0` and `v < 0.0` case, but which behaviour is correct? All these edge cases (including ones like `v == 1.0`) should be tested in the test cases as well.

python/common/python/math/__init__.java
```diff
@@ -0,0 +1,840 @@
...
+            throw new org.python.exceptions.ValueError("factorial() not defined for negative values");
+        } else if (v == 0.0 || v == 1.0) {
+            return new org.python.types.Float(1);
+        } else if (v < 0.0 || isnan(v) || isinf(v)) {
```
cflee: Is there a test case which can make `isnan()` be true without being caught and resulting in TypeError thrown in line 704 above?

python/common/python/math/__init__.java
```diff
@@ -0,0 +1,840 @@
...
+        }
+        if (w == 0.0) {
+            throw new org.python.exceptions.ValueError("math domain error");
+        }
+        if (isinf(v) && w == 0.0) {
+            throw new org.python.exceptions.ValueError("math domain error");
+        }
+        return new org.python.types.Float(v % w);
+    }
+
```
cflee: It would be great if the spacing between all the methods could be made consistent!

python/common/python/math/__init__.java
```diff
@@ -0,0 +1,840 @@
...
+    public static org.python.Object log(org.python.Object v, org.python.Object base) {
+        double b = 0.0;
+        if (base instanceof org.python.types.Float || base instanceof org.python.types.Int) {
+            b = toFloatValue(base);
+        } else if (base == null) {
+            b = 0.0;
+        }
+        double doubleValue;
+        doubleValue = log(toFloatValue(v));
+        return (b == 0.0) ? new org.python.types.Float(doubleValue) : new org.python.types.Float(applyLoggedBase(doubleValue, b));
```
cflee: Could the conditional be moved inside the `Float()` constructor args? Similar to line 203 above

python/common/python/math/__init__.java
dibyadas: It been long since I followed this. I have to look at this carefully to understand the current changes. 
I've gotten quite busy with some other work. I'll try to work on this PR as soon as I get time.

python/common/python/math/__init__.java
dibyadas: Sure. I agree.

python/common/python/math/__init__.java
dibyadas: I completely overlooked this, since the PR passed the checkstyle. I'll try to bring more consistency,.

python/common/python/math/__init__.java
dibyadas: Definitely. It'll shorten the code.

python/common/python/math/__init__.java
dibyadas: Yes. They are the same.

python/common/python/math/__init__.java
dibyadas: Oh. This is wrong. 0! is 1 and this conditional is only for catching negative integers. This will be fixed.

python/common/python/math/__init__.java
dibyadas: I don't think so. Python successfully calculated & displayed ridiculously large numbers without any problem and that's because of the algorithm it uses to do so. It uses partial sums to calculate the factorials. Implementing requires a good understanding of it.

python/common/python/math/__init__.java
dibyadas: It is supposed to be implemented. Since the `__trunc__` method of Int type hadn't been implemented (I don't know if it's done), hence I commented this out, to be left for further discussion.

python/common/python/math/math_gamma.java
dibyadas: The gamma and erf implementation are large in their own respect. I don't think its wise to include this in the main file. This package acts as helpers package.

dibyadas:
@cflee Apologies for being late. I've gotten quite busy with some other things. 
As I said earlier, the factorial implementation will be big. To correctly work like CPython's does, it'll need quite a bit of work. I think its best to merge this PR (after the required changes have been made), and keep  the factorial PR different. Its just my opinion. I'll try to fix and push the other changes and notify you.

With Best Regards,
Dibya Prakash Das

dibyadas:
@eliasdorneles  I don't think so. I've become a bit occupied at this moment. Also, there have been some changes that I haven't followed. It'll take me some time to understand them. Truly sorry about this. :(  
You can close this PR if you wish.

eliasdorneles:
@dibyadas No worries at all! Thanks for contributing anyway -- even if this isn't merged, it was a good discussion! =)
I'll close it, anyone wanting to continue this feel free to create a new one.

PR #496, opened 2017-03-30
Added Contains, NotContains method in FrozenSet

ASP1234:
**Changes**

- Added contains, not Contains method in FrozenSet

- Added tests for the same

**Testing**

- Ant build successful

-  test_frozenSet successful

### DIFF
--- a/tests/datatypes/test_frozenset.py
+++ b/tests/datatypes/test_frozenset.py
@@ -41,26 +41,31 @@ class FrozensetTests(TranspileTestCase):
         # Normal Contains
         self.assertCodeExecution("""
             x = frozenset()
-            print(x)
+            print(1 in x)
 
             x = frozenset({'a'})
             print('a' in x)
+            print('b' in x)
 
             y = {1,2}
             x = frozenset(y)
             print(2 in x)
+            print(3 in x)
 
             y = ['c']
             x = frozenset(y)
             print('c' in x)
+            print('a' in x)
 
             y = (1,2)
             x = frozenset(y)
             print(3 in x)
+            print(1 in x)
 
-            y ='a'
+            y = 'a'
             x = frozenset(y)
             print('e' in x)
+            print('a' in x)
 
             """)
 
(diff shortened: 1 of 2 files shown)


tests/datatypes/test_frozenset.py
```diff
@@ -37,6 +37,60 @@ def test_creation(self):
...
+        self.assertCodeExecution("""
+            x = frozenset()
+            print(x)
+
+            x = frozenset({'a'})
+            print('a' in x)
+
+            y = {1,2}
+            x = frozenset(y)
+            print(2 in x)
```
cflee: For each case, could there be something that tests both true/false cases, just to make sure they're working correctly? So for this one, perhaps `2` and `3`, or something like that.

(Also, is there much difference from the 'a' case above?)

tests/datatypes/test_frozenset.py
```diff
@@ -37,6 +37,60 @@ def test_creation(self):
 
             """)
 
+    def test_contains(self):
+        # Normal Contains
+        self.assertCodeExecution("""
+            x = frozenset()
+            print(x)
```
cflee: I'm not sure what this is testing, the repr should be tested elsewhere. Should this be ensuring that anything `in x` will be false?

tests/datatypes/test_frozenset.py
```diff
@@ -37,6 +37,60 @@ def test_creation(self):
...
+
+            y = ['c']
+            x = frozenset(y)
+            print('c' in x)
+
+            y = (1,2)
+            x = frozenset(y)
+            print(3 in x)
+
+            y ='a'
```
ASP1234: Done 👍

tests/datatypes/test_frozenset.py
ASP1234: Done 👍

tests/datatypes/test_frozenset.py
ASP1234: Changed to print(1 in x) to ensure nothing is in x. 😄

cflee:
Thanks! :sparkles:

PR #497, opened 2017-03-30
Fix complex.__pow__ to handle 0+0j raised to x+0j and negative ints

soandrew:
There was a typo in the helper function `calCPow` of `complex.__pow__` that affected handling of (0 + 0j) base. For example code like the following produced the wrong output.
```
x = 0 + 0j
y = 1 + 0j
print(x ** y)
```
Output was `(1+0j)` when it should have been `0j`


freakboy3742 (changes requested):
Is there any way to add a test for this?

soandrew:
@freakboy3742 I believe that instead of explicitly writing a test case for this, the best way to implement tests would be to add `0+0j` and `1+0j` to the `complex` list in `SAMPLE_DATA` in `tests/utils.py`. This way we make use of the existing test code generation. 

However I ran into precision issues which I wrote about in #499.

Do you want me to add those changes to this PR? I could add `test_power_complex` to the list of `not_implemented` methods in order to have it generate an `expected failure` message and have the Travis check pass.

cflee:
@soandrew It seems like the `r.real.value = 0.0;` fix is in #751 which was just merged, but is there still an issue with the missing `0.0 to a negative or complex power` ZeroDivisionError? If yes, we'd be happy to merge a fix (with tests) for that!

soandrew:
@cflee Sorry for the delayed reply. Indeed there still is an issue when `0+0j` is raised to a negative `int` which causes an error message of `complex division by zero` to be thrown instead of the proper `0.0 to a negative or complex power`.

I made a fix and added explicit tests for the `(0+0j) ** (1+0j)` and `(0+0j) ** -1)` cases which were causing these specific issues.

phildini:

PR #498, opened 2017-03-30
Implement some methods for Set

deep110:


deep110:
@cflee @eliasdorneles there are many tests that are passing but pytest is not giving unexpected success, so should I remove that tests from test_set or let it be that way. 
Other than that this PR is ready for review.

deep110:
@cflee @eliasdorneles Hi, does anything more need to be done in this PR.
Thanks!!

python/common/org/python/types/Set.java
```diff
@@ -416,9 +455,12 @@ public boolean __setattr_null(java.lang.String name, org.python.Object value) {
             args = {"other"}
     )
     public org.python.Object intersection(org.python.Object other) {
-        java.util.Set set = ((Set) this.copy()).value;
-        set.retainAll(((Set) other).value);
-        return new Set(set);
+        try {
+            org.python.types.Set otherSet = new org.python.types.Set(new org.python.Object[] {other}, null);
```
freakboy3742: It would be a lot more efficient to check whether `other` is already a set - and if it is, pass it unchanged to `__and__`, rather than universally duplicating it.

cflee:
@deep110 I think it'd be better to only mark the relevant tests as now implemented/passing, so that we don't have many other unrelated changes in this PR – thanks!

deep110:
@freakboy3742 Thanks for the review. 
I have added type check for the set, and also updated difference method to take any iterable as mentioned in docs, I have added test to ensure its working.

PR #504, opened 2017-04-01
Fixes to get colosseum working on Android.

freakboy3742:


python/common/org/python/types/Function.java
```diff
@@ -19,6 +19,14 @@
     java.util.Map<java.lang.String, org.python.Object> default_kwargs;
     org.python.types.Closure closure;
 
+    public static java.lang.String protect_reserved_words(java.lang.String arg) {
```
eliasdorneles: Hey!
I tried out running the code from this PR, and everything seems fine.
About the code, I was trying to figure out where this method is being used, but couldn't find -- could you clarify?
Thanks!

freakboy3742: Oversight. It was written to make the arguments in org.Python exactly match the official standards - but "default" is a Java keyword, so you can't name an argument "default". However, it turns out this conversion wasn't needed due to the way Java handles argument rollout. I'll delete this method.

PR #508, opened 2017-04-04
Added copy method in FrozenSet

ASP1234:
**Changes**

- Added shallow copy method in FrozenSet

**Reference**

- [copy method in Immutable types](http://stackoverflow.com/questions/37100944/i-think-immutable-types-like-frozenset-and-tuple-not-actually-copied-what-is-th)

**Testing**

- Ant build successful

- FrozenSet tests successful (including newly added copy tests)


python/common/org/python/types/FrozenSet.java
```diff
@@ -124,4 +124,11 @@ public FrozenSet(org.python.Object[] args, java.util.Map<java.lang.String, org.p
     public org.python.Object __not_contains__(org.python.Object other) {
         return new org.python.types.Bool(!this.value.contains(other));
     }
+
+    @org.python.Method(
+            __doc__ = "Return a shallow copy of a FrozenSet."
+    )
+    public org.python.Object copy() {
+        return this;
```
freakboy3742: This isn't a *copy* - it's an internal reference. If you have a set, then copy it, modifying the original shouldn't modify the copy, or vice versa.

ASP1234: @freakboy3742 It's an immutable Set. So, making a copy of the original leads to memory waste. That's why a reference is returned.

Refer : [http://stackoverflow.com/questions/37100944/i-think-immutable-types-like-frozenset-and-tuple-not-actually-copied](http://stackoverflow.com/questions/37100944/i-think-immutable-types-like-frozenset-and-tuple-not-actually-copied-what-is-th)

eliasdorneles: Yup, that's also what CPython does:

```
>>> fs = frozenset()
>>> print(fs.copy.__doc__)
Return a shallow copy of a set.
```

eliasdorneles: Erm, I meant:

```
>>> fs = frozenset([1, 2, 3])
>>> fs
frozenset({1, 2, 3})
>>> fs2 = fs.copy()
>>> id(fs)
140687563610376
>>> id(fs2)
140687563610376
```

PR #509, opened 2017-04-04
Comparison operators implemented for Frozenset

deep110:
Implemented gt, ge, lt, le and eq methods.


deep110:
@ASP1234 For Collections like Arraylist it is O(n^2), but for HashSet it is specifically implemented and is generally O(1) or worst case O(logn).
http://stackoverflow.com/questions/10199772/what-is-the-cost-of-containsall-in-java

I think, Java Collections would be as optimised as possible, so it makes no sense for same python function to look further into optimisations. Though I may be wrong here, but just a thought.

PR #510, opened 2017-04-06
Implemented __reversed__ for list

McMastS:


tests/datatypes/test_list.py
```diff
@@ -574,6 +574,19 @@ def __eq__(self, other):
...
+        self.assertCodeExecution("""
+            list = [123, 'xyz', 'abc']
+            reversedList = ['abc', 'xyz', 123] # to check equality
+            l = []
+            reverse = list.__reversed__()
+            try:
+                while(True):
+                    l.append(next(reverse))
+            except StopIteration:
+                print(l == reversedList)
```
cflee: Could we also add a test to ensure that `list` is not modified?

McMastS:
@cflee I've added the test, and the build is passing as well.

PR #511, opened 2017-04-06
[WIP] Add hypothesis to some str methods

eliasdorneles:
This is a WIP and not ready to merge, but I've been playing with hypothesis lately with help from @damedebugger (hi!).

I don't know if Str.java is the best place to start, but I keep finding new bugs for the existing methods so it'll take a little while for this to be ready.
It might make sense to split this into separate PRs, but as I also want to find a good way of reusing the strategies for other functions I'll continue working on this branch for now.

I hope this will be handy for ensuring quality of the patches, I know I'll be glad to not have to think so hard to cover the edge cases. :)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,3 +14,4 @@ tests/temp
 .pydevproject
 tests/tmp*
 .cache/
+.hypothesis
--- a/tests/datatypes/test_str.py
+++ b/tests/datatypes/test_str.py
@@ -3,7 +3,10 @@ from hypothesis import given, example, strategies as st, note
 import string
 
 
-DIGITS = '0123456789₀₁₂₃₄₅₆₇₈₉⁰¹²³⁴⁵⁶⁷⁸⁹'
+# TODO: replace with line below when unicode numeric is fixed for subscript/superscript
+DIGITS = '0123456789'
+# DIGITS = '0123456789₀₁₂₃₄₅₆₇₈₉⁰¹²³⁴⁵⁶⁷⁸⁹'
+
 HEXDIGITS = string.hexdigits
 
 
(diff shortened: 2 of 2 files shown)


eliasdorneles:
After thinking more about this, trying out things with @mistermocha and talking briefly with @freakboy3742, I now believe it will make more sense to limit the hypothesis usage for development and not on the regular Travis builds, for two reasons:

* it would increase the build time greatly
* it would find many problems that we have no clear solution yet, like the discrepancy between different versions of Unicode data

For reference, here are the Unicode data versions of CPython:

```
$ $(pythonz locate 3.4.4) -c 'import unicodedata; print(unicodedata.unidata_version)'
6.3.0
$ $(pythonz locate 3.5.3) -c 'import unicodedata; print(unicodedata.unidata_version)'
8.0.0
$ $(pythonz locate 3.6.1) -c 'import unicodedata; print(unicodedata.unidata_version)'
9.0.0
```

In the JVM land, [Java 7 uses 6.0.0](https://docs.oracle.com/javase/7/docs/api/java/lang/Character.html), [Java 8 uses 6.2.0](https://docs.oracle.com/javase/8/docs/api/java/lang/Character.html), [Java 9 will be 7.2.0](http://openjdk.java.net/jeps/227).

These causes differences some subtle differences of behavior for methods like `isupper`/`islower` and I believe `isdigit` as well.

Zac-HD:
While the issues noted above are real, I'd also like to note that you can choose how many examples Hypothesis will try - 2000 might be unacceptably slow, but what about 50?  You can also use it for some types but not others - Unicode is always tricky, but numeric and simple collection types could benefit enormously.

As noted on #580, I'd be happy personally and as a maintainer of Hypothesis to assist anyone who wants to take this on.

phildini:

PR #520, opened 2017-04-20
some set method improvements

Saurabh--Kumar:
modified union(), update(), difference(), difference_update(), intersection_update() and intersection() methods to accept variable number of arguments.

### DIFF
--- a/python/common/org/python/types/Set.java
+++ b/python/common/org/python/types/Set.java
@@ -313,12 +313,13 @@ public class Set extends org.python.types.Object {
         throw new org.python.exceptions.TypeError("unsupported operand type(s) for &: '" + this.typeName() + "' and '" + other.typeName() + "'");
     }
 
-    // @org.python.Method(
-    //     __doc__ = ""
-    // )
-    // public org.python.Object __xor__(org.python.Object other) {
-    //     throw new org.python.exceptions.NotImplementedError("__xor__() has not been implemented");
-    // }
+     @org.python.Method(
+         __doc__ = ""
+     )
+     public org.python.Object __xor__(org.python.Object other) {
+         
+         throw new org.python.exceptions.NotImplementedError("__xor__() has not been implemented");
+     }
 
     @org.python.Method(
             __doc__ = ""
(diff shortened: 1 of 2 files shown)


python/common/org/python/types/Set.java
```diff
@@ -417,28 +417,52 @@ public boolean __setattr_null(java.lang.String name, org.python.Object value) {
 
     @org.python.Method(
             __doc__ = "Return the difference of two or more sets as a new set.\n\n(i.e. all elements that are in this set but not the others.)",
-            args = {"other"}
+            args = {"other"},
```
deep110: Hi, I don't think args is needed here, because according to docs this method only takes varags [(here)](https://docs.python.org/3.5/library/stdtypes.html#set.union) and single argument can be handled by tuple with one element.

tests/datatypes/test_set.py
```diff
@@ -162,37 +185,99 @@ def test_remove(self):
...
+                print(err)
             """)
 
     def test_update(self):
+        # successful
         self.assertCodeExecution("""
             x = {1, 2, 3}
             y = {3, 4, 5}
-            print(x.update(y))
-            print(x)
```
freakboy3742: I'm very wary of any test that modifies existing tests. "Single argument update" is a valid check here; you need to *augment* existing tests, not replace them.

python/common/org/python/types/Set.java
Saurabh--Kumar: @freakboy3742 , @deep110 I had used python 3.4 documentation. It seemed to suggest that it takes one args and rest arguments as varargs. Link to the documentation is below. Since the document with installing instructions said that voc uses python 3.4, I used python 3.4 documentation. I'll make the changes and make the code comply with 3.5. https://docs.python.org/3.4/library/stdtypes.html#frozenset

tests/datatypes/test_set.py
Saurabh--Kumar: @freakboy3742 Corrected both the issues.

Saurabh--Kumar:
@freakboy3742, @cflee, @eliasdorneles . Issues with this PR are resolved. I think this is ready for merge.

tests/datatypes/test_set.py
freakboy3742: No - you've made *a* change, but the original test has still been removed. Again - *any* change that *deletes* tests needs to make a *very* strong argument for that removal. The whole purpose of a test suite is to add tests that exercise functionality that you've explicitly implemented, so that any bugs won't re-occur. If you delete tests, you're removing that safety. This isn't just for this particular line - it's for *any* removals in the entire test suite.

cflee:
Thanks for your contribution! However, all the autogenerated Set test cases are currently passing in master, so I'm going to go ahead and close this PR.

PR #523, opened 2017-04-24
[WIP] some set methods

Saurabh--Kumar:
added __xor__() method. This is to test if my changes are successful (as for xor external tests need not be written.) If the build is successful, I'll add __ixxx__() methods also


cflee:
Thanks for your contribution! However, all the Set autogenerated test cases are currently passing in master, so I'm going to go ahead and close this PR.

PR #525, opened 2017-04-27
add filter builtin

ivanistheone:
Joint work with Nasreeen and Elias.

Does this look good?

### DIFF
--- a/tests/builtins/test_filter.py
+++ b/tests/builtins/test_filter.py
@@ -13,10 +13,10 @@ class FilterTests(TranspileTestCase):
         self.assertCodeExecution('print(list(filter(None, [])))')
 
     def test_lambda(self):
-        self.assertCodeExecution('print(list(filter(lambda x: x>1, [3,4,56,1, -11])))')
+        self.assertCodeExecution('print(list(filter(lambda x: x > 1, [3, 4, 56, 1, -11])))')
 
     def test_wrong_argument(self):
-        self.assertCodeExecution('print(list(  filter(None, None)  ))', exits_early=True)
+        self.assertCodeExecution('print(list(filter(None, None)))', exits_early=True)
 
 
 class BuiltinFilterFunctionTests(BuiltinTwoargFunctionTestCase, TranspileTestCase):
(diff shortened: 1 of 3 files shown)


tests/builtins/test_filter.py
```diff
@@ -1,53 +1,22 @@
...
-
-            print(filter(testish, ListLike()))
-            mylist = ListLike()
-            print(filter(testish, mylist).__next__())
-            print(filter(testish, mylist).__next__())
-            print(filter(testish, mylist).__next__())
-            try:
-                print(filter(testish, mylist).__next__())
-            except StopIteration:
-                pass
```
eliasdorneles: Note to other maintainers: this was dead code (and it was broken too), the tests were marked as expected failure and they were failing for the wrong reasons.
We decided to go with new tests that would focus more on the different argument combinations than in different types and predicates, and rely on the already implemented types.

tests/builtins/test_filter.py
```diff
@@ -1,53 +1,22 @@
...
-        self.assertCodeExecution(self.base_code % ("b'123'", "x"))
+    def test_none(self):
+        self.assertCodeExecution('print(list(filter(None, [True, False, True])))')
+        self.assertCodeExecution('print(list(filter(None, [])))')
 
-    @expectedFailure
-    def test_float(self):
-        self.assertCodeExecution(self.base_code % ("[3.14, 2.17, 1.0]", "x > 1"))
+    def test_lambda(self):
+        self.assertCodeExecution('print(list(filter(lambda x: x>1, [3,4,56,1, -11])))')
```
cflee: Could we have some whitespace around `>` operator, and after the commas in the array?

tests/builtins/test_filter.py
```diff
@@ -1,53 +1,22 @@
...
 
-    @expectedFailure
-    def test_int(self):
-        self.assertCodeExecution(self.base_code % ("[1, 2, 3]", "x * 2"))
+    def test_wrong_argument(self):
+        self.assertCodeExecution('print(list(  filter(None, None)  ))', exits_early=True)
```
cflee: I assume the extra spacing around the `filter()` are deliberate – is it supposed to highlight something?

PR #528, opened 2017-04-30
Implement isnumeric for Str and write test case

atmishra:
Refs #36 This is my first PR. I have tried implementing isnumeric method for Str and have written some test case.

### DIFF
--- a/python/common/org/python/types/Str.java
+++ b/python/common/org/python/types/Str.java
@@ -860,8 +860,17 @@ public class Str extends org.python.types.Object {
                     "False otherwise.\n"
     )
     public org.python.Object isnumeric() {
-        throw new org.python.exceptions.NotImplementedError("isnumeric() has not been implemented.");
+        if (this.value.isEmpty()) {
+            return new org.python.types.Bool(false);
+        }
+        for (char ch : this.value.toCharArray()) {
+            if (!(Character.isDigit(ch))) {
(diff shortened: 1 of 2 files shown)


python/common/org/python/types/Str.java
```diff
@@ -860,9 +860,18 @@ private long toPositiveIndex(long index) {
...
+        }
+        for (char ch : this.value.toCharArray()) {
+            if (!(Character.isDigit(ch))) {
+                return new org.python.types.Bool(false);
+            }
+        }
+        return new org.python.types.Bool(true);
     }
 
+
```
cflee: I believe there's just one newline between the other methods in this class – so I think it's not necessary to add this newline here.

tests/datatypes/test_str.py
```diff
@@ -681,6 +681,12 @@ def __str__(self):
                 print(err)
             """)
 
+    def test_isnumeric(self):
+        self.assertCodeExecution("""
+        for str_ in ['123', '123.4', 'abc', '', ' ', 'ABCD', 'ABCD ', '12323445']:
```
cflee: Could you also test some other cases like `'123.', '.12', '1A', 'A1'` to make sure the edge cases are handled correctly?

atmishra:
Did the requested changes.

cflee:
Looks great, thanks for the contribution! ✨

PR #529, opened 2017-05-01
Add bytes.capitalize and bytes.__iter__

katieamazing:
Here is an implementation of the bytes.capitalize and byte.__iter__() methods.

I ran into a problem while testing the capitalize function with other byte values:

    print(b'\\xfa'.capitalize()[0])

I saw an error related to writing to the test suite, and left it alone for now.

Does this look good?

### DIFF
--- a/python/common/org/python/types/Bytes.java
+++ b/python/common/org/python/types/Bytes.java
@@ -563,7 +563,7 @@ public class Bytes extends org.python.types.Object {
         byte[] value = new byte[this.value.length];
         for (int i = 0; i < this.value.length; i++) {
             byte b = this.value[i];
-            if (b < 128) { // TODO: double check if this is sufficient safeguard for ascii
+            if (b < 127 && b > 32) {
                 char c = (char) b;
                 if (i == 0) {
                     c = Character.toUpperCase(c);
--- a/python/common/org/python/types/Bytes.java
+++ b/python/common/org/python/types/Bytes.java
@@ -472,8 +472,12 @@ public class Bytes extends org.python.types.Object {
     @org.python.Method(
             __doc__ = ""
     )
-    public org.python.Iterable __iter__(java.util.List<org.python.Object> args, java.util.Map<java.lang.String, org.python.Object> kwargs, java.util.List<org.python.Object> default_args, java.util.Map<java.lang.String, org.python.Object> default_kwargs) {
-        throw new org.python.exceptions.NotImplementedError("bytes.__iter__ has not been implemented.");
+    public org.python.Iterable __iter__() {
+        java.util.List<org.python.Object> listOfBytes = new java.util.ArrayList<org.python.Object>();
+        for (byte b: this.value) {
+            listOfBytes.add(new org.python.types.Int(b));
+        }
+        return new org.python.types.List(listOfBytes).__iter__();
     }
 
     @org.python.Method(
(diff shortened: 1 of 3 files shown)


python/common/org/python/types/Bytes.java
```diff
@@ -555,8 +559,23 @@ public boolean __setattr_null(java.lang.String name, org.python.Object value) {
     @org.python.Method(
             __doc__ = ""
     )
-    public org.python.Object capitalize(java.util.List<org.python.Object> args, java.util.Map<java.lang.String, org.python.Object> kwargs, java.util.List<org.python.Object> default_args, java.util.Map<java.lang.String, org.python.Object> default_kwargs) {
-        throw new org.python.exceptions.NotImplementedError("bytes.capitalize has not been implemented.");
+    public org.python.Object capitalize() {
+        byte[] value = new byte[this.value.length];
+        for (int i = 0; i < this.value.length; i++) {
+            byte b = this.value[i];
+            if (b < 128) { // TODO: double check if this is sufficient safeguard for ascii
```
freakboy3742: This check should probably be < 127 - ASCII 127 is DEL, which generally isn't printable. You should also have a guard at the "low" end - `>= 32`.

tests/datatypes/test_bytes.py
```diff
@@ -20,6 +20,20 @@ def test_getattr(self):
                 print(err)
             """)
 
+    def test_capitalize(self):
+        self.assertCodeExecution("""
+            print(b'hello, world'.capitalize())
+            print(b'helloWORLD'.capitalize())
+            print(b'HELLO WORLD'.capitalize())
+            print(b'2015638687'.capitalize())
```
freakboy3742: If you want to be really rigorous here, you should add a test for non-ascii bytes. You can use a `\x` sequence to deliberately specify a "high" byte and a "low" byte (might I suggest `\xc8` as a high byte, so you can subtly sign your name into the test suite :smile:)

eliasdorneles: I was watching over @katieamazing's shoulder while she was doing this and we already tried that, but we bumped into https://github.com/pybee/voc/issues/530 .

So perhaps it can be added in a separated test method and marked with `@expectedfailure` for now until issue #530 is fixed.

freakboy3742: Sounds like a good approach to me. Might also be worth flagging that the expected failure test case can be merged into the general case test once the failure is resolved.

freakboy3742: Looking at #530 - I think the fix is trivial, so it may be worth rolling it into this PR.

tests/datatypes/test_bytes.py
```diff
@@ -28,6 +30,12 @@ def test_capitalize(self):
             print(b'2015638687'.capitalize())
         """)
 
+    @unittest.skip("Move this to test_capitalize upon resolution of #530")
```
freakboy3742: This should be expectedFailure, rather than skip. skip is for when a test may/may not run due to some other condition (e.g., whether a setting has been turned on) - in this case, the "setting" is a string, which is always true, so the test won't ever be executed.

eliasdorneles: > skip is for when a test may/may not run due to some other condition (e.g., whether a setting has been turned on)

Isn't that the case only for [`skypIf`](https://docs.python.org/3/library/unittest.html#skipping-tests-and-expected-failures) ?

In my understanding, the [argument for skip() is the reason why the test should be skipped](https://docs.python.org/3/library/unittest.html#skipping-tests-and-expected-failures).

I think skip is more appropriate here because it's not a true failure of the test (as in, assertion failing), because the code never gets to run (and therefore we never get to the point of executing the assertion itself).

What do you think?

freakboy3742: Sorry - you're right - I got `skip` and `skipIf` confused. The point is the same though - skip will *never* run the test; expectedFailure will run it, but accept it when it fails. When, at some point in the future, the bug is fixed, the test will start passing - and the expectedFailure will cause an "unexpected success" - which means someone will be forced to look at that line of code (and the comment), and correct the problem, merging the test cases.

eliasdorneles: Ohh, got it, that makes sense!
Thanks for clearing that out, sorry for the nitpicking. :)

@katieamazing my bad, I agree w/ @freakboy3742, better to go with expectedFailure here. :+1:

katieamazing: Don't be sorry, it was educational for me to hear this discussion on skip/skipIf v. expectedFailure! I've pushed the change. Hopefully that works. :)

freakboy3742:
Hooray! 🎉 Congratulations on your first FOSS commit - may it be the first of many!

PR #539, opened 2017-05-22
Remove output that does not appear.

bjcoleman:


glasnt:
Come see @freakboy3742 for your shiny shiny, if you haven't already :sparkles:

PR #543, opened 2017-05-22
Implements int greater than equal to class.

harrisd19:


cflee:
Thanks for your contribution! However, this test case is currently passing in master, so I'm going to go ahead and close this PR.

PR #547, opened 2017-05-22
Implements bool > class

stalh05:


cflee:
Thanks for the contribution! However, it seems like this is already implemented and passing tests in master, closing this PR.

PR #553, opened 2017-05-22
Make NotImplementedToExpectedFailure work

astronouth7303:
It was trying to get the wrappee from the object that the wrapper is kept on, leading to a recursion error.


Hwesta:
This turned a lot of tests into unexpected successes, and updating that is in progress.  PR will be updated.

astronouth7303:
Appears to be a duplicate of #541

astronouth7303:
Yeah, @PamelaM is a bajilion miles ahead of me.

PR #556, opened 2017-05-23
Fix dict conversion errors for strings

indexOutOfBound5:
When sets and frozen sets were converted to dicts, if they contained
string elements those string elements would not be converted as they
would in python. Fix this.

As part of this commit, the mixed case unit test was removed, this is
due to randomisation present in sets in python 3.4 causing flakey tests.

Signed-off-by: Meghan Halton <meghalton@gmail.com>
Signed-off-by: Pamela McA'Nulty <pamela@mcanulty.org>

### DIFF
--- a/python/common/org/python/types/Dict.java
+++ b/python/common/org/python/types/Dict.java
@@ -66,6 +66,12 @@ public class Dict extends org.python.types.Object {
                             data = ((org.python.types.Tuple) next).value;
                         } else if (next instanceof org.python.types.List) {
                             data = ((org.python.types.List) next).value;
+                        } else if (next instanceof org.python.types.Str) {
+                            org.python.types.Str str = ((org.python.types.Str) next);
+                            data = new java.util.ArrayList<org.python.Object>();
+                            for (int i = 0; i < ((org.python.types.Int) str.__len__()).value; i++) {
+                                data.add(str.__getitem__(new org.python.types.Int(i)));
+                            }
                         } else {
                             throw new org.python.exceptions.TypeError(
                                     "cannot convert dictionary update sequence element #" + generated.size() +
--- a/tests/builtins/test_dict.py
+++ b/tests/builtins/test_dict.py
@@ -12,8 +12,6 @@ class BuiltinDictFunctionTests(BuiltinFunctionTestCase, TranspileTestCase):
         'test_bytearray',
         'test_bytes',
         'test_class',
-        'test_frozenset',
         'test_list',
-        'test_set',
         'test_str',
     ]
--- a/tests/builtins/test_list.py
+++ b/tests/builtins/test_list.py
@@ -1,4 +1,4 @@
-from .. utils import TranspileTestCase, BuiltinFunctionTestCase
+from .. utils import TranspileTestCase, BuiltinFunctionTestCase, SAMPLE_SUBSTITUTIONS
 
 
 class ListTests(TranspileTestCase):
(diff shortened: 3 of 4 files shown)


PamelaM:
@indexOutOfBound5 - Just eyeballed your latest travis logs and your latest changes.

It looks like all the errors are "just" updates to `SAMPLE_SUBSTITUTIONS`.

I've got a little bit of time here, so I'm going to take a stab at updating it then posting the sample here - so you can copy-paste instead of dance with merging and timing.

ps.  I suspect that hashing is different between Python and Java, either algorithmically or simply in using `PYTHONHASHSEED`

PamelaM:
New suggestion:
use this (untested) code:

```
from itertools import permutations
def _all_permutaions(key, value):
	# If key is a text representation of a 'set', use itertools.permutations 
	to generate all possible combinations for the key 
	
	if key[0] != "{" or key[-1] != "}":
		return value
    return ["{%s}" % str(p)[1:-1]  for p in permutations(eval(key))]

SAMPLE_SUBSTITUTIONS = {k: _all_permutations(k, v) for k, v in SAMPLE_SUBSTITUTIONS.items()}
```

Should make things _much_ easier, and all of the values for the set-string keys can default to `None`, so `SAMPLE_SUBSTITUTIONS` can become much more compact, readable code-wise

indexOutOfBound5:
@PamelaM That's great! I went to sleep planning to write that exact function last night, I'll integrate today. Thank you!

I did try out using PYTHONHASHSEED but it's not behaving as expected :(

PamelaM:
You should wait a bit, I'm expanding it a bit to handle input keys of set, list and dict

I'll post what I have before I leave my hotel room.

PamelaM:
Here you go.  No doc string and the eyeball validation at the end isn't needed and I haven't used this inside `voc` yet, but I've got to get going and it runs as a script with python 3.4, 3.5 and 3.6

I hope to check in again around 3 o 4 this afternoon, during my layover.

ps - it was a _lot_ of fun pairing with someone who could keep up with me yesterday.  Great job!

```
from itertools import permutations
def _all_permutations(key, value):
	'''
	
	'''
	if value is not None:
		return value

	try:
		evaluated = eval(key)		
	except Exception as e:
		# Key cannot be evaluated
		print(e)
		return value
		
	if isinstance(evaluated, dict):
		typ, iterable = dict, evaluated.items()
	elif isinstance(evaluated, tuple):
		typ, iterable = tuple, evaluated
	elif isinstance(evaluated, list):
		typ, iterable = list, evaluated
	elif isinstance(evaluated, (set, frozenset)):
		typ, iterable = set, evaluated
	else:
		# Could return value here, but it's really unexpected
		raise TypeError("Cannot handle object type of %s" % type(evaluated))
	
	return [str(typ(p)) for p in permutations(iterable)]
		
		
SAMPLE_SUBSTITUTIONS = {
    # Normalize set ordering
    "{1, 2.3456, 7}": None,
    "{1, 2.3456, 'another'}": None,
    "{'an', 'to', 'on'}": None,
    "{'one', 'two', 'six'}": None,
    "{'a', 'b', 'c'}": None,
    "{'to', 'two', 'one', 'an', 'on', 'six'}": None,
    "{1, 'two', 2.3456, 7, 'one', 'six'}" : None,
    "{'to', 1, 2.3456, 7, 'an', 'on'}" : None,
    "['to', 'an', 'on']": None,
    # Normalize dictionary ordering
    "{'a': 1, 'c': 2.3456, 'd': 7}": None,
    "{'a': 1, 'c': 2.3456, 'd': 'another'}": None,
    "{'a': 'n', 't': 'o', 'o': 'n'}": None,
    # Normalize precision error
    "-3.14159": ["-3.1415900000000008"],
}

SAMPLE_SUBSTITUTIONS = {k: _all_permutations(k, v) for k, v in SAMPLE_SUBSTITUTIONS.items()}

for k, v in sorted(SAMPLE_SUBSTITUTIONS.items()):
	print("%4s  %-40r   %r" % (len(v), k, v[0]))
	
```

PR #557, opened 2017-05-23
Multiple comparisons (Fixes #540)

### LINKED ISSUE #540: Support for chained comparisons
[Chained comparisons](https://docs.python.org/3/reference/expressions.html#comparisons) are used in many standard library modules, including the [urllib.request.parse](https://github.com/pybee/ouroboros/blob/master/ouroboros/urllib/parse.py#L158) which is good to prioritize.

The following test code:

```
print(1 == 1 == 1)
print(1 == 1 == 2)
print(1 == 1 != 2)
print(1 == 1 < 2)
print(3 == 3 < 2)
print(1 < 2 < 3 > 2)
```

When running with CPython, it prints:

```
$ python multiplecomparators.py
True
False
True
True
False
True
```

With VOC, the compilation fails because [the syntax isn't supported yet](https://github.com/pybee/voc/blob/master/voc/python/ast.py#L1743):

    NotImplementedError: Don't know how to resolve multiple comparators

Fixing this will need updating the `visit_Compare` method in `voc/python/ast.py` to generate the proper bytecode for chained comparisons.

elainewong:
@katlings, @eliasdorneles and I figured out multiple comparisons. We believe this is ready for review.

### DIFF
--- a/tests/structures/test_comparisons.py
+++ b/tests/structures/test_comparisons.py
@@ -564,3 +564,15 @@ class ComparisonTests(TranspileTestCase):
             print(y == Anything())
             print(Anything() == y)
             """)
+
+    def test_multiple_comparisons(self):
+        self.assertCodeExecution("""
+            print(1 == 1 == 1 == 1)
+            print(1 > 2 > 3)
+            print(1 == 1 == 2)
+            print(1 == 1 != 2)
+            print(1 == 1 < 2)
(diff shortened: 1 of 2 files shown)


eliasdorneles:
This looks great, thanks @elainewong and @katlings ! :tada: :tada: 
Merging as soon as the Travis build is green!

PR #560, opened 2017-05-23
Added Bytes.index in Java

umrashrf:


python/common/org/python/types/Bytes.java
```diff
@@ -665,10 +665,16 @@ public boolean __setattr_null(java.lang.String name, org.python.Object value) {
...
+            __doc__ = "get an index of any byte from bytes",
+            args = {"sub"},
+            default_args = {"start", "end"}
     )
-    public org.python.Object index(java.util.List<org.python.Object> args, java.util.Map<java.lang.String, org.python.Object> kwargs, java.util.List<org.python.Object> default_args, java.util.Map<java.lang.String, org.python.Object> default_kwargs) {
-        throw new org.python.exceptions.NotImplementedError("bytes.index has not been implemented.");
+    public org.python.types.Int index(org.python.Object sub, org.python.Object start, org.python.Object end) {
+        Int pos = this.find(sub, start, end);
+        if (pos.equals(new Int(-1))) {
+            throw new org.python.exceptions.ValueError("subsection not found");
```
eliasdorneles: I looked into the build failure, and it appears that this message is different in CPython: for versions before 3.6 it says "substring not found", but from 3.6 onwards it says "subsection not found".

Could you add an `if` clause here to make it mirror the CPython behavior?
Here is [an example of how to do that check](https://github.com/pybee/voc/blob/master/python/common/org/python/types/Str.java#L331).

Thank you!

umrashrf:
thank you @eliasdorneles

PR #566, opened 2017-05-24
Added str.splitlines

seocam:


tests/datatypes/test_str.py
```diff
@@ -695,6 +695,20 @@ def test_isidentifier(self):
...
+        str_ = "\\r\\n"
+        print(repr(str_))
+        """)
+
+    def test_splitlines(self):
+        self.assertCodeExecution("""
+        str_ = "aaa\\nbbb\\rccc\\r\\nddd\\n\\reee"
+
+        print(str_.splitlines())
+        print(str_.splitlines(True))
```
eliasdorneles: Hey @seocam, thanks for the PR!

One little thing: the empty string case isn't handled correctly, it's returning a list containing an empty string like this:

```
>>> ''.splitlines()
['']
```

It should return an empty list, like this:

```
$ python
[GCC 4.8.4] on linux
>>> ''.splitlines()
[]
```

Can you add the special handling and add this to the tests too?
Thank you!

eliasdorneles:
Thanks @seocam and @marianamioto !

PR #567, opened 2017-05-24
Implementing almost all of the is_____ functions on bytes

mistermocha:


tests/datatypes/test_bytes.py
```diff
@@ -61,6 +61,42 @@ def test_find(self):
...
 
+    def test_isdigit(self):
+        self.assertCodeExecution("""
+            for byt in [b'123', b'678', b'99999', b'abc', b'2s3x', 'a2b3', b'2S3X', 'A2b3', b'6', b'k', 'GHI', 'Jkl']:
+                print(byt.isdigit())
+        """)
+
+    def test_islower(self):
+        self.assertCodeExecution("""
+            for byt in [b'123', b'678', b'99999', b'abc', b'2s3x', 'a2b3', b'2S3X', 'A2b3', b'6', b'k', 'GHI', 'Jkl']:
```
eliasdorneles: Hey @mistermocha , thanks for the PR!
This looks good, just one small thing here: could you add a test with punctuation as well?

This works in CPython:

```
>>> b'a..'.islower()
True
```

The current implementation is returning `False`.

`islower` should return True if there is at least one character that is letter and lower case, and all of the others are not upper case.

The same applies to `isupper`, with the cases swapped.

tests/datatypes/test_bytes.py
```diff
@@ -63,31 +63,36 @@ def test_find(self):
 
     def test_isdigit(self):
         self.assertCodeExecution("""
-            for byt in [b'123', b'678', b'99999', b'abc', b'2s3x', 'a2b3', b'2S3X', 'A2b3', b'6', b'k', 'GHI', 'Jkl']:
+            for byt in [b'123', b'678', b'99999', b'abc', b'2s3x', 'a2b3', b'2S3X', 'A2b3', b'6',
+                    b'k', b'GHI', b'Jkl', b'!234', b'.', b'?/']:
```
eliasdorneles: Sorry, I wasn't clear.
The test case that triggers the problem is punctuation + lowercase characters, like `b'.a.'` (and `b'.A.'` for isupper)

```
$ python
[GCC 4.8.4] on linux
>>> b'.a.'.islower()
True
>>> b'.A.'.isupper()
True
```

cflee:
Thanks for your contribution! However, as all these functions have since been implemented and merged into master, closing this PR.

PR #573, opened 2017-05-25
  Define unary operation on Complex data type

jack8192:
complex.__bool__ now tests the real part of a complex number for being > zero and then the imaginary component.  If either is NOT equal to zero then the complex number is "truthy".


python/common/org/python/types/Complex.java
```diff
@@ -19,7 +19,8 @@
     }
 
     public org.python.Object byValue() {
-        throw new org.python.exceptions.NotImplementedError("complex.byValue() has not been implemented.");
+        /*throw new org.python.exceptions.NotImplementedError("complex.byValue() has not been implemented."); */
```
phildini: Remove the comment, please. We shouldn't need it anymore.

jack8192:
have made your suggested changes and resubmitted.

PR #574, opened 2017-05-25
Negative Slice Index Support to Bytes.__getitem__

bsthorpe742:
This pull request address Bytes.__getitem__ to support negative indexes when slicing.  Many other objects which support slicing do not support negative indexing at this time.  This could be used as an example to fix other objects which support slicing.  Test cases cover many edge cases with positive and negative step sizes.


tests/datatypes/test_slice.py
```diff
@@ -75,6 +75,65 @@ def __index__(self):
...
+            print("x[-5:] = ", x[-5:])
+            print("x[:-5] = ", x[:-5])
+            print("x[-2:-8] = ", x[-2:-8])
+
+            print("x[::-2] = ", x[::-2])
+            print("x[-5::-2] = ", x[-5::-2])
+            print("x[:-5:-2] = ", x[:-5:-2])
+            print("x[-2:-8:-2] = ", x[-2:-8:-2])
+            """)
+
```
simsausaurabh: Please explain why the testing of negative indexing is done in separate method. I am unable to understand.

freakboy3742: I'm not sure what you're referring to. `test_slice_bytes` is a single test method that contains several examples that are tested.

PR #577, opened 2017-05-25
test_complex.py is pruned to list known not_implemented cases

jack8192:
The not_implemented lists of methods have been pruned to only include known fails.


cflee:
The changes in this PR seem to be in your other PR #573 (since they seem to be originating from the same branch?) which was merged, so closing this PR.

PR #578, opened 2017-05-25
Better str.splitlines (and more)

seocam:
* Fixed __repr__ (somecases missing)
* Added more line breaking chars for splitlines method
* Added a new method to help check if a char is printable or not

Signed-off-by: Sergio Oliveira <seocam@seocam.com>

### DIFF
--- a/python/common/org/python/types/Str.java
+++ b/python/common/org/python/types/Str.java
@@ -66,13 +66,75 @@ public class Str extends org.python.types.Object {
     //     throw new org.python.exceptions.NotImplementedError("__init__() has not been implemented.");
     // }
 
+    private static boolean isCharPrintable(char c) {
+        // ASCII non-printable
+        if ((int) c <= 0x1f || (int) c >= 0x7f && (int) c <= 0xa0 || (int) c == 0xad) {
+            return false;
+        }
+        if ((int) c == 0x2029) {
+            return false;
+        }
+        if (Character.isISOControl(c)) {
+            return false;
+        }
+        return true;
+    }
+
     @org.python.Method(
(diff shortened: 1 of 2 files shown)


python/common/org/python/types/Str.java
```diff
@@ -66,14 +66,76 @@ public Str(org.python.Object[] args, java.util.Map<java.lang.String, org.python.
...
     )
     public org.python.Object __repr__() {
-        String repr = this.value.replaceAll("\\n", "\\\\n").replaceAll("\\r", "\\\\r");
-        return new org.python.types.Str("'" + repr + "'");
+    /*
+    * Reference: https://www.python.org/dev/peps/pep-3138/#id7
+    * TODO: Need to treat the leading surrogate pair characters
+    */
+        StringBuilder sb = new StringBuilder();
+        boolean has_double_quote = false;
```
cflee: This is a really minor style point, but Java usually has camelCase variable names instead of snake_case.

python/common/org/python/types/Str.java
```diff
@@ -66,14 +66,76 @@ public Str(org.python.Object[] args, java.util.Map<java.lang.String, org.python.
...
+            }
+        }
+
+        // Decide if we wanna wrap the result with single or double quotes
+        String quote;
+        String repr = sb.toString();
+
+        if (has_single_quote) {
+            if (has_double_quote) {
+                quote = new String("'");
```
cflee: We can just have the string literal `"'"` here, instead of the `new String()` constructor – similarly for next two instances of this.

cflee:
Thanks @eliasdorneles !

PR #579, opened 2017-05-26
Add support for test_list::InplaceListOperationTests::test_add_str

chernrick:
My first pull request in PyBee - just getting used to the process.

### DIFF
--- a/tests/datatypes/test_list.py
+++ b/tests/datatypes/test_list.py
@@ -685,7 +685,6 @@ class InplaceListOperationTests(InplaceOperationTestCase, TranspileTestCase):
         'test_add_dict',
         'test_add_frozenset',
         'test_add_range',
-        'test_add_set',
 
         'test_and_class',
         'test_and_frozenset',
--- a/tests/datatypes/test_list.py
+++ b/tests/datatypes/test_list.py
@@ -686,7 +686,6 @@ class InplaceListOperationTests(InplaceOperationTestCase, TranspileTestCase):
         'test_add_frozenset',
         'test_add_range',
         'test_add_set',
-        'test_add_str',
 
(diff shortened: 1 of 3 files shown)


tests/datatypes/test_list.py
```diff
@@ -686,7 +686,6 @@ class InplaceListOperationTests(InplaceOperationTestCase, TranspileTestCase):
         'test_add_frozenset',
         'test_add_range',
         'test_add_set',
```
chernrick: I did not remove this test because it incorrectly relies on the Python and Java output for sets to be in the same order

freakboy3742: We have a way to compensate for this - if you look at [test_complex.py](https://github.com/pybee/voc/blob/master/tests/datatypes/test_complex.py#L50), you'll see a "substitutions" dictionary declared in the test. This dictionary defines all the possible ways that a result might be rendered, and provides alternatives. Any test output will be normalised using these substitutions - any string found in the list of values will be replaced with the key value from the dictionary. So, pick one possible output ordering from the set.  enumerate all other possible outputs as alternatives, and the test should pass.

chernrick:
@freakboy3742 Thanks for the review! Since the output of list_add_set is expected to be the original list with an unsorted set appended (for all SAMPLE_DATA), there are a lot of permutations. So, I have updated the SAMPLE_SUBSTITUTIONS in utils.py to handle these cases as well.

cflee:
Thanks for your contribution! However, the handling for Set and Str was added in #643 (see https://github.com/pybee/voc/pull/643/files#diff-d2a25b564b697db2462757d0b7fa5dfbR111), so closing this PR.

PR #581, opened 2017-05-26
Initial hypothesis tests for proof-of-concept

mistermocha:
I got the hypothesis test working, but only when I put in some boundaries around generated letters. Doing the first 500 codepoints of Unicode generates an array of words and limits the Unicode characters so we don't hit unsupported characters.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,7 +11,7 @@ env:
   - CIRCLE_NODE_TOTAL=5 CIRCLE_NODE_INDEX=4
 install:
   - "pip install ."
-  - "pip install pytest"
+  - "pip install pytest hypothesis"
   - "pip install git+https://github.com/pybee/pytest-circleci"
   - "pip install flake8"
 script:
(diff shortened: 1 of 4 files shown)


eliasdorneles:
So I've been thinking about this and here is a more concrete proposal for what to do about it:

* keep using manually written tests for the main cases
* add hypothesis tests, but skip them from the normal builds
* run the hypothesis tests weekly on [scheduled builds](https://docs.travis-ci.com/user/cron-jobs/)

This way we don't block a PR that already makes things better but that don't fully support things like Unicode related edge cases, and we still keep track of the problems and cases that we don't support yet through the hypothesis results.

If we do this, than this PR should be split in two: one with the Java code modifications and manually written tests to exercise them, and another that adds just the hypothesis style tests to be skipped unless [the environment variable `TRAVIS_EVENT_TYPE` is set to `cron`](https://docs.travis-ci.com/user/cron-jobs/#Detecting-Builds-Triggered-by-Cron).

What do you think?

phildini:

PR #583, opened 2017-05-26
[WIP] Fix binary arithmetic operation methods

cflee:
Currently, the arithmetic operations `+, -, *, /, //, %, **, <<, >>, &, ^, |` are implemented in the dunder `__add__()` methods, but are implemented and tested using the operators `+`. However, similar to the comparison methods (#122), the dunder methods should often be returning `NotImplemented` instead of throwing `TypeError`, but yet the operators like `+` need to support the reflected operation logic [specified in the Python data model](https://docs.python.org/3.5/reference/datamodel.html#emulating-numeric-types).

**TODO:**
- [ ] Fix dunder methods for binary arithmetic operations
- [ ] Fix dunder methods for reflected binary arithmetic operations
- [ ] (?) Fix dunder methods for in-place binary arithmetic operations
- [ ] Fix operator logic to generate appropriate errors and error messages

- [ ] Check when `__binop__()` should bail when faced with an exception


cflee:
This is more problematic than I thought. It's largely possible to get the direct `__add__()`, `__radd__()` methods correct by simplifying them and usually changing them to return `NotImplemented`, but generating the appropriate `TypeError` messages is a problem. Different datatypes all have edge cases for different operators, like how Bytes has a different message for `+` as it's internally a concat instead of addition operation.

The current implementation requires implementing all these datatype-specific edge cases in `org.python.types.Object`, which doesn't seem right.

cflee:
Closing – the ongoing work to use interfaces to streamline the datatype implementations should be a better fix!

Issue #586, opened 2017-06-04
Missing support for inner classes


trbedwards:
I cannot compile a python script with inner classes. For example, in my `voc_test.py` file:

```
class Thing(object):
    class InnerThing(object):
        pass
```

I invoke `voc voc_test.py` and get this error:

```
Traceback (most recent call last):
  File "/Users/tre11/anaconda2/envs/python3.5/lib/python3.5/site-packages/voc/python/ast.py", line 160, in visit
    super().visit(node)
  File "/Users/tre11/anaconda2/envs/python3.5/lib/python3.5/ast.py", line 245, in visit
    return visitor(node)
  File "/Users/tre11/anaconda2/envs/python3.5/lib/python3.5/site-packages/voc/python/ast.py", line 49, in dec
    fn(self, node, *args, **kwargs)
  File "/Users/tre11/anaconda2/envs/python3.5/lib/python3.5/site-packages/voc/python/ast.py", line 284, in visit_ClassDef
    klass = self.context.add_class(class_name, extends, implements)
AttributeError: 'Class' object has no attribute 'add_class'

Problem occurred in voc_test.py
Node: ClassDef(name='InnerThing', bases=[
    Name(id='object', ctx=Load(), lineno=2, col_offset=21),
  ], keywords=[], body=[
    Pass(lineno=3, col_offset=8),
  ], decorator_list=[], lineno=2, col_offset=4)
```

I am using python 3.5 and the latest master version of voc

trbedwards:
For context, I'm trying to compile [`dateutil.parser`](https://dateutil.readthedocs.io/en/stable/parser.html)

eliasdorneles:
Thanks for reporting!
The solution to this is to add an `add_class` method to [`Class` class in the klass.py file](https://github.com/pybee/voc/blob/e7479dc3dbd0051a69ed3876fbdc9d20683bd553/voc/python/klass.py#L19) similar [to the one in module.py](https://github.com/pybee/voc/blob/e7479dc3dbd0051a69ed3876fbdc9d20683bd553/voc/python/modules.py#L177-L232).
Which is also quite similar to the one in [methods.py](https://github.com/pybee/voc/blob/master/voc/python/methods.py#L372-L427), it might be interesting to refactor the common parts.

PR #591, opened 2017-06-20
Allow class definition inside a class

uranusjr:
This implements the inner class functionality following the solution @eliasdorneles mentions in #586.

The `add_class` methods in `Class`, `Method`, and `Module` are actually completely compatible if all `self` reference is pointed to `self.module` (which points to `self` in `Module`) instead. It seems weird to me to move it to `Block` though… maybe move it to a mixin instead? (What should it be called?)

### DIFF
--- a/voc/python/klass.py
+++ b/voc/python/klass.py
@@ -1,9 +1,15 @@
 import os
 
 from ..java import (
-    Annotation, Class as JavaClass, Code as JavaCode, ConstantElementValue,
-    Field as JavaField, Method as JavaMethod, RuntimeVisibleAnnotations,
-    SourceFile, opcodes as JavaOpcodes,
+    Annotation,
+    Class as JavaClass,
+    Classref as JavaClassref,
+    Code as JavaCode,
+    ConstantElementValue,
+    Field as JavaField,
+    Method as JavaMethod, RuntimeVisibleAnnotations,
+    SourceFile,
+    opcodes as JavaOpcodes,
 )
 from .blocks import Block, IgnoreBlock
 from .methods import (
@@ -20,14 +26,14 @@ class Class(Block):
     CONSTRUCTOR = InitMethod
 
     def __init__(
-            self, module, name,
+            self, parent, name,
             namespace=None, extends=None, implements=None,
             public=True, final=False, methods=None, fields=None, init=None,
             verbosity=0, include_default_constructor=True):
-        super().__init__(parent=module, verbosity=verbosity)
+        super().__init__(parent=parent, verbosity=verbosity)
         self.name = name
         if namespace is None:
-            self.namespace = '%s.%s' % (module.namespace, module.name)
+            self.namespace = '.'.join([self.module.namespace, self.module.name])
         else:
             self.namespace = namespace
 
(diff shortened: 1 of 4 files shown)


voc/python/klass.py
```diff
@@ -51,6 +58,10 @@ def __init__(
         self._constructor = None
 
     @property
+    def full_name(self):
+        return '.'.join(self.namespace.split('.')[1:] + [self.name])
```
freakboy3742: What's the generated class name here? What happens if I compile the following code:
```
class MyClass1:
    class InnerClass
        ...

class MyClass2:
    class InnerClass
        ...
```

Do the two `InnerClass` instances collide?

uranusjr: Good point—they do. Python uses `<module>.MyClass.InnerClass` in the `repr` result, so I guess we can follow this convention.

I wonder if this should be handled here, or in the class namespace. In the above example, should the inner classes’ namespaces be `<module>`, or `<module>.MyClass1` and `<module>.MyClass2`? And if it is the latter, how do we infer that inside the inner class (which does not have any information about the outer class, only the module it is in).

The answer is probably that we need to specify the namespace when creating the inner class in `add_class`:

```python
def add_class(self, ...):
    klass = Class(
        self.module,
        name=class_name,
        namespace=self.class_name,   # This.
        extends=extends,
        implements=implements,
    )
    ...
```

This would also affect both `descriptor` and `class_name` of the inner class. I’m not familiar enough with the code base to know if this would have unintended consequences.

uranusjr:
Added a test to illustrate the name conflict problem, and tried to solve it with the approach mentioned above.

voc/python/klass.py
freakboy3742: The complication here will be what Java can actually support. Java has major problems when you have a class and a namespace with the same name. If there are capitalisation changes, you're generally OK, but having a `MyClass1.class` and `MyClass1/InnerClass.class` in the same project will be a problem.

Based on that, I suspect the best option will be to use the `<module>` namespace, but use a classname of `MyClass1$InnerClass` (which is Java's general naming approach for "anonymous" or "inner" pieces).

uranusjr: Ah, yes, I should have thought of this! Updated.

uranusjr: This is more involved than I expected… Some thoughts while I wait for the tests to finish. (Also see: commit message of a75bc9c.)

Say we have this structure:

```python
class MyClass1:
    class InnerClass
        ...
```

The inner class would be called `__main__.MyClass1.InnerClass` (Voc currently names it `MyClass1.InnerClass`, which is wrong, but that problem is for another day), but in JVM its descriptor needs to be `python/MyClass$InnerClass`. Previously a class’s descriptor is calculated by replacing `.`s with `/`s, but with inner classes in the picture we need to know when to use `$` or `/`.

I use a different logic behind `class_name`, `full_name`, and `descriptor`. For the former two, a class can get its own value by getting its parent’s value, and simply append `.{self.name}`; for the descriptor, however, only its parent knows whether it needs `$` or `/` before its name. I delegated the logic to `build_child_class_descriptor()` so that the parent can lend a hand on this.

Everything after is pretty straightforward. An inner class’s parent should be its outer class, and by implementing the same properties duck-typing takes care of the rest. The transpiled class name also needed changing.

uranusjr:
I hope I nailed it this time!

uranusjr:
Don’t like this version either… does not fit with the current closure implementation. I need to this this through.

niconorsk:
Was thinking on trying to push that PR across the line but I'm not sure what the main missing components are.  A far as I can tell, the CI tests are failing because of https://github.com/pybee/voc/issues/595 Fixing that seems easy enough but should that just be a completely different PR?

uranusjr:
@niconorsk Honestly I can’t remember exactly, but I think I achieved what I wanted in c4cfb43, and was just waiting for #595. Since that is fixed in #607 I think I can rebase after it is merged, or you can do it (and open a new PR) yourself as well, I have no problem with either.

uranusjr:
As a side note, one thing I remember myself wondered is whether we can refactor and combine `add_class` in `Class` and `Module`. In Python (the language) adding a class in a module and class behave exactly the same, and the voc implementation should probably reflect that. But that is not trivial work and probably worth another PR.

uranusjr:
Rebased (and squashed) to include #607!

PR #592, opened 2017-06-21
[WIP] Built-in ascii()

uranusjr:
The implementation is not complete, but maybe a pull request help make myself getting it done.

I added some extra tests to test input string edge cases. They likely don’t cover all our bases, but seems good enough for now.

Tests now marked as `not_implemented` but really shouldn’t:

* [x] ~~FIXED `test_slice`. `slice.__repr__` is not implemented.~~
    * [ ] `builtins.test_slice`. `test_bool` is returning `slice(None, 1, None)` but should be `slice(None, True, None)`.
* [ ] `test_class`. VOC reports the class’s repr as `<class 'MyClass'>` but CPython says `<class '__main__.MyClass'>`
* [ ] `test_complex`. This one I have absolutely no clue. Exception in complied program:
    ```
    Exception in thread "main" RuntimeError: org.python.types.NotImplementedType cannot be cast to org.python.types.Bool
        at python.complex_ascii.module$import(complex_ascii.py:1)
        at python.complex_ascii.main(complex_ascii.py)
    ```


uranusjr:
The `slice` implementation has more problems than I anticipated (never knew `slice('a')` is valid!). I’ll probably split that part into a separate PR, get it merged, and come back to this after that.

PR #593, opened 2017-06-23
Slice refactor

uranusjr:
Slice-specific stuff extracted from #592.

Python’s `slice()` literally accepts *anything* as its arguments, and does not check for types until it's applied to a sequence. This refactors the constructor and attributes to match this fact, and provide an extra method (and inner class) for its callee to perform type checking.

Also implements `__repr__` to match the standard behaviour.


uranusjr:
(rebased to master)

uranusjr:
Rebased to include #607.

PR #594, opened 2017-06-23
Fix complex.__repr__

uranusjr:
Complex-specific stuff pulled from #592.

`complex.__str__` == `complex.__repr__` in CPython, but we have two methods here. I moved the `__str__` implementation (which seems to work) to `__repr__`, and removed the original `__repr__` implementation (which does not).

Also fixed float handling in `Integer.__eq__` which is part of the reason why the old `__repr__` did not work.


phildini:

Issue #432, opened 2017-03-17
Fix keyword-only functions


eliasdorneles:
The following valid Python 3 code:

```
def func(*, a=1):
    print(a)

func(a=2)
func()
```

... gives the following output in CPython:

```
$ python functest.py 
2
1
```

... but raises NullPointerException if compiled and ran with VOC:

```
+ java -cp ../voc/dist/python-java-support.jar:. python.functest.__init__
2
Exception in thread "main" RuntimeError: java.lang.NullPointerException
	at org.python.types.Function.invoke(Function.java:412)
	at org.python.types.Function.invoke(Function.java:347)
	at python.functest.__init__.func(functest.py:2)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:498)
	at org.python.types.Function.invoke(Function.java:388)
	at org.python.types.Function.invoke(Function.java:347)
	at python.functest.__init__.module$import(functest.py:5)
	at python.functest.__init__.main(functest.py)
```

Ostrea:
I would like to work on it. Could you provide some hints on where to start?

eliasdorneles:
@Ostrea Hey! I don't really know much more than what's in the above report at the moment.
The problem might be on the bytecode generated (in ast.py file) or in the code for calling functions (since it happens at runtime), it needs some debugging.

Ostrea:
Thanks, will try to find problem

gautamworah96:
Hello, I tried debugging this problem , This runtime error is caused by  `java.lang.reflect.InvocationTargetException e` on line no : 399 in the Function.java file
 ..this exception in turn occurs when there is an error in the function which is wrapped by invoke.
.Thus there appears to be a bug in the `adjustArguments(instance, args, kwargs)` method..I think the bug is in this code :` if (n_args == 0) {
            return null;
        }`
line: 262
Also to check whether the control is passing on to this function or no can we use `System.out.println()`
or do we have to use another method..
Kindly guide me ahead..or am I going in the wrong direction :(

Ostrea:
[...]

PR #596, opened 2017-07-01
Fix keyword-only function definition

uranusjr:
This solves the two problems mentioned in #432.

First, the transpiler throws TypeError for a keyword-only argument without a default value because it tries to visit the `None` value as a node. Checking against this situation avoids the issue.

Second, the compiled JVM bytecode throws a NullPointerException when a function with keyword-only arguments is called using the default value because the bytecode emitter does not put the default value correctly for keyword-only arguments (it only did this for `ArgType.POSITIONAL_OR_KEYWORD`). Submit the default value for `ArgType.KEYWORD_ONLY` and this is fixed.

Two more tests are now expected to pass :D

### DIFF
--- a/tests/structures/test_function.py
+++ b/tests/structures/test_function.py
@@ -1,7 +1,5 @@
 from ..utils import TranspileTestCase
 
-from unittest import expectedFailure
-
 
 class LambdaTests(TranspileTestCase):
     def test_lambda(self):
@@ -20,8 +18,6 @@ class LambdaTests(TranspileTestCase):
 
             """)
 
-    @expectedFailure
-    # keyword only args not handled
     def test_lambda_kwonly(self):
         self.assertCodeExecution("""
             print((lambda *,x='something': x + '!!!')())
@@ -308,8 +304,6 @@ class FunctionTests(TranspileTestCase):
             print(i.a_method())
             """, exits_early=True)
 
-    @expectedFailure
-    # keyword only args not handled
     def test_function_kwonly(self):
         self.assertCodeExecution("""
             def myfunc(*,x=100):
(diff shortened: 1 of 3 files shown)


uranusjr:
This still leaves has one edge case unresolved:

~~~python
def func(*, a):
    print(a)

func()
~~~

CPython:

~~~
TypeError: f() missing 1 required keyword-only argument: 'a'
~~~

Voc:

~~~
Exception in thread "main" RuntimeError: java.lang.NullPointerException
~~~

This is similar to the second case above, but without a default value to fall back to, a NullPointerException occurs.

This is much more difficult to resolve because the Function class does not currently have a good way to record required keyword arguments. It uses a Map to store kwargs and their defaults, but providing a null value there means providing a `None` default, not *without* a default value. I can think of two possibilities:

a. Pass `NoneType.NONE` instead of `null` when you want a `None` default; pass `null` when there’s no default.
b. Introduce a new list to store required keyword arguments.

The second solution probably requires less changes, but the first is arguably semantically more correct? I am not sure about this. Also not sure how much impact this would cause; maybe it is better to merge this first, and work on this in a separate issue/PR.

eliasdorneles:
About the check, although `checkMissingArgs` doesn't have the full context, the `adjustArguments` function that calls it has (notably, it has the `kwonlyargcount` variable).
If we pass that to the `checkMissingArgs`, I think it's doable (I need to think on the exact conditions with more patience).

uranusjr:
@eliasdorneles I don’t think `kwonlyargcount` is enough. Consider the following two functions:

```python
def func1(*, a):
    print(a)

def func2(*, a=None):
    print(a)
```

They would have exactly identical `co_*` values (well, except `co_name` and `co_firstlineno`), but need to be adjusted differently when passed no arguments. I think I’ll try my first approach first…

PR #597, opened 2017-07-13
negative index list slices

Nasreen123:
Hi,
this adds negative index list slicing - does it look ok?

### DIFF
--- a/python/common/org/python/types/List.java
+++ b/python/common/org/python/types/List.java
@@ -338,7 +338,7 @@ public class List extends org.python.types.Object {
                         long start;
                         if (slice.start != null) {
                             if (slice.start.value < 0) {
-                                start = this.value.size() + slice.start.value;
+                                start = Math.max((this.value.size() + slice.start.value), -1);
                             } else {
                                 start = Math.min(slice.start.value, this.value.size());
                             }
@@ -349,14 +349,13 @@ public class List extends org.python.types.Object {
                         long stop;
                         if (slice.stop != null) {
                             if (slice.stop.value < 0) {
-                                stop = this.value.size() + slice.stop.value;
+                                stop = Math.max((this.value.size() + slice.stop.value), 0);
                             } else {
                                 stop = Math.min(slice.stop.value, this.value.size());
                             }
                         } else {
                             stop = -1;
                         }
-
                         for (long i = start; i > stop; i += step) {
                             sliced.add(this.value.get((int) i));
                         }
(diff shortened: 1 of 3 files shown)


python/common/org/python/types/List.java
```diff
@@ -337,14 +337,22 @@ public boolean __setattr_null(java.lang.String name, org.python.Object value) {
                     if (step < 0) {
                         long start;
                         if (slice.start != null) {
-                            start = Math.min(slice.start.value, this.value.size());
+                            if (slice.start.value < 0) {
+                                start = this.value.size() + slice.start.value;
```
uranusjr: Does this take account if `start` is less than 0? If not it should have a `Math.max` around it as well (similar to the positive case).

eliasdorneles: @uranusjr This line will only be executed for the case where `start` is less than 0. But you're right that it still needs the Math.max around it, for the case when start is lesser than `-1 -
 size of the list`. =)

@Nasreen123 thanks for contributing! :tada: 
Can you please add a test for the case `x[- 5 - len(x): -1]` ?

Nasreen123:
Hi @uranusjr and @eliasdorneles 
I added Math.max and extra tests - how does it look now?

python/common/org/python/types/List.java
```diff
@@ -337,36 +337,50 @@ public boolean __setattr_null(java.lang.String name, org.python.Object value) {
                     if (step < 0) {
                         long start;
                         if (slice.start != null) {
-                            start = Math.min(slice.start.value, this.value.size());
+                            if (slice.start.value < 0) {
+                                start = Math.max((this.value.size() + slice.start.value), -1);
+                            } else {
+                                start = Math.min(slice.start.value, this.value.size());
```
eliasdorneles: Hey @Nasreen123 , it looks great now!! :tada: 
However, while I was testing it, I found another bug that is already present in the previous code, I think this line should be:

    start = Math.min(slice.start.value, this.value.size() - 1);

The only difference is the `-1` at the end.
This will handle correctly the cases of: `x[len(x)::-2]` and `x[len(x)+2::-2]`, which are currently crashing with `IndexOutOfBounds`.

Can you sneak the tests and bugfix for that in this same PR? =)
Thank you!!

eliasdorneles:
Oh, the build failed because a test which wasn't passing before is now passing -- yay, that's good! =)
@Nasreen123 can you remove the line with `test_subscr_slice` in: https://github.com/pybee/voc/blob/master/tests/datatypes/test_list.py#L604 ?
The build should be green after that and the PR will be ready to merge!

PR #598, opened 2017-07-15
Add many tests, especially for dunder operator methods.

gkhaburzaniya:
Noticed that the dunder methods for many operators weren't being tested, so added tests for them. As well as tests for the. Also the divmod and round methods, which are very much like a binary operator.

### DIFF
--- a/docs/how-to/contribute.rst
+++ b/docs/how-to/contribute.rst
@@ -22,7 +22,7 @@ instead of using the official PyBee repository, you'll be using your own
 Github fork.
 
 As with the getting started guide, these instructions will assume that you
-have Python 3.4+, a Java 7 or Java 8 JDK, and Apache ANT installed.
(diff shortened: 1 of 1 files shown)


gkhaburzaniya:
Not sure what to do about the failing tests. Python 3.5 and Python 3.6 have differing behaviors for round(True, None). And the increased number of tests is probably making it go past the time limit for the other situations.

cflee:
Just in case it's useful, I have some datatype changes in PR #583 to get the math operators' direct dunder methods passing, but they make the non-direct (using operators) tests fail.

gkhaburzaniya:
@cflee Sounds like these two PRs will work well together then. I originally tried to do some of what you're doing in #583 but I couldn't quite get it working, so I decided to just write the tests to show that it isn't currently working.

PR #608, opened 2017-08-07
Use built-in virtual environment (not 3rd party)

treyhunner:
Python 3 comes with a `venv` package which allows you to create a virtual environment.

Asking users to use `virtualenv` presumes that they've installed the third-party `virtualenv` package. That package is necessary for using virtual environments in Python 2, but there isn't any benefit to using it (that I've ever seen at least) over the built-in Python 3 `venv` package.

I think it would probably be easiest for new users if we removed the requirement for `virtualenv` by using the built-in `venv`. This is how I usually teach Python 3 and this is what the Django Girls tutorial [recommends](https://tutorial.djangogirls.org/en/installation/#virtual-environment).

### DIFF
--- a/docs/internals/contributing.rst
+++ b/docs/internals/contributing.rst
@@ -23,7 +23,7 @@ instead of using the official PyBee repository, you'll be using your own
 Github fork.
 
 As with the getting started guide, these instructions will assume that you
-have Python 3.4+, a Java 7 or Java 8 JDK, and Apache ANT installed, and have virtualenv available for use.
+have Python 3.4+, a Java 7 or Java 8 JDK, and Apache ANT installed.
 
 **Note:** If you are on Linux, you will need to install an extra package to be able to run the test suite. 
  * **Ubuntu** 12.04 and 14.04: ``libpython3.4-testsuite`` This can be done by running ``apt-get install libpython3.4-testsuite``.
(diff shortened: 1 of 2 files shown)


docs/internals/contributing.rst
```diff
@@ -51,7 +51,7 @@ For Windows the use of cmd under Administrator permission is suggested instead o
 
 .. code-block:: batch
 
-    > virtualenv -p "C:\Python34\python.exe" env
+    > C:\Python34\python.exe -m venv env
```
goanpeca: This assumes users will install on that location. This is not what windows10 offers as a default location.

treyhunner: I agree that this is a problem. I was planning to make another pull request after this one to remove that full path and use the built-in `py -3` that comes when install Python on Windows.

treyhunner: A full path is currently already assumed to be that location. So while still not fully accurate, I do think this is an improvement on what is currently in the tutorial.

treyhunner:
I also changed the full path to `py -3` while I was at it. [That works on Windows](https://docs.python.org/3.4/using/windows.html?highlight=windows#from-the-command-line).

Re-review needed @freakboy3742.

PR #609, opened 2017-08-07
Refs #36: Implement power between int and float, and inplace power between int and (int or float)

Venergon:


Venergon:
For some reason this isn't running all of the tests, I don't know why

PR #623, opened 2017-08-09
Refs #212

chandlerbaggett:
Updated annotations for methods in the builtins and individual datatypes to reflect the API and documentation that is exposed by Python. Fixed something since last commit

### DIFF
--- a/python/common/org/python/types/Bool.java
+++ b/python/common/org/python/types/Bool.java
@@ -61,7 +61,7 @@ public class Bool extends org.python.types.Object {
     // }
 
     @org.python.Method(
-            __doc__ = ""
+            __doc__ = "Return repr(self)."
     )
     public org.python.types.Str __repr__() {
         if (this.value) {
@@ -86,7 +86,7 @@ public class Bool extends org.python.types.Object {
     }
 
     @org.python.Method(
-            __doc__ = "",
+            __doc__ = "Return self<value."
             args = {"other"}
     )
     public org.python.Object __lt__(org.python.Object other) {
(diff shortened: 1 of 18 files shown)


freakboy3742:
If you're looking for something to follow up with - you've covered the `__doc__` - the next step is to describe all the args and kwargs accepted by each method.

chandlerbaggett:
Yeah I can do that. Are you guys using some sort of website that has all the args and kwargs listed or just filling it out through your own gumption? I tried using something like thing.thinginthing.args and wasn't getting anything.

Thanks!
Chandler

eliasdorneles:
Hi @chandlerbaggett, great work, thanks for contributing ! :heart: 

So, CPython 3 is the source of truth.
Usually to find out what should be accepted, I look at the help of a given type or method using the CPython shell:

    >>> help(int)  # will show all the help for the integer type, with a list of all the methods and their help
    >>> help(int.__round__)  # you can also see only the help for a specific method

You can [see here the method `Int.__round__` in VOC](https://github.com/pybee/voc/blob/b1174a77e185949808deadb2cb4475151db157bd/python/common/org/python/types/Int.java#L868-L877), you'll notice that the Java method already accepts an `ndigits`, but it's missing the declaration in the `@org.python.Method` signature, which means if you try to call `print((3.14159).__round__(3))` it will fail with unexpected arguments error. It needs [`ndigits` declared as a default args in this case](http://voc.readthedocs.io/en/latest/internals/signatures.html#a-function-with-a-default-argument), so like `"default_args": {"ndigits"}`.

You will find other examples of methods with this problem looking for other methods that have a Java signature but are missing the Python signature ([skimming through Dict.java I also found these](https://github.com/pybee/voc/blob/master/python/common/org/python/types/Dict.java#L285-L310))

Does this make sense?

chandlerbaggett:
@eliasdorneles Yes it does! I appreciate you walking through that with me. I'll take a shot tonight.

Regards,
Chandler

Issue #348, opened 2017-01-28
Integrate Ouroboros into the build


freakboy3742:
[pybee/Ouroboros](https://github.com/pybee/ouroboros) provides a pure Python implementation of the Python standard library. That Python code can be compiled directly into Java classifies and converted into a standard library that can ship as part of the VOC support libraries, supplemented by native implementations as required (see pybee/voc#74).

eliasdorneles:
So, how do you think this integration could work initially?

Right now, ouroboros doesn't compile with VOC yet, do you think we should get VOC complete enough to compile ouroboros before doing this integration?

Or do you think it would be valuable start with something like, adding an extra build step that tries to compile every file in ouroboros, check how many compile without errors and report if that number diminishes?
I know that simply compiling without errors doesn't mean it's generating the correct code, but it may be a cheap way to prevent some kinds of breaking changes.

eliasdorneles:
I've just ran a quick check in my laptop, to see how many files would compile without errors:

    for f in ouroboros/*.py; do voc $f && (echo WORKS >> log.txt) || (echo FAILS >> log.txt); done

It resulted in this:

    $ sort log.txt | uniq -c
    125 FAILS
     45 WORKS

Well, this is a run only for the files in the top level, but it means it's compiling ~26% of the files without errors.
It would be cool to ensure that this percentage doesn't go down.

eliasdorneles:
Full output that can be useful: https://gist.github.com/eliasdorneles/2087a1f8df316d23ac0520533c9c47a5
I'll mine it for interesting bugs later. :)

freakboy3742:
An approach like Batavia uses would be helpful - keep a list of modules that are implemented and tested, and use that list as the manifest of what to compile into the main sources. That effectively white-lists modules that we know work (like the 45 that are currently compiling without problem)

The debug output is definitely interesting, though - lots of good material to mine there.

eliasdorneles:
I've re-run the compilation for Ouroboros using this command and I've added the results [to this gist](https://gist.github.com/eliasdorneles/c64fae2d6aa1dea85602be29508f3c40), together with a simple log analysis command that display the most common error messages, this may help prioritize what to fix.

eliasdorneles:
I think we can close this one already, the tool from #626 is running every time we do `ant java`.

PR #626, opened 2017-08-11
Make test script to compile ouroboros with voc 

goanpeca:
Related to #348

After a quick chat with @eliasdorneles I came up with this script, it can be run with 

`$ python tools/compile_stdlib.py`

And outputs something like:

```bash
$ python tools/compile_stdlib.py

UPDATING OUROBOROS CODE
=======================
Pulling from master on repo: https://github.com/pybee/ouroboros.git

Repo dir: /<BASE_DIR>/voc/ouroboros-repo/ouroboros

Already up-to-date.

From https://github.com/pybee/ouroboros
 * branch            master     -> FETCH_HEAD


Found: 716 files



START VOC COMPILE TESTS
=======================
...
PASSED: ouroboros-repo/ouroboros/turtledemo/clock.py -> build/java/python/turtledemo/clock.class
PASSED: ouroboros-repo/ouroboros/venv/__init__.py -> build/java/python/venv/__init__.class
PASSED: ouroboros-repo/ouroboros/xml/parsers/expat.py -> build/java/python/xml/parsers/expat.class
xfail:  ouroboros-repo/ouroboros/plat-sunos5/IN.py
PASSED: ouroboros-repo/ouroboros/wsgiref/__init__.py -> build/java/python/wsgiref/__init__.class
PASSED: ouroboros-repo/ouroboros/turtledemo/colormixer.py -> build/java/python/turtledemo/colormixer.class
PASSED: ouroboros-repo/ouroboros/xml/__init__.py -> build/java/python/xml/__init__.class
PASSED: ouroboros-repo/ouroboros/turtledemo/forest.py -> build/java/python/turtledemo/forest.class
PASSED: ouroboros-repo/ouroboros/xml/dom/__init__.py -> build/java/python/xml/dom/__init__.class
PASSED: ouroboros-repo/ouroboros/xml/etree/__init__.py -> build/java/python/xml/etree/__init__.class
xfail:  ouroboros-repo/ouroboros/plat-sunos5/STROPTS.py
PASSED: ouroboros-repo/ouroboros/plat-sunos5/TYPES.py -> build/java/python/plat-sunos5/TYPES.class
PASSED: ouroboros-repo/ouroboros/plat-unixware7/IN.py -> build/java/python/plat-unixware7/IN.class
PASSED: ouroboros-repo/ouroboros/plat-unixware7/STROPTS.py -> build/java/python/plat-unixware7/STROPTS.class
...

Process duration: 0:03:00.079574

Tests completed. 549/716 files compiled successfully (76.7%)
```

It takes 2 minutes using all of my 8 cores with this machine:

screen shot 2017-08-11 at 13 39 56


tests/compile_ouroboros.py
```diff
@@ -0,0 +1,670 @@
...
+
+    repo_folder = ouroboros_folder()
+    if os.path.isdir(repo_folder):
+        print('Pulling from master on repo: {}\n'.format(url))
+        args = ['git', 'pull', 'origin', 'master']
+    else:
+        print('Cloning repo: {}\n'.format(url))
+        args = ['git', 'clone', url]
+
+    p = Popen(args, stdout=PIPE, stderr=PIPE, cwd=repo_folder)
```
eliasdorneles: It is failing for me, saying that the `ouroboros/ouroboros` directory doesn't exist:

```
$ python tests/compile_ouroboros.py 

UPDATING OUROBOROS CODE
=======================
Cloning repo: https://github.com/pybee/ouroboros.git

  File "tests/compile_ouroboros.py", line 658, in <module>
    clone_update_repo()
  File "tests/compile_ouroboros.py", line 551, in clone_update_repo
    p = Popen(args, stdout=PIPE, stderr=PIPE, cwd=repo_folder)
  File "/home/elias/.pythonz/pythons/CPython-3.4.4/lib/python3.4/subprocess.py", line 859, in __init__
    restore_signals, start_new_session)
  File "/home/elias/.pythonz/pythons/CPython-3.4.4/lib/python3.4/subprocess.py", line 1463, in _execute_child
    raise child_exception_type(errno_num, err_msg)
FileNotFoundError: [Errno 2] No such file or directory: '/home/elias/src/voc-dev/voc/ouroboros/ouroboros'
```

tests/compile_ouroboros.py
```diff
@@ -0,0 +1,670 @@
...
+    'ouroboros/xml/sax/saxutils.py',
+    'ouroboros/xml/sax/xmlreader.py',
+]
+
+
+def ouroboros_folder():
+    """Return the folder where the ouroboros repo was cloned."""
+    os.chdir(HERE)
+    os.chdir('../')
+    return os.path.join(os.getcwd(), 'ouroboros', 'ouroboros')
```
eliasdorneles: Can't this be: `os.path.realpath(os.path.join(HERE, '../', 'ouroboros'))` ? (to avoid the `chdir`)

tests/compile_ouroboros.py
eliasdorneles: After bypassing the above error, `git pull` is just updating my VOC cloned repo.

Now that I think of it, it's probably a bad idea to have a git repo inside another git repo.
Maybe it could go take a parameter for the ouroboros directory, and if not given it would download from Github to a temporary directory.

What do you think?

tests/compile_ouroboros.py
```diff
@@ -0,0 +1,670 @@
...
+"""Check modules from Ouroboros that compile with voc."""
+
+# Standard library imports
+from datetime import datetime
+from subprocess import Popen, PIPE
+import multiprocessing
+import os
+
+HERE = os.path.dirname(os.path.realpath(__file__))
+MODULES_WHITELIST = [
```
eliasdorneles: Maybe it would be best to have the other way around: instead of a whitelist, this would be an "ignorelist" of the things we know we aren't compiling yet -- just so that it works also as a "TODO" list.

freakboy3742: +1 - what remains to be done is a much more interesting list.

freakboy3742: (and the same change to Batavia would also be helpful :-)

tests/compile_ouroboros.py
```diff
@@ -0,0 +1,670 @@
+# -*- coding: utf-8 -*-
```
freakboy3742: For consistency, having this as "compile_stdlib.py" for consistency with batavia would be helpful.

tests/compile_ouroboros.py
freakboy3742: One other nitpick - there needs to be a mode for overrides. While in many cases we will be able to just compile the `.py` file from ouroboros and use it, for performance we may sometimes use a native Java implementation of a module. So - if there's a native module in the `python` subdirectory, the ouroboros version should be skipped.

tests/compile_ouroboros.py
goanpeca: Probably :-p

tests/compile_ouroboros.py
goanpeca: Sure, I can do that

tests/compile_ouroboros.py
goanpeca: @freakboy3742 are the modules inside python subdirectory following the same structure, so that I can add the override mode based on that?

goanpeca:
>The one piece that appears to be missing (from a quick review - I haven't taken this for a spin) is integration into the overall build system. When you do a full ant build, it produces a python-java-support.jar and python-android-support.jar; the files generated by this script should be integrated into that jar file as part of a full ant build.

I will need some help here @freakboy3742

tests/compile_ouroboros.py
goanpeca: Should I move this to the root of the repo as well?

goanpeca:
@freakboy3742 do you mean that the overrides correspond to `/voc/python/common/python` ?

tests/compile_ouroboros.py
freakboy3742: Actually, putting it in the "tools" subdirectory would be better. There's no analog in Batavia (yet), but there probably should be.

tests/compile_ouroboros.py
freakboy3742: There will either be a `python/<platform>/python/<module>.java`, or a `python/<platform>/python/<module>/__init__.java`, where `<platform>` is one of `common`, `java` or `android`. 

By chance, one of the modules you've flagged as working already has a native implementation - `types`.

freakboy3742:
@goanpeca The Java source code compiles into the `build/android` and `build/java` directories; the jar files are then just the contents of these directories. The script should output into the same build directories, so that the compiled class files form part of the jar file. They shouldn't be checked in, because they're compiled artefacts.

goanpeca:
@freakboy3742 got it. Now another question is... voc always compiles inside a python folder?

I am not sure what should be the right output of for example

`ouroboros/async/somefile.py` would be compiled to `build/java/python/async/somefile.java` ?

tests/compile_ouroboros.py
eliasdorneles: Wouldn't it be better to do the overrides via a configuration instead of convention?
For cases like the Java implementation is buggy for a particular module or for some other reason someone wants the compiled Python version?

goanpeca:
@eliasdorneles I fixed most of the issues raised. Could you give it a try?

I need to add a small argparser to allow choosing between the java or android target. And I need to test this on windows.

freakboy3742:
@goanpeca I don't think a java/android flag is needed, as you won't get different output on each platform. The only difference is the output directory. 

Ideally, we'd only need to compile the standard library once (say, to `build/common`), and then have it included in the two different output jar files. The same is true of the build instructions for the Java code that is part of VOC itself - but since the Ouroboros parts will take much longer to compile, it makes sense to do this cleanup now.

eliasdorneles:
Nice work, @goanpeca !
I gave it a try, worked nicely for me (took `0:03:49.871943` in my machine). :+1:

goanpeca:
@freakboy3742 this already compiles to the build folder as suggested. Any other comments?

goanpeca:
Cool :-p

eliasdorneles:
Awesome work, fellows! :)

PR #639, opened 2017-08-22
Implement builtin reversed

uranusjr:
* [x] Make the reversed builtin callable.
* [x] Implement generic reversed type on top of `__len__` and `__getitem__`.
* [x] Implement `__reversed__` on `list` to return a `list_reverseiterator`.
* [x] Implement `__reversed__` on `range` to return a `range_iterator`. [^a]
* [x] Add tests to `test_reversed.py::ReversedTests` to actually test running through the iterators.

[^a]: Reverse of a range is `range(start+(n-1)*step, start-step, -step)`, where `n` is the number of integers in the range. The principle is simple, but implementation can be very tedious to dive into.

### DIFF
--- a/python/common/org/Python.java
+++ b/python/common/org/Python.java
@@ -1451,7 +1451,32 @@ public class Python {
             args = {"sequence"}
     )
     public static org.python.Object reversed(org.python.Object sequence) {
-        return sequence.__reversed__();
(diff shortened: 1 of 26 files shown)


uranusjr:
🍸

PR #645, opened 2017-08-30
[WIP] Implemented extend method for List

amymok:
## Additions:
- Added logic for `List.extend`
- Added corresponding tests

## Todos:
- Make sure all datatypes are working correctly

### DIFF
--- a/python/common/org/python/types/List.java
+++ b/python/common/org/python/types/List.java
@@ -649,9 +649,31 @@ public class List extends org.python.types.Object {
     }
 
     @org.python.Method(
-            __doc__ = "L.extend(iterable) -> None -- extend list by appending elements from the iterable"
+            __doc__ = "L.extend(iterable) -> None -- extend list by appending elements from the iterable",
+            args = {"other"}
     )
     public org.python.Object extend(org.python.Object other) {
+        if (other instanceof org.python.types.List) {
+            this.value.addAll(((org.python.types.List) other).value);
+        } else if (other instanceof org.python.types.FrozenSet) {
+            this.value.addAll(((org.python.types.FrozenSet) other).value);
+        } else if (other instanceof org.python.types.Set) {
+            this.value.addAll(((org.python.types.Set) other).value);
+        } else if (other instanceof org.python.types.Tuple) {
+            this.value.addAll(((org.python.types.Tuple) other).value);
+        } else if (other instanceof org.python.types.Dict) {
+            this.value.addAll(((org.python.types.Dict) other).value.keySet());
+        } else if (other instanceof org.python.types.Iterator) {
+            try {
+                while (true) {
+                    org.python.Object next = other.__next__();
+                    this.value.add(next);
+                }
+            } catch (org.python.exceptions.StopIteration si) {
+            }
+        } else {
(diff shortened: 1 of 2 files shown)


tests/datatypes/test_list.py
```diff
@@ -119,6 +119,62 @@ def test_append(self):
...
+            x.extend((5, "world", "hello"))
+            print(x)
+            """)
+
+        # extend a frozenset
+        self.assertCodeExecution("""
+            x = [1, 2, 3]
+            y = frozenset([8, "theta"])
+            x.extend(y)
+            set([x[3], x[4]]) == y
```
freakboy3742: I can see what you're trying for here, but there are two problems:

Firstly, this won't actually test anything. `assertCodeExecution()` performs it's test based on the output of the program. It's not enough to just do the equality test - you actually have to print something, so that there's something in the output to compare.

Secondly, if the problem you're getting is output ordering - we've got a mechanism in place to work around this. Any TranspileTestCase instance can define a "substitutions" dictionary. If present, any "values" in the dictionary that are present in the sample output will be replaced by it's "key". Essentially, this normalises unpredictable output so that it will reliably pass tests.

There's an example in the InplaceListOperationTests class in the same file that programatically generates a substitutions dictionary, but for simpler cases, you can statically define them, too.

amymok: Oops, yes forgot the `print`.

Ah, I see how you are doing it, will take a crack at it tomorrow.

Thanks for the feedback.

python/common/org/python/types/List.java
```diff
@@ -661,6 +661,27 @@ public void __delitem__(org.python.Object index) {
...
+        if (other instanceof org.python.types.List) {
+            this.value.addAll(((org.python.types.List) other).value);
+        } else if (other instanceof org.python.types.FrozenSet) {
+            this.value.addAll(((org.python.types.FrozenSet) other).value);
+        } else if (other instanceof org.python.types.Set) {
+            this.value.addAll(((org.python.types.Set) other).value);
+        } else if (other instanceof org.python.types.Tuple) {
+            this.value.addAll(((org.python.types.Tuple) other).value);
+        } else if (other instanceof org.python.types.Dict) {
+            this.value.addAll(((org.python.types.Dict) other).value.keySet());
```
uranusjr: I wonder whether the above can be vastly simplified with a generic method…

tests/datatypes/test_list.py
```diff
@@ -724,7 +752,7 @@ class InplaceListOperationTests(InplaceOperationTestCase, TranspileTestCase):
     data_type = 'list'
 
     # Several tests add a list to an iterable with a non-deterministic
-    # ordering. We define these substitions here where the first part of the
+    # ordering. We define these subsitutions here where the first part of the
```
uranusjr: The typing is still wrong 😂😂😂😂

amymok: oh great :)

python/common/org/python/types/List.java
amymok: I will look into that.

tests/datatypes/test_list.py
amymok: Added a new function here to create the substitutions for those tests

Issue #649, opened 2017-09-05
[Proposal] Have a Java inheritance chain for standard types implemented


uranusjr:
Currently (almost) all types inherit from `org.python.types.Object`. While this reflects the inheritance chain on the Python side (most types inherit from `object`), this makes the implementation unnecessarily difficult, and require a lot of duplicate code because we can’t really utilise Java polymorphism idioms, such as Generics and Interfaces. This is most problematic in collection types, but also poses a problem with other types.

One example is the `List::extend` implementation merged in #645—the whole if-else block is unnecessary if Generics can me used.

As @freakboy3742 mentions in #486, the Java inheritance chain does not affect the Python inheritance chain. Since inheritance and interfacing is front and centre in Java’s OOP, we should use it.

Python provides a set of abstract base classes that can offers a good sense how the inheritance chain should look like. Some extra classes might be necessary (I think) to abstract the internal structures, but not much. This could probably be a beginning for solution toward #582.

cflee:
This is a very interesting idea. Interfaces certain seem like they could be very useful, as long as we can contain and prevent them from exploding in number with some careful design...

freakboy3742:
This definitely looks like something worth exploring - although I suspect the devil will be in the details. 

Have you got any concrete ideas for how to implement this? It is a case of picking features one at a time and interfacing/inheriting them until everything is cleaned up?

uranusjr:
Not really, what I currently have in mind is to loosely follow the collection ABCs, implementing what we think is good along the way. Classes that can provide the most benefits right now IMO are Iterable  (#645 and a lot of other similar code would benefit) and ByteString (Bytes and Bytearray rightly share most of their implementation). Another one I can think of is the number classes (#582).

PR #651, opened 2017-09-13
[POC] Use interface to simply implementation

uranusjr:
I dabbled at how it makes the most sense to implement #649. Eventually what makes the most sense is actually some bridge interfaces that provides access to the underlying Java implementation, and just use (Java’s) native reflection to back the Python interface. I tried this on some collection classes to simplify `List.extend` (and related methods).

I put the interface in `org.python.java` since this is more indicative to the Java implementation rather than the Python interface.

I’m going to try this on some other parts of the source and see how things go. Please do provide any feedback!


freakboy3742 (commented):
This looks pretty solid to me. 

The only question I'd have is whether the `python/java/XXX.java` interfaces should be named/structured around the *Java* naming or the *Python* naming. At the implementation level, they're using `java.util.Collection`, but the "interface" they're implementing is "iterable" or "file-like object" or similar.

uranusjr:
I have doubts on naming as well, but unfortunately I don’t have much experience in Java API design and couldn’t come up something better. The best I can think of is `CollectionBacked`…

uranusjr:
Pushed a large changeset. Int and Bool are now all cleaned up and share most of their implementation via the `org.python.internals.IntegerOperation` helper class. I think this is better than subclassing because the call graph is much easier to trace. Yes there are quite some duplicate code bridging `__magic__` methods to `IntegerOperation`, but I think this is actually better because we can’t guarantee them to always share the same implementation.

uranusjr:
Oh wow, so Java didn’t have `Math.floorMod()` before 1.8!? How can anyone bare with this standard library for so long I could never understand.

phildini:

PR #656, opened 2017-10-01
 Add test/method __contains__ in bytes

JvMaiia:


tests/datatypes/test_bytes.py
```diff
@@ -216,6 +216,13 @@ def test_index(self):
             print(b'abcd'.find(b'ab', 3, 4))
         """, exits_early=True)
 
+    def test_contains(self):
+        self.assertCodeExecution("""
+            print(b'py' in b'pybee')
+            print(b'bee' in b'pybee')
+            print(b'ybe' in b'pybee')
+            print(b'test' in b'pybee')
```
eliasdorneles: This is a good start!

What about handling cases like:

```
>>> 100 in b'pybee'  # 100 is the character b'd'
False
>>> 101 in b'pybee'  # 101 is the character b'e'
True
>>> 300 in b'pybee'  # should throw ValueError because 300 is not a valid byte
ValueError: byte must be in range(0, 256)
```

python/common/org/python/types/Bytes.java
```diff
@@ -170,10 +170,31 @@ public Bytes(org.python.Object[] args, java.util.Map<java.lang.String, org.pytho
...
+                    counter_slice++;
+                }
+                else {
+                    counter_slice = 0;
+                }
+            }
+            return new org.python.types.Bool(false);
+        }
+        else {
+              throw new org.python.exceptions.TypeError("invalid type for this method");
```
eliasdorneles: The exception messages should match the messages that CPython does.
CPython (testing on 3.5.2) throws a different message, see:

```
>>> ['p', 'y'] in b'pybee'
TypeError: a bytes-like object is required, not 'list'
```

You can build the message like it's being done here: https://github.com/pybee/voc/blob/9e0dca72dc874249977a11ea40bf112f7af2485f/python/common/org/python/types/Bytes.java#L667

JvMaiia:
Thanks for your help, @eliasdorneles !

Issue #640, opened 2017-08-22
Is the checkstyle job on CI broken?


uranusjr:
Running `ant checkstyle` locally now generates errors:

```
Buildfile: /Users/uranusjr/Documents/programming/python/voc/voc/build.xml

checkstyle:
[checkstyle] Running Checkstyle 6.19 on 133 files
[checkstyle] [ERROR] /Users/uranusjr/Documents/programming/python/voc/voc/python/common/org/python/types/Range.java:75:86: 'typecast' is not followed by whitespace. [WhitespaceAfter]
[checkstyle] [ERROR] /Users/uranusjr/Documents/programming/python/voc/voc/python/common/org/python/types/Range.java:91:51: 'typecast' is not followed by whitespace. [WhitespaceAfter]
[checkstyle] [ERROR] /Users/uranusjr/Documents/programming/python/voc/voc/python/common/org/python/types/Range.java:185:69: 'typecast' is not followed by whitespace. [WhitespaceAfter]
[checkstyle] [ERROR] /Users/uranusjr/Documents/programming/python/voc/voc/python/common/org/python/types/Slice.java:39:38: 'typecast' is not followed by whitespace. [WhitespaceAfter]
[checkstyle] [ERROR] /Users/uranusjr/Documents/programming/python/voc/voc/python/common/org/python/types/Slice.java:46:42: 'typecast' is not followed by whitespace. [WhitespaceAfter]
[checkstyle] [ERROR] /Users/uranusjr/Documents/programming/python/voc/voc/python/common/org/python/types/Slice.java:63:19: Use a single space to separate non-whitespace characters. [SingleSpaceSeparator]
[checkstyle] [ERROR] /Users/uranusjr/Documents/programming/python/voc/voc/python/common/org/python/types/Slice.java:63:45: 'typecast' is not followed by whitespace. [WhitespaceAfter]
[checkstyle] [ERROR] /Users/uranusjr/Documents/programming/python/voc/voc/python/common/org/python/types/Slice.java:80:43: 'typecast' is not followed by whitespace. [WhitespaceAfter]

BUILD FAILED
/Users/uranusjr/Documents/programming/python/voc/voc/build.xml:31: Got 9 errors and 0 warnings.

Total time: 2 seconds
```

None of those are reported by the CI. I don’t have enough knowledge on either Ant, Checkstyle, or Beekeeper, but this smells wrong to me…

cflee:
[...]

PR #658, opened 2017-10-07
Refs #36 Implement binary slice comparison operations

thehungrysmurf:
This change implements the following binary slice comparison operations:

```
__eq__
__gt__
__ge__
__lt__
__le__
```

`__ne__` is also implemented as a byproduct, as it uses the inverse of `__eq__`.

In Python3 `__cmp__()` is eliminated and all comparisons are now rich comparisons, so this change covers the corresponding `direct` (rich comparison) binary operations as well.

I'm a first time contributor to this project, feedback is very much appreciated!

### DIFF
--- a/python/common/org/python/types/Slice.java
+++ b/python/common/org/python/types/Slice.java
@@ -177,7 +177,9 @@ public class Slice extends org.python.types.Object {
             org.python.types.Tuple thisSliceAsTuple = convertSliceToTuple(this);
             return thisSliceAsTuple.__ge__(otherSliceAsTuple);
         }
-        else return org.python.types.NotImplementedType.NOT_IMPLEMENTED;
+        else {
+            return org.python.types.NotImplementedType.NOT_IMPLEMENTED;
+        }
     }
 
     @org.python.Method(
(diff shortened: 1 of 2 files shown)


eliasdorneles:
Hello @thehungrysmurf , thanks for contributing, this looks good functionality-wise! :+1: 
It appears that the build didn't finish because of the style checks, can you have a look?
It says:

```
  * python/common/org/python/types/Slice.java
    - Line 180: [NeedBraces] 'else' construct must use '{}'s.
```

thehungrysmurf:
Thanks a lot @eliasdorneles ! 
Sorry for the delay, I hadn't realized the tests were staggered to fail fast. I'd have fixed the issue sooner. 
Question for ya: when I ran the style check locally (`flake8 && ant checkstyle`) this problem didn't come up. Is there a different local style check I should be running?

eliasdorneles:
Thank _YOU_, @thehungrysmurf :)
Yeah, we're having problems with the style check, it's indeed being different than what we're running locally but I still haven't figured it out (there is a hint in #640 ).

For now, I recommend doing `flake8; ant checkstyle` (namely, avoid the `&&`, because it will skip the Java checks if any Python check fails).

PR #665, opened 2017-10-10
Add __contains__ for bytearray type

nickbom:
Implementation of "__contains__()" builtin method for bytearray type with 10 tests cases on "test_bytearray.py".

### DIFF
--- a/tests/datatypes/test_bytearray.py
+++ b/tests/datatypes/test_bytearray.py
@@ -19,6 +19,29 @@ class BytearrayTests(TranspileTestCase):
             except AttributeError as err:
                 print(err)
             """)
+    def test_contains(self):
+        self.assertCodeExecution("""
+            x = bytearray([1,2,3])
+            y = bytearray([1,2])
+            print(x in y)
+            print(y in x)
+            """)
+        self.assertCodeExecution("""
+            print(bytearray([1,2,3]) in bytearray([1,2]))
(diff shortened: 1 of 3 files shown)


tests/datatypes/test_bytearray.py
```diff
@@ -19,6 +19,23 @@ def test_getattr(self):
...
+            print(bytearray([1,2]) in bytearray([1,2,3]))
+            print(bytearray([1,2,4]) in bytearray([1,2,3]))
+            print(bytearray([8,9,0,1]) in bytearray([1,2,3]))
+            print(101 in bytearray([1,2,3]))
+            print(101 in bytearray([1,2,3,101]))
+            print(b'pybee' in bytearray([1,2]))
+            print(bytearray([1,2]) in b'pybee')
+        """)
+        self.assertCodeExecution("""
+            print(300 in bytearray([1,2,3])) #Test ValueError invalid integer range
```
freakboy3742: The better way to test this is to use a try/catch block:
```
try:
    print(300 in bytearray([1,2,3]))
    print("No error raised")
except ValueError:
    print("Raised a ValueError")
```
This code will *always* run; but if a ValueError isn't raised, there will be a discrepancy in the output (which will be detected and reported)

nickbom: Changed that, thanks

PR #667, opened 2017-10-10
Add isspace for bytes and byte array

Dayof:
- Implementation of "isspace()" builtin method for bytes with 5 tests cases on "test_bytes.py";
- Implementation of "isspace()" builtin method for byte array with 5 tests cases on "test_bytearray.py";

### DIFF
--- a/tests/datatypes/test_bytearray.py
+++ b/tests/datatypes/test_bytearray.py
@@ -22,11 +22,14 @@ class BytearrayTests(TranspileTestCase):
 
     def test_capitalize(self):
         self.assertCodeExecution("""
-            print(bytearray(b"abc").capitalize())
+            print(bytearray(b'abc').capitalize())
             print(bytearray().capitalize())
-            """)
+        """)
 
     def test_islower(self):
+        # TODO: add this test when adding support for literal hex bytes
+        # print(b'\xf0'.islower())
+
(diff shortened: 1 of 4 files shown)


tests/datatypes/test_bytearray.py
```diff
@@ -40,6 +40,15 @@ def test_islower(self):
         """)
         # self.assertCodeExecution("""""")
 
+    def test_isspace(self):
+        self.assertCodeExecution("""
+            print(bytearray(b'testupper').isspace())
+            print(bytearray(b'test isspace').isspace())
+            print(bytearray(b' ').isspace())
+            print(bytearray(b'').isspace())
+            print(bytearray(b'\x46').isspace())
```
eliasdorneles: Can you please add these cases?

```
b'   \t\t' # several spaces mixed with tabs
b' \t\n\r\x0b\f' # others kinds of space characters
```

See: https://docs.python.org/3/library/stdtypes.html#bytearray.isspace

Dayof:
@eliasdorneles I added support to ``\t``, ``\x0b`` and ``\f``, but for the cases bellow shows the following error ``AssertionError: EOL while scanning string literal``. Do you know how to fix this?

```
print(b' \n'.isspace())
print(b' \r'.isspace())
```

PR #675, opened 2017-10-14
Adding center to bytearray

nickbom:
Adding center to bytearray with 10 test cases and adding bytearray to arguments of bytes center with 1 test case for that. `print(b'pybee'.center(True, bytearray(b'a')))`

### DIFF
--- a/python/common/org/python/types/ByteArray.java
+++ b/python/common/org/python/types/ByteArray.java
@@ -689,7 +689,7 @@ public class ByteArray extends org.python.types.Object {
         byte[] fillByte;
         if (byteToFill instanceof org.python.types.ByteArray || byteToFill instanceof org.python.types.Bytes) {
             if(byteToFill instanceof org.python.types.ByteArray){
-                fillByte = ((org.python.types.ByteArray) byteToFill).value; 
+                fillByte = ((org.python.types.ByteArray) byteToFill).value;
             } else {
                 fillByte = ((org.python.types.Bytes) byteToFill).value; 
             }
--- a/python/common/org/python/types/ByteArray.java
+++ b/python/common/org/python/types/ByteArray.java
@@ -691,7 +691,7 @@ public class ByteArray extends org.python.types.Object {
             if(byteToFill instanceof org.python.types.ByteArray){
                 fillByte = ((org.python.types.ByteArray) byteToFill).value;
             } else {
-                fillByte = ((org.python.types.Bytes) byteToFill).value; 
+                fillByte = ((org.python.types.Bytes) byteToFill).value;
             }
             if (fillByte.length != 1) {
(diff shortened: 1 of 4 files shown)


python/common/org/python/types/ByteArray.java
```diff
@@ -679,10 +679,66 @@ public boolean __setattr_null(java.lang.String name, org.python.Object value) {
...
-    public org.python.Object center(java.util.List<org.python.Object> args, java.util.Map<java.lang.String, org.python.Object> kwargs, java.util.List<org.python.Object> default_args, java.util.Map<java.lang.String, org.python.Object> default_kwargs) {
-        throw new org.python.exceptions.NotImplementedError("bytearray.center has not been implemented.");
+            __doc__ = "B.center(width[, fillchar]) -> copy of B" +
+            "\n\nReturn B centered in a string of length width.  Padding is" +
+            "\ndone using the specified fill character (default is a space).",
+            args = {"width"},
+            default_args = {"byteToFill"}
+    )
+    public org.python.Object center(org.python.Object width, org.python.Object byteToFill) {
+        byte[] fillByte;
```
eliasdorneles: Hi @nickbom, thanks for contributing!
But note that this is duplicating the code from Bytes in ByteArray, but it seems you wanted to do otherwise -- was this really intended?
Thanks!

nickbom: Hi @eliasdorneles, this is because in bytearray.center needs to return bytearray, diferent from the other methods I used bytes implementation, which returned int or bool. I prefered to duplicate the code, so don't need to convert Bytes to bytearray, what you think is a better way to do this, convert or duplicate ?

eliasdorneles: I like best the way [`lower` and `capitalize` are implemented](https://github.com/pybee/voc/blob/18f00f2ce7e5e97abb75fe85a273a0d5acdbb421/python/common/org/python/types/Bytes.java#L1101-L1115), they have a static version (`_lower` and `_capitalize`) that uses only the Java types, and then `lower` and `capitalize` are responsible for calling the static function and converting to the Python types.

eliasdorneles: Then, from ByteArray, it calls the Java-only version directly with `Bytes._lower(this.value)` and `Bytes._capitalize(this.value)`

eliasdorneles: I think in this case the Java-only code would return an `ArrayList<Byte[]>` -- does this make sense?
Another option would be to call `Bytes.center(new Bytes(this.value))`, which will return a `org.python.types.List`, and then iterate through that list creating `ByteArray` objects.

nickbom: I finished just returning `byte[]`, check it out if it's okay

PR #681, opened 2017-10-26
Implement bytes.istitle method

onyb:


onyb:
@eliasdorneles Nice catch! Indeed both the tests you propose were failing. Please re-review.

I also took opportunity to simplify the `bytes.title` method, taking inspiration from the Apache Commons.

onyb:
@eliasdorneles I noticed that the `str.istitle` and `str.title` implementations are inaccurate, and would fail most of the tests implemented in this PR. Do you think I should fix them in a separate PR?

eliasdorneles:
Looks great, thanks a lot @onyb ! :star2: 
And yes, feel free to send another PR for str.istitle and str.title =)

PR #682, opened 2017-10-26
Add more tests for the sum builtin

onyb:
**Note:** The goal of this PR is to add some meaningful tests for `sum`, as this builtin has many use-cases.


onyb:
Thanks for your review. I will keep these remarks in mind in the future.

PR #685, opened 2017-10-27
Fix str.title and str.istitle implementations

onyb:
The tests are taken from `tests/datatypes/test_bytes.py`, and some of them were failing. I made small adjustments to the implementation of `str.title()`, while retaining the original logic; though my version of `bytes.title()` (see #680 and #681) takes a different approach.

### DIFF
--- a/python/common/org/python/types/Str.java
+++ b/python/common/org/python/types/Str.java
@@ -972,12 +972,15 @@ public class Str extends org.python.types.Object {
         if (this.value.isEmpty()) {
             return new org.python.types.Bool(false);
         }
-        for (int c = 1; c < this.value.length(); c++) {
-            if (this.value.charAt(c - 1) == ' ' && !(Character.isUpperCase(this.value.charAt(c)))) {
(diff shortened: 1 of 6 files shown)


python/common/org/python/types/Str.java
```diff
@@ -972,12 +972,15 @@ private long toPositiveIndex(long index) {
         if (this.value.isEmpty()) {
             return new org.python.types.Bool(false);
         }
-        for (int c = 1; c < this.value.length(); c++) {
-            if (this.value.charAt(c - 1) == ' ' && !(Character.isUpperCase(this.value.charAt(c)))) {
-                return new org.python.types.Bool(false);
+
+        boolean matches = this.value.equals(_title(this.value));
+
```
eliasdorneles: This works fine, but it could be better if you return false here if `matches` is false -- you only need to check if there is a letter if matches is `true`.

onyb:
**Update:** Also implemented `bytearray.title` and `bytearray.istitle` method.

PR #687, opened 2017-10-29
Add implementation for __pow__

### LINKED ISSUE #36: Complete implementation of operations on standard types

ganeshhubale:
Fixes #36

### DIFF
--- a/python/common/org/python/types/Float.java
+++ b/python/common/org/python/types/Float.java
@@ -450,7 +450,10 @@ public class Float extends org.python.types.Object {
     )
     public org.python.Object __pow__(org.python.Object other, org.python.Object modulo) {
         if (modulo != null) {
-            throw new org.python.exceptions.NotImplementedError("float.__pow__() with modulo has not been implemented");
+            //throw new org.python.exceptions.NotImplementedError("float.__pow__() with modulo has not been implemented");
+            if (other instanceof org.python.types.Float) {
+                throw new org.python.exceptions.TypeError("pow() 3rd argument not allowed unless all arguments are integers");
+
         }
 
         if (other instanceof org.python.types.Int) {


onyb:
@ganeshhubale The PR doesn't fix anything, as the check you introduced is already taken care of here: https://github.com/pybee/voc/blob/master/python/common/org/Python.java#L1308-L1328

Ideally, the `pow()` implementation linked above should do nothing more than calling `__pow__`. All the checks should instead be performed in each `__pow__` implementation. Please take a look at the following examples to know why. Note that this issue is not just limited to `pow()`.

The following works in Python:
```py
class HypotheticalInt:
    def __pow__(self, y, z):
        return 'Hypothetical result'

>>> pow(1, -2, 3)
TypeError: pow() 2nd argument cannot be negative when 3rd argument specified
>>> x = HypotheticalInt()
>>> pow(x, -2, 3)
'Hypothetical result'
```

But the same fails on VOC with an error, albeit a different one.

```py
>>> x = HypotheticalInt()
>>> pow(x, -2, 3)
TypeError: pow() 3rd argument not allowed unless all arguments are integers
```

python/common/org/python/types/Float.java
```diff
@@ -450,7 +450,10 @@ public boolean __setattr_null(java.lang.String name, org.python.Object value) {
     )
     public org.python.Object __pow__(org.python.Object other, org.python.Object modulo) {
         if (modulo != null) {
-            throw new org.python.exceptions.NotImplementedError("float.__pow__() with modulo has not been implemented");
+            //throw new org.python.exceptions.NotImplementedError("float.__pow__() with modulo has not been implemented");
+            if (other instanceof org.python.types.Float) {
```
onyb: Actually you don't need to perform this check, as `this` itself is a `Float` and using `modulo` should be disallowed regardless of whether `other` is an instance of `org.python.types.Float` or not.

eliasdorneles:
@onyb I do agree that fixing the issue you mention makes sense -- I'm not sure what's the best approach, but you're welcome to try something. :)

eliasdorneles:
@onyb you might want to have a look into the proof-of-concept in #651

PR #694, opened 2017-11-11
transpiler needs to specify source file encoding

whydoubt:
When running 'ant java' on Windows, the shlex module from ouroboros
failed to build.  The module contains UTF-8 characters that need to be
accounted for in the transpiler.


freakboy3742:
Thanks for the contribution! It’s odd that this doesn’t fail on other platforms... but it’s the right change, regardless.

PR #695, opened 2017-11-11
Updated documentation layout.

freakboy3742:


eliasdorneles:
It would be nice if we could do these changes keeping redirect links for backwards compatibility.

@goanpeca has fixed the links in the website in https://github.com/pybee/pybee.github.io/pull/252 -- but links that people added to their bookmarks or shared on twitter/etc remain broken.

goanpeca:
We could add redirects but where do we stop. I dont think its too important.

PR #702, opened 2017-11-15
Fixes exception raising (fixes #352)

### LINKED ISSUE #352: Exceptions usage fail in unexpected ways

eliasdorneles:
This PR contains some fixes on the behavior for exceptions: notably VOC doesn't crash anymore with `raise some_module.SomeException` (used in a few modules of the standard library).

It also introduces the `.args` attribute (a python tuple, mirroring CPython behavior) that contains the exception arguments and fixes behavior for multiple arguments.

There is some special behavior for KeyError as well (which prints messages differently).

Does this look good?

### DIFF
--- a/python/common/org/python/exceptions/BaseException.java
+++ b/python/common/org/python/exceptions/BaseException.java
@@ -17,8 +17,15 @@ public class BaseException extends org.python.types.Object {
     }
 
     public BaseException(org.python.Object[] args, java.util.Map<java.lang.String, org.python.Object> kwargs) {
-        super(buildTuple(args).toString());
+        super(buildMessage(buildTuple(args)));
         this.args = buildTuple(args);
+    }
+
+    private static String buildMessage(org.python.types.Tuple tupArgs) {
+        if (tupArgs.value.size() == 0) {
+            return "";
+        }
+        return tupArgs.toString();
     }
 
     private static org.python.types.Tuple buildTuple(org.python.Object[] args) {
--- a/python/common/org/python/exceptions/BaseException.java
+++ b/python/common/org/python/exceptions/BaseException.java
@@ -4,18 +4,25 @@ package org.python.exceptions;
         __doc__ = "Common base class for all exceptions"
 )
 public class BaseException extends org.python.types.Object {
+    @org.python.Attribute()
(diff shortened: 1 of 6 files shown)


python/common/org/python/exceptions/BaseException.java
```diff
@@ -4,18 +4,32 @@
...
+        this.args.value.add(new org.python.types.Str(msg));
     }
 
     public BaseException(org.python.Object[] args, java.util.Map<java.lang.String, org.python.Object> kwargs) {
-        super(args[0].toString());
+        super(buildMessage(buildTuple(args)));
+        this.args = buildTuple(args);
+    }
+
+    private static String buildMessage(org.python.types.Tuple tupArgs) {
```
eliasdorneles: Note: these static methods were needed because we need these to execute before `super()`

freakboy3742: Heh - clever. It occurs to me that this is a generic pattern that would work for any "run this code before super()" situation.

eliasdorneles:
I'd like to acknowledge the help of @onyb on investigating this and thank @dibyadas for getting the ball rolling in #470 on this -- thank you, fellows! <3

freakboy3742:
A big :clap: for @onyb and @dibyadas then :-)

Issue #610, opened 2017-08-07
Testing suite fails for str


hazzashirt:
In Windows 10 64-bit: When running the test suite via the cmd (>setup.py test...) or via Cricket, test_str in all suites fails with the following error:

`AssertionError: '>>> [321 chars]abs\n>>> x = "Mÿ hôvèrçràft îß fûłl öf éêlś"\n[319 chars]==\n' != '>>> [321 chars]abs\n<class \'UnicodeEncodeError\'> : \'charma[341 chars]==\n'`

More details:
`tests.builtins.test_any.BuiltinAnyFunctionTests.test_str`

```

Traceback (most recent call last):

  File "C:\python34\Lib\unittest\case.py", line 57, in testPartExecutor
    yield

  File "C:\python34\Lib\unittest\case.py", line 574, in run
    testMethod()

  File "c:\Users\Harry\pybee\voc-dev\voc\tests\utils.py", line 1037, in func
    substitutions=getattr(self, 'substitutions', SAMPLE_SUBSTITUTIONS)

  File "c:\Users\Harry\pybee\voc-dev\voc\tests\utils.py", line 1075, in assertBuiltinFunction
    run_in_function=False,

  File "c:\Users\Harry\pybee\voc-dev\voc\tests\utils.py", line 399, in assertCodeExecution
    self.assertEqual(java_out, py_out, context)

  File "C:\python34\Lib\unittest\case.py", line 794, in assertEqual
    assertion_func(first, second, msg=msg)

  File "C:\python34\Lib\unittest\case.py", line 1167, in assertMultiLineEqual
    self.fail(self._formatMessage(msg, standardMsg))

  File "C:\python34\Lib\unittest\case.py", line 639, in fail
    raise self.failureException(msg)

AssertionError: '>>> [157 chars]any\n>>> x = "Mÿ hôvèrçràft îß fûłl öf éêlś"\n[154 chars]==\n' != '>>> [157 chars]any\n<class \'UnicodeEncodeError\'> : \'charma[231 chars]==\n'
  >>> f = any
  >>> x = ""
  >>> f(x)
  False
  
  >>> f = any
  >>> x = "3"
  >>> f(x)
  True
  
  >>> f = any
  >>> x = "This is another string"
  >>> f(x)
  True
  
  >>> f = any
+ <class 'UnicodeEncodeError'> : 'charmap' codec can't encode character '\u0142' in position 28: character maps to <undefined>
[...]

Issue #237, opened 2016-08-16
Issues when trying to run tests on Windows


AnthonyBriggs:
(This is on Windows 8.1; I imagine similar issues appear on other Windows systems)

I've had a try at running the tests under Windows, the main failure that I run into seems to be related to the default encoding under Windows (cp1252 rather than utf-8):

`E   AssertionError: 'utf-8' codec can't decode byte 0xff in position 10: invalid start byte`

This seems to be an issue on both the Java and Python versions, see attached [voc_output.txt](https://github.com/pybee/voc/files/419318/voc_output.txt), which is a dump of the main_code for both `runAsPython()` and `runAsJava()`

Another, possibly related issue is that Windows has a "charmap" encoding in it's terminals (cmd and powershell), which doesn't handle some unicode characters. I've also attached 
[unicode_test.py.txt](https://github.com/pybee/voc/files/419326/unicode_test.py.txt) which blows up with a similar error.

In the process of debugging, I've also found a couple of places where utf-8 encoding seems to have been missed - pull request here: https://github.com/pybee/voc/pull/236

AnthonyBriggs:
I've been (slowly) chasing this up this evening, and making some progress through judicious commenting-out of exception handlers :) This is more of an infodump than anything, but might be helpful.

Commenting out the first exception handler in assertCodeExecution (lines 345/346), reveals that it's `runAsJava` that's throwing the dodgy string, specifically 

    `line = self.jvm.stdout.readline().decode("utf-8")`

on line 505. Printing out the return string from `self.jvm.stdout.readline()` shows that it's returning the hovercraft-is-full-of-eels string as

    `b'>>> x = "M\xff h\xf4v\xe8r\xe7r\xe0ft \xee\xdf f\xfb?l \xf6f \xe9\xeal?"\n'`

which I'm pretty sure is not right. `\xff` is a beginning-of-unicode-string recognition character, for a start, which is why it's exploding. (For the record, it's supposed to be "Mÿ hôvèrçràft îß fûłl öf éêlś")

I've tried a few simple things to try and work out what's going on. Encoding the java output to `cp1252` instead of utf-8 just turns it into 

     `x = "M├┐ h├┤v├¿r├ºr├áft ├«├ƒ f├╗?l ├Âf ├®├¬l?"`

and various other combinations of encoding/decoding that string in a test script (without explicitly setting `\xff` didn't replicate the exact error.

[...]

Issue #395, opened 2017-03-07
Problems with non-ascii characters in __str__() and __repr__() functions


gEt-rIgHt-jR:
`__str__()` , `__repr__()` functions don't work properly with non-ascii characters.

When I run this code:
```
s = "Ç"
print(s.__str__())
print(s.__repr__())
```

In Python I get:
```
Ç
'Ç'
```

But in VOC I get:
```
├âΓÇí
'├âΓÇí'
```

It is giving me problems while implementing `ascii()` function under `Python.java`

dibyadas:
@gEt-rIgHt-jR I am not getting that error. Can you try that again ? and with other non-ascii characters . 
Because I tried Ç , ¢ , ¥ ,©  and all seem to be working as expected  . Are there any other issues here?

gEt-rIgHt-jR:
@dibyadas On windows I am getting that output, on Ubuntu I am getting correct output.

PR #703, opened 2017-11-18
Work around Unicode encoding error on Windows

whydoubt:
For Python <3.6 on Windows, encoding strings for output to the Windows
command prompt may result in a UnicodeEncodingError.  VOC does not take
encoding into consideration, and so the output differs.

Until such time as proper encoding support is implemented in VOC, use an
environment variable that Python provides for overriding the IO
encoding.  By setting this to UTF-8, the output may appear garbled, but
the error is avoided, and it matches the run-as-Java output.  For
consistency, pass the environment variable to Java as well.

Addresses #610 and #237.

### DIFF
--- a/tests/utils.py
+++ b/tests/utils.py
@@ -116,7 +116,7 @@ def adjust(text, run_in_function=False):
     return '\n'.join(final_lines)
 
 
-def runAsPython(test_dir, main_code, extra_code=None, args=None):
+def runAsPython(test_dir, main_code, extra_code=None, args=None, env=None):
     """Run a block of Python code with the Python interpreter."""
     # Output source code into test directory
     with open(os.path.join(test_dir, 'test.py'), 'w', encoding='utf-8') as py_source:
@@ -143,6 +143,7 @@ def runAsPython(test_dir, main_code, extra_code=None, args=None):
         stdout=subprocess.PIPE,
         stderr=subprocess.STDOUT,
         cwd=test_dir,
+        env=env,
     )
     out = proc.communicate()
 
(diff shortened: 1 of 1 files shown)


freakboy3742:
Thanks for looking into this.

I'm intrigued what you'd consider "proper" encoding support in this context. I can't deny that Windows is definitely having difficulties here; I imagine there would be many Linux configurations that have similar problems due to odd codepage configurations.

However, in the Linux space at least, my understanding is that this is something that is considered an error of usage - somewhere between "you're doing it wrong" and "you're doing it in a way that makes it impossible to know what is right". What *is* the right approach here?

whydoubt:
For one instance, to really match what Cpython does, sys.stdout.write() should be encoding from string to stream using sys.stdout.encoding.  Typically for Linux and for Windows-with-Python>=3.6 that will be UTF-8, but it still should not be assumed.  WIth voc, sys.stdout.encoding does not even exist.  I believe this is at least part of the cause of #395, and if that's the case, that one can _only_ be solved with some semblance of encoding support.

eliasdorneles:
If I understood correctly, the issue is that VOC output is always UTF-8, and the CPython output depends on the environment.
So, this PR sets the `PYTHONIOENCODING` environment variable to tell CPython to use UTF-8 as well, regardless of the environment.
It sounds good to me.
We could make it more explicit that this is a workaround for Windows, by setting it variable only for windows, checking `sys.platform`.

@freakboy3742 any objections about merging this?

freakboy3742:
I don't have any particular objection to merging, other than wanting to have a slightly better understanding of what the "real" fix is. If it's as simple as defining `sys.platform.encoding`, and using that in system.out.write() calls... then why not just do that?

whydoubt:
There is more to it than the portion I mentioned in my previous comment.  I had thought it would require creating translation tables for each encoding we wish to support.  However, I may give a shot at using Java's encoding support to make it work.

prateek3255 (approved):
I am on a windows machine and the test_title and test_case_changes in the test_str.py file were failing earlier, but after making these changes they are working fine.

phildini:

PR #709, opened 2017-11-28
 adding method bytes.split() with tests

JvMaiia:


tests/datatypes/test_bytes.py
```diff
@@ -474,6 +474,27 @@ def test_istitle(self):
...
         """)
 
+    def test_split(self):
+        self.assertCodeExecution("""
+            print(b''.split())
+            print(b'py bee'.split())
+            print(b'pyXbXee'.split(b'X'))
+            print(b'pyXbee'.split(b'z'))
+            print(b'pyZZbee'.split(b'ZZ'))
+            print(b'pybebyp'.split(b'e', 1))
```
eliasdorneles: Could you add a test for the case: `b'one two three'.split(maxsplit=1)`

python/common/org/python/types/Bytes.java
```diff
@@ -1251,10 +1251,79 @@ public Bytes rstrip() {
...
+                }
+            }
+        }
+
+        byte[] bsep;
+        if (sep == null) {
+            bsep = " ".getBytes();
+        } else if (!(sep instanceof org.python.types.Bytes)) {
+            if (org.Python.VERSION < 0x30500f0) {
+                throw new org.python.exceptions.TypeError("'" + sep.typeName() + "' does not support the buffer interface");
```
eliasdorneles: This is optional, but there are some bits of code that are repeating, like this -- what do you think about doing all these type checks in the beginning of the method (before the check on `this.value.length`) and throw any of these errors early?
This would make the code simpler and reduce the need for some of those test cases. :)

tests/datatypes/test_bytes.py
JvMaiia: I added the test case you have suggested and the check of errors in the beginning of the method, thank for your advices!

eliasdorneles:
Parfait !
Thanks @JvMaiia ! :octopus:

PR #711, opened 2017-12-01
Fix bugs in args attribute of exception objects

onyb:
- Fix a sneaky bug introduced in #702 
- Fix string representation of `BaseException` and `KeyError` objects. Did some small cleanups.

Hat tip @eliasdorneles !

### DIFF
--- a/tests/structures/test_exception.py
+++ b/tests/structures/test_exception.py
@@ -85,7 +85,7 @@ class ExceptionTests(TranspileTestCase):
                 try:
                     raise exc("one")
                 except Exception as e:
-                    print(e, e.args)
+                    print(e, e.args, str(e), repr(e))
                 try:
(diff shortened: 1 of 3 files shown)


tests/structures/test_exception.py
```diff
@@ -85,11 +85,11 @@ def test_raising_exceptions_multiple_args(self):
                 try:
                     raise exc("one")
                 except Exception as e:
-                    print(e)
+                    print(e, e.args)
```
eliasdorneles: Maybe we could have `str(e)` and `repr(e)` also in these tests?

onyb: Done

eliasdorneles:
Thank you, @onyb ! :cake:

PR #713, opened 2017-12-06
implement dict.update method

JvMaiia:


tests/datatypes/test_dict.py
```diff
@@ -287,6 +287,46 @@ def test_fromkeys(self):
...
+            print(sorted(b))
+            c = {}
+            c.update(a=1, b=2)
+            print(sorted(c))
+        """)
+
+        self.assertCodeExecution("""
+            a = {}
+            a.update([('a', 1, 2), ('b',2)])
+        """, exits_early=True)
```
freakboy3742: You should generally avoid using `exits_early`. It's necessary because there are some tests where you *can't* can't the termination condition (for example, when you're testing `KeyboardInterrupt`); but if you can, the better pattern is to put the code you're testing in a try/catch:
```
try:
    a = {}
    a.update([('a', 1, 2), ('b', 2)])
    print("An error should have been raised!")
except ValueError:
    print("Received a ValueError as expected")
```

This allows you to check the error that is being raised, not just that "the program terminated in the same place."

JvMaiia:
Thanks for your advices @freakboy3742 ! I improved tests, removed an useless variable in the method and a test of a method that not be implemented yet!

PR #714, opened 2017-12-09
Str.java "rsplit" method

hcheng761:
Implementation of rsplit method for strings, test cases included.

### DIFF
--- a/python/common/org/python/types/Str.java
+++ b/python/common/org/python/types/Str.java
@@ -1404,7 +1404,7 @@ public class Str extends org.python.types.Object {
                         count++;
                     }
                 }
-                if (maxsplit != null && java.lang.Integer.parseInt(maxsplit.toString()) >= 0) {
+                if (maxsplit != null && java.lang.Integer.parseInt(maxsplit.toString()) >= 0) {	//check for positive maxsplit
                     number = java.lang.Integer.parseInt(maxsplit.toString());
                 } else {
                     number = count;
--- a/python/common/org/python/types/Str.java
+++ b/python/common/org/python/types/Str.java
@@ -1454,8 +1454,8 @@ public class Str extends org.python.types.Object {
                             } else {
                                 result_list.insert(new org.python.types.Int(0), new org.python.types.Str(value.substring(j+sepStr.length(),temp)));
                                 temp = j;
-                           }
-                           break;
+                            }
+                            break;
                         }
                     }
                 }
--- a/python/common/org/python/types/Str.java
+++ b/python/common/org/python/types/Str.java
@@ -1346,8 +1346,115 @@ public class Str extends org.python.types.Object {
                     "splits are done. If sep is not specified, any whitespace string\n" +
                     "is a separator.\n"
     )
-    public org.python.Object rsplit() {
-        throw new org.python.exceptions.NotImplementedError("rsplit() has not been implemented.");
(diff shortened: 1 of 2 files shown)


freakboy3742:
@hcheng761 You can run these lint checks locally so you can put them all into one commit, instead of removing once trailing space per commit. Run `ant checkstyle` to get all the problems in one report.

python/common/org/python/types/Str.java
```diff
@@ -1344,10 +1344,124 @@ private long toPositiveIndex(long index) {
...
                     "working to the front.  If maxsplit is given, at most maxsplit\n" +
                     "splits are done. If sep is not specified, any whitespace string\n" +
-                    "is a separator.\n"
+                    "is a separator.\n",
+            default_args = {"sep", "maxsplit"}
     )
-    public org.python.Object rsplit() {
-        throw new org.python.exceptions.NotImplementedError("rsplit() has not been implemented.");
+    public org.python.Object rsplit(org.python.Object sep, org.python.Object maxsplit) {
+        if (this.value.isEmpty()) { //handles empty strings
```
eliasdorneles: If I'm not missing anything,one of the first things we could do here is to check for the case of `maxsplit` not provided (i.e., if it is == null), and then just return the result of calling `split(sep, null)`.

hcheng761: @eliasdorneles Could you elaborate? From what I have written so far, I check for cases in which maxsplit is not provided, regardless of the value of sep.

eliasdorneles: Hey @hcheng761, so this was just an idea to simplify the logic, because the behavior of `maxsplit == null` is the same as for `split()`, so i was thinking we could short-circuit that.
Makes sense?

hcheng761: @eliasdorneles I made some recent changes according to your idea. It should be only using the split() method for a null maxsplit value now.

eliasdorneles: Sorry, I think i was not clear.

The idea was to exploit the fact that `str.rsplit` behaves the same as `str.split` if `maxsplit` is not provided.

The idea was to do this, at the beginning of the method:

```
if (maxsplit == null) {
    return this.split(sep, null);  // delegate to split
}
// in here, implement rsplit for the case when maxsplit is not null
```

hcheng761: I think I understand. I reverted the code and call the split implementation in the same file, is this correct?

tests/datatypes/test_str.py
```diff
@@ -713,6 +713,20 @@ def __str__(self):
...
             """)
+    def test_rsplit(self):
+        self.assertCodeExecution(r"""
+           for s in ['This is for rsplit', ' eeed d ded  eded   ', 'átomo', '',' t e s t i n g   r s p l i t ', 'a b c']:
+                print(s.rsplit())
+                print(s.rsplit("e"))
+                #print(s.rsplit(maxsplit=2))
+                #print(s.rsplit(maxsplit=-5))
+                #print(s.rsplit("e",-9))
+                #print(s.rsplit(" ", -10))
```
eliasdorneles: Maybe you want to uncomment these tests now?

eliasdorneles:
Thanks for contributing, @hcheng761 ! :+1:

PR #718, opened 2018-01-01
Refs #36

Rushane:
Implement test_multiply_str under test_float.py


eliasdorneles:
Hello @Rushane, thanks for the interest in contributing.

I'm afraid this change is not acceptable, however, because 1) it's adding a copy of a virtualenv directory, 2) the `test_multiply_str` test is already generated dynamically, there is no need to write it.

Note that a test name under the list `not_implemented` means that the code to handle that particular case (in this case, multiplying a string by a float) wasn't yet implemented, but it doesn't mean that the test itself isn't implemented -- it's generated and run by the test suite.

Rushane:
Thanks for replying.

So, how I would know if a test is run by the test suite? is it that all of the tests that are supposed to generate errors are run by the test suite?

Issue #538, opened 2017-05-22
Replace `ant` with `gradle` for build.


phildini:
Chatting with @freakboy3742 and @eliasdorneles in the run-up to Pycon sprints, the consensus seems to be that we can replace `ant` with `gradle` for VoC building.

This would be one less dependency in the stack, and is a Good Thing™.

eliasdorneles:
An acceptable initial implementation of this may be too add a `build.gradle` that just calls the ant tasks, this way we can just tell people to use Gradle and gradually migrate away of Ant.

The ideal would be to get rid of Ant altogether.

eliasdorneles:
Gradle docs on calling ant tasks: https://docs.gradle.org/current/userguide/ant.html

This should work without having to install ant manually.

drurenia:
This would be my first contribution. Can I grab this one?

eliasdorneles:
Sure, go ahead. :)
Note that this is the kind of thing that will be good for you to gather feedback early in the process.
So don't hesitate in sharing your work when you think you're 30% done, so that we can try it out and discuss it.

drurenia:
Great! I’ll do it no problem. When ready should I share it via PR?

eliasdorneles:
Yes. :)

Le 29 déc. 2017 11:52 PM, "Diego Urenia" <notifications@github.com> a
écrit :

PR #719, opened 2018-01-01
Partial migration from Ant to Gradle

drurenia:
- I started with the easy tasks and migrated few tasks to Gradle: all, upload, checkstyle, clean;
- Added a test task as well to make it easier to execute it (not working because it lacks the venv as specified below);
- Upload and test (newly added) tasks are not working for now because a virtual env is needed to execute them. I haven't figured out yet how to achieve this on Gradle so any suggestion would be great;
- Added Gradle wrapper to eliminate the need for the installation; 
- Importing all remaining tasks from Ant automatically using ant.importBuild "build.xml", so everything still works fine using both ant and gradle commands;

PS: Nothing was added regarding the Exec tasks to address multiple OS's because the Exec tasks are still not working propertly as mentioned above.

Addresses #538


phildini:
Hi there! Sorry for the long delay in feedaback. What would you suggest is the best way to test this? Can these be run locally easily?

drurenia:
Not a problem @phildini. Since we have a Gradle wrapper in place, you could easily test it using:
`./gradlew <task_name>`

Please bear in mind that I opened this pull request because @eliasdorneles asked me to but I don't think it is supposed so be merged yet. I think since it's my first contribution to this project he wanted to check whether I am going to the right direction or not.

build.gradle
```diff
@@ -0,0 +1,56 @@
...
+    workingDir = "tools"
+    commandLine = ["python", "upload.py"]
+}
+
+task clean(type: Delete) {
+    description = "Clean up"
+    delete = ["build", "dist"]
+}
+
+task test(type: Exec) {
```
eliasdorneles: This target should be removed: besides the fact that this is hardcoding a virtualenv path, the gradle build config should concern only the Java resources, not the python stuff.

build.xml
```diff
@@ -43,17 +31,4 @@
     <ant antfile="build-android.xml" target="dist"/>
   </target>
 
-  <target name="all" depends="java, android" />
-
-  <target name="upload" depends="java, android" description="Upload assets to S3">
-    <exec executable="python">
-      <arg value="tools/upload.py"/>
-      <arg value="b${release}"/>
```
eliasdorneles: This argument seems to be missing in the Gradle version.

eliasdorneles (changes requested):
Hi @drurenia !
Sorry for the looong time to review -- I've been eyeing this PR for a while, but life happened! :)
So, I think this is a good start, but there is a few more things I'd hope to see before this is ready to be included -- ideally, at least we'd be able to do `./gradlew build` and it would compile all the Java stuff with Gradle.
I've made some comments below on some things that are not quite right.
How big of a task do you think it would be to move from Ant to Gradle completely?
Thank you!!

phildini:

PR #720, opened 2018-01-08
ComplexTests - test_eq_complex - Work in Progress

Rushane:
#36 

I'm getting an error message in reference to tests/utils.py about float._mul_ not being implemented:
 File "/root/voc-dev/voc/tests/utils.py", line 399, in assertCodeExecution
AssertionError: '### EXCEPTION ###\nNotImplementedError: f[53 chars]:2\n' != 'True\nTrue\n===end of test===\n'
- ### EXCEPTION ###
- NotImplementedError: float.__mul__() has not been implemented.
-     test.py:2
  
  I'm wondering if I would have to implement this float.__mul__() in  tests/utils.py. Is there any documentation that would help me achieve this?


cflee:
@Rushane if you could enclose the code segments with triple quotes to format it as a code block, it would make the output easier to read! It would look something like this instead:

```
File "/root/voc-dev/voc/tests/utils.py", line 399, in assertCodeExecution
AssertionError: '### EXCEPTION ###\nNotImplementedError: f[53 chars]:2\n' != 'True\nTrue\n===end of test===\n'
- ### EXCEPTION ###
- NotImplementedError: float.__mul__() has not been implemented.
-     test.py:2
```

> I'm getting an error message in reference to tests/utils.py about float.mul not being implemented:

First, it seems that this `float.__mul__()` error is due to the `y = x.real+x.imag*1j` line in your test – the real and imag components of the Complex are actually a Float, so `x.imag * 1j` is the multiplication that is failing. Hence, this test is actually kind of failing for the wrong reasons – it's not just testing `x == y` or `x.__eq__(y)`, it's also testing constructing Complex using `Float + Float * 1j` format!

Fixing that could be an interesting task. The relevant implementation is in `python/common/org/python/types/Float.java`, see the `__mul__()` method. If you want to get that specific test working (though renamed to something other than `test_eq_complex`), you'll need to get this method to handle multiplying a Float with a `1j` and returning a Complex.

--

Second, the other thing is that the `test_eq_complex` test in `InplaceComplexOperationTests` that you uncommented in the `not_implemented` array is automatically generated by `tests/utils.py`. If you just wanted to get that sort of generated test to pass, you should just be uncommenting it and then running the test to see how it fails, before you go and fix it / add more tests.

**However,** the weird thing is that if you go over to the InplaceOperationTests in `tests/utils.py`, there are no `test_eq_<datatype>` tests, _because there's no inplace 'equality' operator_! It doesn't make sense to exist, unlike `+= -= *= /=` and other inplace operators. So there's actually a problem with the InplaceComplexOperationTests `not_implemented` list. If you want to fix that list to correspond to the InplaceOperationTests-generated tests that really exist and are failing, that would be great too!

phildini:

PR #721, opened 2018-01-11
updated str.__rmul__() and tested it 

lumingc:


cflee:
@lumingc We're currently not testing `.__rmul__()` and similar 'dunder' methods directly. Instead we've been focused on getting the direct operators working correctly, as those are more frequently used.

While we _would_ like to get to having everything working exactly the same, it's quite tricky to get everything to work identically to CPython – please see related PRs #583, #598, and some discussion in less related PRs #687 #651. I think fixing this will require some work in VOC arg handling internals, not just on individual datatypes.

phildini:

Issue #554, opened 2017-05-23
Round misimplemented for integers


gkhaburzaniya:
round(15, -1) gives 15

Should give 20

eliasdorneles:
Yeah, currently rounding integers are just returning the same value unmodified: https://github.com/pybee/voc/blob/master/python/common/org/python/types/Int.java#L757

According [to the docs:](https://docs.python.org/3/library/functions.html#round)

> For the built-in types supporting round(), values are rounded to the closest multiple of 10 to the power minus ndigits

shreyateeza:
Can anyone please tell me how does this round function works so that I can help in correcting this bug.

eliasdorneles:
@shreyateeza check the documentation for rounding in Python 3: https://docs.python.org/3/library/functions.html#round

What we're missing is the implementation for rounding integers, which would go in here: https://github.com/pybee/voc/blob/cad52e31d005e9c1139daf5c378d79ab4ba0a61e/python/common/org/python/types/Int.java#L867

Instead of just returning another Int object with the same value, it should round the value before, as per specified in the docs (see previous comment).

hogesonline:
Hi, I think this issue is under-stated because "round" seems to only return the integer value at all because that's all math.round does in Java.

uranusjr:
@hogesonline VOC’s round implementation doesn’t have to have anything to do with Java’s, and the current implementation doesn’t use `Math.round` either. It’s just not implemented correctly.

abhayptp:
#724 I have implemented it here.

PR #724, opened 2018-01-17
Round function implemented for integers

abhayptp:


uranusjr:
@abhayptp Oops, I didn’t realise this existed until reading you comment on #554, and [just implemented the same thing](https://github.com/uranusjr/voc/commit/d636aac330e72afe86b442a1d00e1963c5415440). This seems a bit unnecessarily complicated though; mine is based on CPython’s implementation, and AFAICT works well (with `long`s; the current implementation can’t do big integers like Python’s).

abhayptp:
Oh! I removed that doc by mistake. @uranusjr can go ahead with his implementation as that is standard implementation. I am closing this.

Issue #414, opened 2017-03-13
Format of TypeError message is wrong for missing arguments in some functions.


deep110:
For example, when I was implementing `dict.fromkeys()` and checked with no args, this was result.
>fromkeys() missing 1 required positional argument: 'iterable' --> getting in voc
fromkeys expected at least 1 arguments, got 0 --> expected

This is an issue in many functions beacuse as @cflee pointed out in https://github.com/pybee/voc/pull/391 this error message is triggered by cpython's [getargs.c unpack_stack()](https://github.com/python/cpython/blob/6f0eb93183519024cb360162bdd81b9faec97ba6/Python/getargs.c#L2365) and can be triggered elsewhere too, not just from Dict objects.

cflee:
I'm not entirely sure if it's related yet, but I just spotted some similar in `str.rfind()`:

```
Observed output:
  rfind() missing 1 required positional argument: 'item'
Expected output:
  rfind() takes at least 1 argument (0 given)
```

PR #727, opened 2018-02-13
Add str.zfill()

SanketDG:
Hi, I have added `str.zfill()` to the builtin `str` type but I can't seem to figure out why my tests are failing.

Here is my log:
```
FAIL: test_zfill (tests.datatypes.test_str.StrTests)
  File "/home/sanketdg/projects/voc/tests/datatypes/test_str.py", line 785, in test_zfill
    """)
  File "/home/sanketdg/projects/voc/tests/utils.py", line 399, in assertCodeExecution
AssertionError: '### EXCEPTION ###\nTypeError: zfill() tak[54 chars]:2\n' != '00042\n===end of test===\n'
+ 00042
+ ===end of test===
- ### EXCEPTION ###
- TypeError: zfill() takes 0 positional arguments but 1 was given
-     test.py:2
 : Global context

Ran 1 test in 21.531s
```

### DIFF
--- a/python/common/org/python/types/Str.java
+++ b/python/common/org/python/types/Str.java
@@ -1764,8 +1764,47 @@ public class Str extends org.python.types.Object {
     public org.python.Object upper() {
         return new org.python.types.Str(this.value.toUpperCase());
     }
-}
 
+    @org.python.Method(
+            __doc__ = "S.zfill() -> str\n" +
+                    "\n" +
+                    "Return a copy of the string left filled with ASCII '0' \n" +
+                    "digits to make a string of length width.\n",
+            args = {"width"}
+    )
+    public org.python.Object zfill(org.python.Object width) {
+
+
+        if (width instanceof org.python.types.Float) {
+            throw new org.python.exceptions.TypeError("integer argument expected, got float");
+        }
+        else if (!(width instanceof org.python.types.Int)) {
+            throw new org.python.exceptions.TypeError("'" + org.Python.typeName(width.getClass()) +
+                                                      "' object cannot be interpreted as an integer");
+        }
+
+        int w = (int) ((org.python.types.Int) width).value;
+
+        if (this.value.length() >= w) {
+            return new org.python.types.Str(this.value);
+        }
+
+        int fill = w - this.value.length();
+
+        java.lang.StringBuffer str = new java.lang.StringBuffer(w);
+
+        if (this.value.length() != 0 && (this.value.charAt(0) == '-' || this.value.charAt(0) == '+')) {
+            str.append(this.value.charAt(0));
+            this.value = this.value.substring(1);
+        }
+
+        for(int i = 0; i < fill; i++) {
+            str.append('0');
+        }
+
+        return new org.python.types.Str(str.toString() + this.value);
+    }
+}
 
 final class PythonFormatter {
 
--- a/tests/datatypes/test_str.py
+++ b/tests/datatypes/test_str.py
@@ -778,6 +778,40 @@ class StrTests(TranspileTestCase):
         print(s1.splitlines(True))
         """)
 
+    def test_zfill(self):
+        self.assertCodeExecution("""
+            s = '42'
+            print(s.zfill(5))
+
+            try:
+                print(s.zfill('string'))
+            except TypeError as err:
+                print(err)
+
+            try:
+                print(s.zfill({}))
+            except TypeError as err:
+                print(err)
+
+            s = '-42'
+            print(s.zfill(5))
+
+            s = '+42'
+            print(s.zfill(5))
+
+            s = ''
+            print(s.zfill(5))
+
+            s = '-.-42'
+            print(s.zfill(6))
+        """)
+
+    @expectedFailure
+    def test_zfill_arg_handling():
+        self.assertCodeExecution("""
+            s = "42"
+            s.zfill()
+        """)
 
 class UnaryStrOperationTests(UnaryOperationTestCase, TranspileTestCase):
     data_type = 'str'


python/common/org/python/types/Str.java
```diff
@@ -1764,8 +1764,36 @@ public static String _title(String input) {
     public org.python.Object upper() {
         return new org.python.types.Str(this.value.toUpperCase());
     }
-}
 
+    @org.python.Method(
```
eliasdorneles: You need to declare the argument with this `@org.python.Method` annotation -- this maps the Python method signatures to Java arguments.
Check here for examples: http://voc.readthedocs.io/en/latest/reference/signatures.html

SanketDG: Oops, don't know how I missed that. Updated the PR.

eliasdorneles: Using `default_args` is still not the answer, a default argument means that the argument is optional. You want to use `args` instead.
You also need to test the argument handling -- that is, if it's raising the correct error when given invalid arguments.

tests/datatypes/test_str.py
```diff
@@ -778,6 +778,17 @@ def test_splitlines(self):
...
+    def test_zfill(self):
+        self.assertCodeExecution("""
+            s = '42'
+            print(s.zfill(5))
+        """)
+
+    def test_zfill_negative(self):
+        self.assertCodeExecution("""
+            s = '-42'
+            print(s.zfill(5))
```
eliasdorneles: This test is not doing anything of value, the "negative" thing here is a string and not a number, the `zfill` method is getting the same argument as in the other test.

SanketDG: Sorry I dont follow. `zfill` just checks if the first char is a `-` or a `+`, and then inserts it at the first, it doesn't matter if its a number or a string.

eliasdorneles: Ooops, my apologies, you're right!
Indeed, it seems the missing tests are for the empty string, a positive sign and something like `'-.2'.zfill(5)`

uranusjr: Might also want to include test cases from CPython https://github.com/python/cpython/blob/master/Lib/test/string_tests.py#L830

SanketDG: Wow almost got that bug in without the empty string test case, :) I have updated the tests.

tests/datatypes/test_str.py
```diff
@@ -778,6 +778,43 @@ def test_splitlines(self):
...
+        self.assertCodeExecution("""
+            s = '42'
+            print(s.zfill(5))
+        """)
+
+    @expectedFailure
+    def test_zfill_missing_arg(self):
+        self.assertCodeExecution("""
+            s = '42'
+            print(s.zfill())
```
eliasdorneles: This should not be an `@expectedFailure` (as this would allow any kind of failure), but instead it should just catch the right exception and print the exception -- the output should be the same as CPython.
Look at the tests for `rpartition` to see examples of this.

tests/datatypes/test_str.py
eliasdorneles: Cool, thanks, this is getting good! :+1: 
May I ask you to group these happy-path tests in the same assertion in same test case?
While in many projects it's good to have separate cases, in VOC each separate assertion requires compiling the code given as argument and run it separately in Java and CPython -- so putting several related cases under the same test saves some resources.

SanketDG:
> so putting several related cases under the same test saves some resources.

Makes sense! I have done that, but revisiting test files, I have seen many tests are having different functions, I am assuming they need to be fixed? For example:

https://github.com/pybee/voc/blob/master/tests/datatypes/test_str.py#L346

> Look at the tests for rpartition to see examples of this.

I got around to doing this, but I got the following error message:

```
- zfill() missing 1 required positional argument: 'width'
+ zfill() takes exactly 1 argument (0 given)
```

I was not sure what's causing the error, so I tried emulating other functions with similar signatures, like `s.startswith()` (no test exists for empty argument), and I got the same message above. Does this have something to do with how VOC handles arguments, or am I missing something trivial?

eliasdorneles:
> I have seen many tests are having different functions, I am assuming they need to be fixed?

Those could be refactored, yes.

> I was not sure what's causing the error, so I tried emulating other functions with similar signatures, like s.startswith() (no test exists for empty argument), and I got the same message above. Does this have something to do with how VOC handles arguments, or am I missing something trivial?

Hm yeah, it seems we're hitting a little problem on the way VOC handles function arguments in [Function.checkMissingArgs](https://github.com/pybee/voc/blob/master/python/common/org/python/types/Function.java#L182).
I think this different argument handling happens in CPython for functions that are implemented in C like `str.zfill`, the argument handling seems to be different for those cases -- we'll need to do something different for this kind of stuff.

So maybe for this PR you can move that failing test for the argument handling into a `test_zfill_arg_handling` with `@expectedFailure`, and then we can fix it later when we have a better approach for it.
Does it make sense?

Thanks!

SanketDG:
> I think this different argument handling happens in CPython for functions that are implemented in C like str.zfill, the argument handling seems to be different for those cases -- we'll need to do something different for this kind of stuff.

Do we have a tracking issue for this?

> So maybe for this PR you can move that failing test for the argument handling into a test_zfill_arg_handling with @expectedFailure, and then we can fix it later when we have a better approach for it.

Updated, thanks!

cflee:
> Do we have a tracking issue for this?

I think this is related to #414, though not the specific linked `unpack_stack()` method in CPython, probably `vgetargs1_impl()` instead... that issue description should be general enough, but would be great to add a comment there to link/explain this. The last time we saw this we decided to do the same separate-and-mark-expectedFailure, so that's great, thanks!

eliasdorneles:
Thanks, this looks good!
If I may ask one final request, we also need to throw `TypeError` for the case of wrong type.

In CPython:

```
>>> '42'.zfill('5')
TypeError: 'str' object cannot be interpreted as an integer
```

With the code from this PR:

```
>>> '42'.zfill('5')
Exception in thread "main" RuntimeError: org.python.types.Str cannot be cast to org.python.types.Int
	at python._tempfile.module$import(_tempfile.py:2)
	at python._tempfile.main(_tempfile.py)
```

python/common/org/python/types/Str.java
```diff
@@ -1764,8 +1764,42 @@ public static String _title(String input) {
...
+            __doc__ = "S.zfill() -> str\n" +
+                    "\n" +
+                    "Return a copy of the string left filled with ASCII '0' \n" +
+                    "digits to make a string of length width.\n",
+            args = {"width"}
+    )
+    public org.python.Object zfill(org.python.Object width) {
+
+        if (width == null) {
+            throw new org.python.exceptions.TypeError("zfill() takes exactly 1 argument (0 given)");
```
eliasdorneles: you can remove this error handling too, `width` will never be null with the declaration of arguments as `args` (it might be `None` though, which would still be a org.python.Object)

python/common/org/python/types/Str.java
```diff
@@ -1764,8 +1764,42 @@ public static String _title(String input) {
...
+        }
+
+        int w = (int) ((org.python.types.Int) width).value;
```
eliasdorneles: This is the place where we need the error handling, for the case `width` is not an integer.
You can see similar error handling here: https://github.com/pybee/voc/blob/master/python/common/org/python/types/ByteArray.java#L89-L91

SanketDG:
> If I may ask one final request, we also need to throw TypeError for the case of wrong type.

This was different for float and other types, so I handled them differently. Updated, thanks!

eliasdorneles:
Parfait !
Thanks @SanketDG ! :100:

PR #728, opened 2018-02-17
Dict keys

shakram02:
Hello, It's my first time here. I got interest in the project from GSoC's page so I'm trying to take my first steps.

I wanted to implement `keys()` method. This is a work in progress as I still need some guidance.
I should return a `dict_keys` object. How should I do that? the Java way

Input:
```Python
x = {}
x['1'] = "no"
x[5] = 7
x["iot"] = "2020"
print(x.keys())
```

Output:
```
AssertionError: "['1', 5, 'iot']\n===end of test===\n" != "dict_keys(['1', 5, 'iot'])\n===end of test===\n"
- ['1', 5, 'iot']
+ dict_keys(['1', 5, 'iot'])
  ===end of test===
 : Global context
```


jvmaia:
The methods dict.keys, dict.items, dict.values, returns a dict_view object that uses a MappingProxyType to create a read-only view. I guess you'll need to create first the MappingProxyType.
https://docs.python.org/3/library/stdtypes.html#dict-views

shakram02:
I'll give it a spin, thank you so much

eliasdorneles:
Hi there!
This is not needed anymore, as it was implemented in: https://github.com/pybee/voc/pull/786
Thanks for contributing and getting the ball rolling!

PR #729, opened 2018-02-18
Bytearray expandtabs

shakram02:
Implementation of `bytearray.expandtabs()`

Any missing test cases?
Do I need to check for overflows anywhere?

One more question, I should've used the already implemented `expandtabs()` in `Str`? like it's the same concept, but just for different types, I'm thinking about creating some utility classes for stuff like `iterable` or some base abstract type to have the implementation of those common methods between iterables instead of implementing them multiple times in the codebase ?


python/common/org/python/types/Bytes.java
```diff
@@ -971,10 +972,70 @@ public boolean __setattr_null(java.lang.String name, org.python.Object value) {
...
+
+        int tabsize_int = 8;
+
+        if (tabsize != null) {
+            tabsize_int = (int) ((org.python.types.Int) tabsize).value;
+        }
+
+
+        int output_size = 0;
+        for (int i = 0, j = 0; i < this.value.length; i++) {
```
eliasdorneles: I don't understand what this for loop is accomplishing, could you clarify?

It _looks_ like it's meant to compute the output_size, but the code below doesn't seem to be using it in a meaningful way.

python/common/org/python/types/Bytes.java
```diff
@@ -971,10 +972,70 @@ public boolean __setattr_null(java.lang.String name, org.python.Object value) {
...
+                j++;
+
+                if (c == '\n' || c == '\r') {
+                    output_size += j;
+                    j = 0;
+                }
+            }
+        }
+
+        StringBuilder sb = new StringBuilder();
```
eliasdorneles: We should probably not use StringBuilder here -- a Bytes object contains only bytes (8 bits) and is backed by a `byte[]` (Java byte array), while a String contain Unicode codepoints (by default using UTF-16 charset, which means double the usage of memory).

You should probably just use an ArrayList<Byte>, initializing it with capacity of `this.value.length`, let it expand to the right size and at the end convert it back to an array of `byte` primitives.

eliasdorneles (changes requested):
Hi, thanks for contributing, I've made some comments below.

python/common/org/python/types/Bytes.java
shakram02: ah yes, I forgot to change the variable name, I didn't fully understand it in the beginning, then I went through the documentation. It adjusts the tab sizes using columns, details [here](https://docs.python.org/3/library/stdtypes.html#str.expandtabs)

python/common/org/python/types/Bytes.java
shakram02: Thanks for the information, Will do

python/common/org/python/types/Bytes.java
shakram02: ah, I'm sorry about that I've forgotten it from an older version of the code, the C implementation first calculates the size of the output then makes a buffer accordingly, we don't need that now.

python/common/org/python/types/Bytes.java
shakram02: Thanks for the information, I updated it +1

python/common/org/python/types/Bytes.java
```diff
@@ -971,10 +972,56 @@ public boolean __setattr_null(java.lang.String name, org.python.Object value) {
...
 
     @org.python.Method(
-            __doc__ = "B.expandtabs(tabsize=8) -> copy of B\n\nReturn a copy of B where all tab characters are expanded using spaces.\nIf tabsize is not given, a tab size of 8 characters is assumed."
+            __doc__ = "B.expandtabs(tabsize=8) -> copy of B\n\nReturn a copy of B where all tab characters are expanded using spaces.\nIf tabsize is not given, a tab size of 8 characters is assumed.",
+            default_args = {"tabsize"}
     )
-    public org.python.Object expandtabs(java.util.List<org.python.Object> args, java.util.Map<java.lang.String, org.python.Object> kwargs, java.util.List<org.python.Object> default_args, java.util.Map<java.lang.String, org.python.Object> default_kwargs) {
-        throw new org.python.exceptions.NotImplementedError("bytes.expandtabs has not been implemented.");
+    public org.python.Object expandtabs(org.python.types.Object tabsize) {
+        // CPython implementation https://github.com/python/cpython/blob/master/Objects/stringlib/transmogrify.h#L21
```
cflee: Just a note that this link's `L21` line number will become outdated as the `master` branch changes – it will be better to link it to a tag (`v3.6.4`) or a specific commit, either of which won't change.

shakram02: That's a very neat suggestion -never thought about it this way before-, Thanks! I'll fix it

python/common/org/python/types/Bytes.java
```diff
@@ -971,10 +972,56 @@ public boolean __setattr_null(java.lang.String name, org.python.Object value) {
...
+        }
+
+        if (this.value.length == 0) {
+            return new org.python.types.Bytes(new byte[0]);
+        }
+
+        int tabsizeInt = 8;
+
+        if (tabsize != null) {
+            tabsizeInt = (int) ((org.python.types.Int) tabsize).value;
```
eliasdorneles: You should add a test for a wrong argument (for example, `b''.expandtabs(" ")` and `b''.expandtabs(None)`), and throw the correct exception and error message here in case tabsize is invalid.

python/common/org/python/types/Bytes.java
```diff
@@ -971,10 +972,56 @@ public boolean __setattr_null(java.lang.String name, org.python.Object value) {
...
+        }
+
+        ArrayList<Character> buffer = new ArrayList<>();
```
eliasdorneles: This should probably be of type `Byte` instead of `Character` (one Character uses two bytes in UTF-16 -- the default Java charset).
You could also pass an initial capacity here to `ArrayList<>(this.values.length)` to reduce the number of reallocations needed.

python/common/org/python/types/Bytes.java
```diff
@@ -971,10 +972,56 @@ public boolean __setattr_null(java.lang.String name, org.python.Object value) {
...
+                if (c == '\n' || c == '\r') {
+                    lineIndex = 0;
+                }
+            }
+        }
+
+        byte[] bytes = new byte[buffer.size()];
+        for (int i = 0; i < buffer.size(); i++) {
+            bytes[i] = (byte) ((char) buffer.get(i));
+        }
```
eliasdorneles: This conversion from `ArrayList<Byte>` to `byte[]` would probably be useful as a `listToArray` method, to be reused by other code.

python/common/org/python/types/Bytes.java
```diff
@@ -971,10 +972,56 @@ public boolean __setattr_null(java.lang.String name, org.python.Object value) {
...
     @org.python.Method(
-            __doc__ = "B.expandtabs(tabsize=8) -> copy of B\n\nReturn a copy of B where all tab characters are expanded using spaces.\nIf tabsize is not given, a tab size of 8 characters is assumed."
+            __doc__ = "B.expandtabs(tabsize=8) -> copy of B\n\nReturn a copy of B where all tab characters are expanded using spaces.\nIf tabsize is not given, a tab size of 8 characters is assumed.",
+            default_args = {"tabsize"}
     )
-    public org.python.Object expandtabs(java.util.List<org.python.Object> args, java.util.Map<java.lang.String, org.python.Object> kwargs, java.util.List<org.python.Object> default_args, java.util.Map<java.lang.String, org.python.Object> default_kwargs) {
-        throw new org.python.exceptions.NotImplementedError("bytes.expandtabs has not been implemented.");
+    public org.python.Object expandtabs(org.python.types.Object tabsize) {
+        // CPython implementation https://github.com/python/cpython/blob/3.6/Objects/stringlib/transmogrify.h#L21
+        // TODO: check overflows
```
eliasdorneles: What exactly is needed checking here?

eliasdorneles (changes requested):
Hello!
Thanks for contributing, I think this is overall in the right direction -- I've made some comments on some things I noticed below.

python/common/org/python/types/Bytes.java
shakram02: in the C implementation, it checks for integer overflows. I'm not sure if we should check it here

python/common/org/python/types/Bytes.java
shakram02: Where should I place that new method?

eliasdorneles: You can start by just creating a public static method in this same class.
Another option is to create a `BytesUtils.java` file for adding static methods like these, since they are probably going to be used in `ByteArray.java` as well.

python/common/org/python/types/Bytes.java
eliasdorneles: Well, tbh I'm having trouble to follow the logic in this implementation. Maybe you got too much inspiration from the C implementation? =)
It's a bit strange that this is so different from the one for Str, have you checked that one? Why is this so much more involved, is this handling some cases which the other doesn't?

phildini:

PR #730, opened 2018-02-18
Implement startswith and endswith for ByteArray

dgrcode:
I've implemented it using a `Byte` instance. What do you think?


uranusjr:
I’d think this can be better implemented with inheritance to composition. My preference is composition; move `endsOrStartsWith` to an internal `ByteUtils` class (or similar) as a static method, and call it in both `Bytes` and `ByteArray`.

dgrcode:
I agree. I'll work on it when I find the time.

Thanks for the feedback!

PR #733, opened 2018-02-24
Modified Str.isCharPrintable to check for Unicode category

BPYap:


uranusjr:
Is there a reason why the additional checks are not added in `isCharPrintable`, but after it?

BPYap:
My bad for mixing up `IsCharPrintable` and `isprintable`, the original commits titile should be "Modified Str.isprintable to check for Unicode category".

I looked into the implementation of `IsCharPrintable` and I think the checking can be replaced with the built-in `Character.getType` method. 

I passed the` test_str` test with the new implementation, however when I added a high-surrogate value '\uD840' into the test cases, I get the following error: `UnicodeEncodeError: 'utf-8' codec can't encode character '\ud840' in position 0: surrogates not allowed`, while it should return false when `'\uD840'.isprintable()` is executed in Python 3.6. Should I open a new ticket regarding this issue?

cflee:
Thanks! For future reference, the high-surrogate issue was filed at #757.

PR #736, opened 2018-02-25
Implementation of test-power-complex for Float.java

Deepayan-Ghosh:
Implementation of a floating number raised to a complex power, corresponding to the test ```test_power_complex``` in ```test_float.py``` for type ```Float.java```.

### DIFF
--- a/python/common/org/python/types/Float.java
+++ b/python/common/org/python/types/Float.java
@@ -313,6 +313,8 @@ public class Float extends org.python.types.Object {
             return new org.python.types.Float(((double) this.value) * ((org.python.types.Float) other).value);
         } else if (other instanceof org.python.types.Bool) {
             return new org.python.types.Float(this.value * (((org.python.types.Bool) other).value ? 1 : 0));
+        } else if (other instanceof org.python.types.Complex) {
+            return ((org.python.types.Complex) other).__mul__(this);
         } else if (other instanceof org.python.types.Dict) {
             throw new org.python.exceptions.TypeError("unsupported operand type(s) for *: 'float' and '" + other.typeName() + "'");
         } else if (other instanceof org.python.types.NoneType) {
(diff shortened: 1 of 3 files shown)


Deepayan-Ghosh:
I do not how the build failed, I ran the test again on my laptop using ```python setup.py test -s tests.datatypes.test_float``` it is not showing any errors. Am I doing something wrong with the testing?

eliasdorneles:
Hm, looks like it's failing for this case:

```
E     >>> x = 2.3456
E     >>> y = 3.14159265j
E     >>> x **= y
E     >>> x
E   - (-0.8946025309573877+0.44686274358536093j)
E   ?                                      ^^^
E   + (-0.8946025309573877+0.446862743585361j)
E   ?  
```

Maybe it's a matter of precision?
This will need some more digging...

cflee:
If it's purely a rounding issue, can add to the `substitutions`, like:

https://github.com/pybee/voc/blob/da6dca3efe5501427c358a87a9166ce886050f94/tests/datatypes/test_complex.py#L50-L60

Deepayan-Ghosh:
@cflee On adding the following lines to ```test_float.py```:
```
class BinaryFloatOperationTests(BinaryOperationTestCase, TranspileTestCase):
    data_type = 'float'

    substitutions = {
        "(-0.8946025309573877+0.446862743585361j)": [
            "(-0.8946025309573877+0.44686274358536093j)"
        ]
    }
    ...........................
```
the test ```test_float.BinaryFloatOperationTests.test_modulo_int``` seems to fail. The error seems to lie here:
```
  >>> x = -3.14159
  >>> y = -5
  >>> x % y
- -3.1415900000000008
+ -3.14159
```
Should I add the substitution to ```test_complex.py``` rather than ```test_float.py``` but we are testing float operation, right?

cflee:
Look inside the `tests/utils.py` file, there are existing substitutions including that one for pi. Those lines are _replacing_ the entire substitutions dict, but you can just copy/add to the substitutions instead. There is at least one other place in the test suite which does that, if you get your text editor to search for it.

Deepayan-Ghosh:
@cflee Done. Checks passed.

Deepayan-Ghosh:
@eliasdorneles Can you please review this PR? Thanks. :smile:

PR #737, opened 2018-02-26
This fixes test_modulo_complex in test_float.java

student-06:
#36 this operation float%complex now gives appropriate error.

### DIFF
--- a/python/common/org/python/types/Float.java
+++ b/python/common/org/python/types/Float.java
@@ -355,7 +355,9 @@ public class Float extends org.python.types.Object {
             } else {
                 throw new org.python.exceptions.ZeroDivisionError("float division by zero");
             }
-        }
+        }// else if (other instanceof org.python.types.Complex){
+           // return other.__truediv__(this);
+        //}
         throw new org.python.exceptions.TypeError("unsupported operand type(s) for /: 'float' and '" + other.typeName() + "'");
     }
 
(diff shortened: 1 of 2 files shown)


python/common/org/python/types/Float.java
```diff
@@ -355,7 +355,9 @@ public boolean __setattr_null(java.lang.String name, org.python.Object value) {
             } else {
                 throw new org.python.exceptions.ZeroDivisionError("float division by zero");
             }
-        }
+        }// else if (other instanceof org.python.types.Complex){
+           // return other.__truediv__(this);
+        //}
```
eliasdorneles: Please don't add commented out code, these don't add anything useful and only confuse someone reading the code.

python/common/org/python/types/Float.java
```diff
@@ -423,9 +425,11 @@ public boolean __setattr_null(java.lang.String name, org.python.Object value) {
                     double result = (((((double) this.value) % other_val) + other_val) % other_val);
                     return new org.python.types.Float(result);
                 }
+            } else if (other instanceof org.python.types.Complex){
+                throw new org.python.exceptions.TypeError("can't mod complex numbers.");
             }
         } catch (org.python.exceptions.TypeError e) {
-            throw new org.python.exceptions.TypeError("unsupported operand type(s) for %: 'float' and '" + other.typeName() + "'");
+            throw e;
```
eliasdorneles: If we're just rethrowing this exception, maybe we should simply not catch `TypeError` and let it raise.

python/common/org/python/types/Float.java
```diff
@@ -456,7 +460,7 @@ public boolean __setattr_null(java.lang.String name, org.python.Object value) {
         }
 
         if (other instanceof org.python.types.Int) {
-            long other_val = ((org.python.types.Int) other).value;
+            long other_val = ((org.python.types.Int) other).value; // exponent
```
eliasdorneles: Was this change intended?
Seems a bit out of scope for this PR. Anyway, instead of adding a comment, a more useful change would be to rename the variable to be `exponent`.

python/common/org/python/types/Float.java
student-06: Oh sorry for this. I have removed comments. I won't repeat this again and thanks.

python/common/org/python/types/Float.java
student-06: Oh sorry my bad. I was just reading code and wrote some comments over here and there. I will remove those. Thanks again.

python/common/org/python/types/Float.java
student-06: So should I remove try catch block ?

eliasdorneles:
Looks good, thanks @student-06 !

PR #716, opened 2017-12-13
Working in Progress

siddharthkawar91:
In this ticket I fixed the `test_subscr_slice` in class BinaryStrOperationTests(tests/datatypes/test_str.java)
When I run the below test command `python setup.py test -s tests.datatypes.tests_str` test ends with `ok`.
When I run the full test command `python setup.py test`. The test ends with expected failure and the failure is the unexpected success in test_tuple (tests.builtins.test_set.BuiltinSetFunctionTests) then re-ran the test command without my changes and again this resulted in `unexpected success`. Kindly guide me.

PR #738, opened 2018-02-26
Fix test_subscr_slice

BPYap:
related #36 
Fix the same issue as #716 which is still in WIP status

### DIFF
--- a/python/common/org/python/types/Str.java
+++ b/python/common/org/python/types/Str.java
@@ -282,24 +282,35 @@ public class Str extends org.python.types.Object {
             if (index instanceof org.python.types.Slice) {
                 org.python.types.Slice.ValidatedValue slice = ((org.python.types.Slice) index).validateValueTypes();
                 java.lang.String sliced;
-
                 if (slice.start == null && slice.stop == null && slice.step == null) {
                     sliced = this.value;
                 } else {
                     long start;
                     if (slice.start != null) {
-                        start = toPositiveIndex(slice.start.value);
+                        if (slice.start.value < 0) {
+                            // computes starting index from end of string
+                            // sets to 0 if computed value went below 0
+                            start = Math.max(0, this.value.length() + slice.start.value);
+                        } else {
+                            start = slice.start.value;
+                        }
                     } else {
                         start = 0;
                     }
 
                     long stop;
                     if (slice.stop != null) {
-                        stop = toPositiveIndex(slice.stop.value);
+                        if (slice.stop.value < 0) {
+                            // computes ending index from end of string
+                            // sets to 0 if computed value went below 0
+                            stop = Math.max(0, this.value.length() + slice.stop.value);
+                        } else {
+                            // if stop goes beyond string length, sets it to the length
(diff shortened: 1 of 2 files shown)


python/common/org/python/types/Str.java
```diff
@@ -282,24 +282,35 @@ public boolean __setattr_null(java.lang.String name, org.python.Object value) {
...
                 org.python.types.Slice.ValidatedValue slice = ((org.python.types.Slice) index).validateValueTypes();
                 java.lang.String sliced;
-
                 if (slice.start == null && slice.stop == null && slice.step == null) {
                     sliced = this.value;
                 } else {
                     long start;
                     if (slice.start != null) {
-                        start = toPositiveIndex(slice.start.value);
+                        if (slice.start.value < 0) {
```
cflee: I know the original logic flow was like this, and we might not really want to touch 'working' code, but I think we can take this opportunity to make this section of code a bit cleaner since it's looking a bit more involved... Declare and instantiate `start` to 0 immediately above, since that's the 'default' value. Then that removes the need for the `else` block below.

Similarly in the rest of this function.

python/common/org/python/types/Str.java
```diff
@@ -309,11 +320,21 @@ public boolean __setattr_null(java.lang.String name, org.python.Object value) {
...
+                        if (start < stop) {
+                            sliced = this.value.substring((int) start, (int) stop);
+                        } else {
+                            sliced = "";
+                        }
                     } else {
                         java.lang.StringBuffer buffer = new java.lang.StringBuffer();
-                        for (long i = start; i < stop; i += step) {
+                        long i = start;
+                        while (true) {
```
cflee: Would it be clearer to have a `while (condition)` instead of following `while (true)` with an if block to break?

Also, I'm just curious, is there a specific reason to switch from a for loop to a while loop? We could essentially express the same logic as the condition in the for loop, and the logic is still counting steps away from start.

BPYap:
@cflee Thank you so much for your feedback 😄 

I refactored the codes in `Str.java` as suggested and the reason I replaced for-loop with while-loop is only for aesthetic reason as the terminating statement is too long to fit in for-loop...

I've also added a new test case on top of `test_str.py`. I tried to only keep essential test cases because if the size of test case gets too large, I will get an error stating `voc.python.blocks.BlockCodeTooLarge`. I could have multiple executions of `assertCodeExecution` on subsets of test cases to avoid the error but I feel that the extended test cases is already sufficient, any suggestion on this?
Also, am I implementing the additional test cases correctly?

freakboy3742:
@BPYap Ah.. BlockCodeTooLarge... :-)

So - an explanation: It turns out that Java has a hard limit on the number of opcodes that a single method can have. Sometimes VOC isn't very efficient with it's use of opcodes, so if you have a test method that rolls out lots of examples, it's easy to hit that limit.

If you need to exceed that limit, the fix is to break the code into two separate test methods. However, in this case, I think you've got enough tests to cover the obvious edge cases.

BPYap:
Thanks for explanation and approval 😄

PR #739, opened 2018-02-26
added ctime() code in time.java

prithvisingh18:
Implementing `ctime()` from `time.py`(standard library) in `time.java` .

### DIFF
--- a/python/common/python/time.java
+++ b/python/common/python/time.java
@@ -101,6 +101,18 @@ public class time extends org.python.types.Module {
             long currentTimeInMillis = System.currentTimeMillis();
             date = new java.util.Date(currentTimeInMillis);
         } else {
+            //Type checking the parameter
+            String parameterType = seconds.getClass().getSimpleName();
+            if (!parameterType.equals("Int") && !parameterType.equals("Float")) {
+                if (parameterType.equals("Str")) {
+                    parameterType = "str";
+                } else if (parameterType.equals("List")) {
+                    parameterType = "list";
+                } else if (parameterType.equals("Tuple")) {
+                    parameterType = "tuple";
+                }
+                throw new org.python.exceptions.TypeError("an integer is required (got type " + parameterType + ")");
+            }
             date = new java.util.Date(((org.python.types.Int) seconds.__int__()).value * 1000L);
         }
         java.text.SimpleDateFormat ft = new java.text.SimpleDateFormat("E MMM dd HH:mm:ss yyyy");
--- a/python/common/python/time.java
+++ b/python/common/python/time.java
@@ -106,7 +106,7 @@ public class time extends org.python.types.Module {
         return new org.python.types.Str(ft.format(currentDate));
     }
     public static org.python.Object ctime(org.python.Object seconds) {
-        java.util.Date currentDate = new java.util.Date(((org.python.types.Int) seconds.__int__()).value * 1000);
+        java.util.Date currentDate = new java.util.Date((long) (((org.python.types.Int) seconds.__int__()).value * 1000));
         java.text.SimpleDateFormat ft = new java.text.SimpleDateFormat ("E MMM dd HH:mm:ss yyyy");
         return new org.python.types.Str(ft.format(currentDate));
     }
--- a/python/common/python/time.java
+++ b/python/common/python/time.java
@@ -102,16 +102,8 @@ public class time extends org.python.types.Module {
             date = new java.util.Date(currentTimeInMillis);
         } else {
             //Type checking the parameter
-            String parameterType = seconds.getClass().getSimpleName();
-            if (!parameterType.equals("Int") && !parameterType.equals("Float")) {
-                if (parameterType.equals("Str")) {
-                    parameterType = "str";
-                } else if (parameterType.equals("List")) {
-                    parameterType = "list";
-                } else if (parameterType.equals("Tuple")) {
-                    parameterType = "tuple";
-                }
-                throw new org.python.exceptions.TypeError("an integer is required (got type " + parameterType + ")");
+            if (!(seconds instanceof org.python.types.Int) && !(seconds instanceof org.python.types.Float)) {
+                throw new org.python.exceptions.TypeError("an integer is required (got type " + seconds.typeName() + ")");
             }
             date = new java.util.Date(((org.python.types.Int) seconds.__int__()).value * 1000L);
         }
--- a/python/common/python/time.java
+++ b/python/common/python/time.java
@@ -100,15 +100,15 @@ public class time extends org.python.types.Module {
                       "errors defaults to 'strict'."
     )
     public static org.python.Object ctime() {
-            long currentDateTime = System.currentTimeMillis();
-            java.util.Date currentDate = new java.util.Date(currentDateTime);
-            java.text.SimpleDateFormat ft = new java.text.SimpleDateFormat ("E MMM dd HH:mm:ss yyyy");
-            return new org.python.types.Str(ft.format(currentDate));
+        long currentDateTime = System.currentTimeMillis();
+        java.util.Date currentDate = new java.util.Date(currentDateTime);
+        java.text.SimpleDateFormat ft = new java.text.SimpleDateFormat ("E MMM dd HH:mm:ss yyyy");
+        return new org.python.types.Str(ft.format(currentDate));
     }
     public static org.python.Object ctime(org.python.Object seconds) {
-            java.util.Date currentDate = new java.util.Date(((org.python.types.Int) seconds.__int__()).value * 1000);
-            java.text.SimpleDateFormat ft = new java.text.SimpleDateFormat ("E MMM dd HH:mm:ss yyyy");
-            return new org.python.types.Str(ft.format(currentDate));
+        java.util.Date currentDate = new java.util.Date(((org.python.types.Int) seconds.__int__()).value * 1000);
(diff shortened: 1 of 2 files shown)


python/common/python/time.java
```diff
@@ -89,10 +89,26 @@ public time() {
     }
 
     @org.python.Method(
-            __doc__ = ""
+            __doc__ = "str(object='') -> str\n"+
```
eliasdorneles: This docstring seems completely unrelated to the one for the ``time.ctime`` function.

python/common/python/time.java
```diff
@@ -89,10 +89,26 @@ public time() {
...
+                      "str(bytes_or_buffer[, encoding[, errors]]) -> str\n"+
+                      "Create a new string object from the given object. If encoding or"+
+                      "errors is specified, then the object must expose a data buffer"+
+                      "that will be decoded using the given encoding and error handler."+
+                      "Otherwise, returns the result of object.__str__() (if defined)"+
+                      "or repr(object)."+
+                      "encoding defaults to sys.getdefaultencoding().\n"+
+                      "errors defaults to 'strict'."
     )
     public static org.python.Object ctime() {
```
eliasdorneles: This function is supposed to take an optional argument: you can pass the number of seconds:

```
>>> import time
>>> time.ctime(0)
'Thu Jan  1 01:00:00 1970'
>>> now = time.time()
>>> time.ctime(now - now % 3600)
'Mon Feb 26 15:00:00 2018'
```

The above are good candidates for tests, btw.

python/common/python/time.java
prithvisingh18: I agree but this is what python3 returns when `time.ctime().__doc__` .
Is there an alternative ?

python/common/python/time.java
prithvisingh18: I have overloaded the function `ctime()` for optional parameter, what could be the reason for the following error, 
`TypeError: ctime() takes 0 positional arguments but 1 was given`
Thanks.

python/common/python/time.java
eliasdorneles: That's giving you the docstring of the result of calling the function (which is a string).
Try `time.ctime.__doc__`.

python/common/python/time.java
eliasdorneles: Check the docs section on Python signatures for Java-defined methods: http://voc.readthedocs.io/en/latest/reference/signatures.html

python/common/python/time.java
```diff
@@ -106,7 +106,7 @@ public time() {
         return new org.python.types.Str(ft.format(currentDate));
     }
     public static org.python.Object ctime(org.python.Object seconds) {
-        java.util.Date currentDate = new java.util.Date(((org.python.types.Int) seconds.__int__()).value * 1000);
+        java.util.Date currentDate = new java.util.Date((long) (((org.python.types.Int) seconds.__int__()).value * 1000));
```
uranusjr: Would it be a good idea to cast to long *before* multiplying?

prithvisingh18: I am not sure if `int` would be large enough hence, I used `(long)` as the seconds could be a large.

uranusjr: I understand the line of thinking, but it’s the implementation I’m having doubts about. With the current implementation, the number is first times 1000, and then casted to long. This wouldn’t prevent anything because the cast happens after the overflow (if there is one).

To actually prevent the overflow you’d need to do something like

```java
new java.util.Date(((long) ((org.python.types.Int) seconds.__int__()).value) * 1000)
```

or (probably better)

```java
new java.util.Date(((org.python.types.Int) seconds.__int__()).value * 1000L)
```

so the large number is casted to long first, and then multiplied.

python/common/python/time.java
```diff
@@ -89,10 +89,30 @@ public time() {
...
     )
-    public static org.python.Object ctime() {
-        throw new org.python.exceptions.NotImplementedError("time.ctime() has not been implemented.");
+    public static org.python.Object ctime(org.python.Object seconds) {
+        java.util.Date date;
+        if(seconds == null){
+            long currentTimeInMillis = System.currentTimeMillis();
+            date = new java.util.Date(currentTimeInMillis);
+        } else {
+            date = new java.util.Date((long) (((org.python.types.Int) seconds.__int__()).value * 1000));
```
eliasdorneles: Is the call to `__int__` really needed here?
Wouldn't simply `((org.python.types.Int) seconds).value` work already?

tests/stdlib/test_time.py
```diff
@@ -106,12 +106,19 @@ def test_clock(self):
...
             import time
             print(time.ctime())
             """)
+    def test_ctime_with_parameter(self):
+        self.assertCodeExecution("""
+            import time
+            print(time.ctime(0))
+            print(time.ctime(1000))
+            now = time.time()
+            time.ctime((now - (now % 3600)))
```
eliasdorneles: This line is missing a print: if we don't print, the result isn't checked.

tests/stdlib/test_time.py
```diff
@@ -106,12 +106,19 @@ def test_clock(self):
 
     #######################################################
     # ctime
-    @expectedFailure
     def test_ctime(self):
         self.assertCodeExecution("""
             import time
             print(time.ctime())
```
eliasdorneles: This test will be very hard to pass because it depends too much on the time of execution.

Your best bet for this test is to discard some of the precision, like: `print(time.ctime()[:10], time.ctime()[-4:])`

This is considering only the `date` information and discarding the hours. It's still a flaky test because it will fail when it runs at exactly midnight, but it's at least giving us good confidence that the code does what it's supposed to do.

Makes sense?

prithvisingh18: I've also had the same problem, seems right will do it now.

python/common/python/time.java
prithvisingh18: Thanks.

goanpeca:
ping ping @eliasdorneles  :-p

tests/stdlib/test_time.py
```diff
@@ -106,12 +106,19 @@ def test_clock(self):
...
+            print(time.ctime()[:10], time.ctime()[-4:])
             """)
+    def test_ctime_with_parameter(self):
+        self.assertCodeExecution("""
+            import time
+            print(time.ctime(0))
+            print(time.ctime(1000))
+            now = time.time()
+            print(time.ctime((now - (now % 3600))))
+        """)
```
eliasdorneles: Can you add a test for the case of wrong parameter, like if we give it a string `time.ctime('today')` ?
It should show the same message as CPython (you'll need to do add a type-check for the argument).

python/common/python/time.java
```diff
@@ -89,10 +89,30 @@ public time() {
...
+        java.text.SimpleDateFormat ft = new java.text.SimpleDateFormat("E MMM dd HH:mm:ss yyyy");
+        String padded_date = ft.format(date);
+        if(Character.toString(padded_date.charAt(8)).equals("0")) {
+            java.lang.StringBuilder sb = new StringBuilder();
+            sb.append(padded_date.substring(0,7));
+            sb.append("  ");
+            sb.append(padded_date.substring(9,padded_date.length()));
+            padded_date = sb.toString();
+        }
+        return new org.python.types.Str(padded_date);
```
eliasdorneles: Can you please run `ant checkstyle` locally and fix the style errors that it finds?
Normally we can see these in the CI, but [it seems to not be working correctly right now](https://github.com/pybee/voc/issues/640).

prithvisingh18: Yeah sure.

python/common/python/time.java
```diff
@@ -89,10 +89,42 @@ public time() {
...
+            String parameterType = seconds.getClass().getSimpleName();
+            if (!parameterType.equals("Int") && !parameterType.equals("Float")) {
+                if (parameterType.equals("Str")) {
+                    parameterType = "str";
+                } else if (parameterType.equals("List")) {
+                    parameterType = "list";
+                } else if (parameterType.equals("Tuple")) {
+                    parameterType = "tuple";
+                }
+                throw new org.python.exceptions.TypeError("an integer is required (got type " + parameterType + ")");
```
eliasdorneles: As you can see, this will not work for all existing types.
There are several examples in the codebase on how to do this, look at the implementation of `List.insert` for example.

python/common/python/time.java
```diff
@@ -89,10 +89,34 @@ public time() {
...
-    public static org.python.Object ctime() {
-        throw new org.python.exceptions.NotImplementedError("time.ctime() has not been implemented.");
+    public static org.python.Object ctime(org.python.Object seconds) {
+        java.util.Date date;
+        if (seconds == null) {
+            long currentTimeInMillis = System.currentTimeMillis();
+            date = new java.util.Date(currentTimeInMillis);
+        } else {
+            //Type checking the parameter
+            if (!(seconds instanceof org.python.types.Int) && !(seconds instanceof org.python.types.Float)) {
```
eliasdorneles: I see this is testing for float but we're not exercising it in the tests. Can you add a test passing a float as well?

prithvisingh18: I added another float test case too. BTW [print(time.ctime((now - (now % 3600)))](https://github.com/pybee/voc/pull/739/files#diff-3e3a3c1d82261ce04faab17d5b8024c0R121) already had a float test.

eliasdorneles:
Thanks @prithvisingh18, merging !! :tada:

PR #742, opened 2018-02-26
This fixes test_subscr_bool in List.java

student-06:
#36

### DIFF
--- a/python/common/org/python/types/List.java
+++ b/python/common/org/python/types/List.java
@@ -417,6 +417,13 @@ public class List extends org.python.types.Object {
                     }
                 }
                 return new org.python.types.List(sliced);
+            } else if (index instanceof org.python.types.Bool){
+                int idx = ((org.python.types.Bool)index).value ? 1 : 0;
+                if (idx >= this.value.size()){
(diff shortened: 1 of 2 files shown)


uranusjr:
I feel most of this can be merged into the `int` case…

student-06:
@uranusjr Well I merged bool and int case. Please review again and thanks for the suggestion.

python/common/org/python/types/List.java
```diff
@@ -418,7 +418,13 @@ public boolean __setattr_null(java.lang.String name, org.python.Object value) {
                 }
                 return new org.python.types.List(sliced);
             } else {
-                int idx = (int) ((org.python.types.Int) index).value;
+                int idx;
+                if (index instanceof org.python.types.Bool){
+                    idx = ((org.python.types.Bool)index).value ? 1 : 0;
```
eliasdorneles: This seems the case of doing:

    idx = ((org.python.types.Bool) index).__int__().value;

student-06:
@eliasdorneles Thanks for the review, mentioned minor change is incorporated.

PR #748, opened 2018-03-01
Accept non-Android vendors as Java by default. Fixes #747 (jumbo jet)

### LINKED ISSUE #747: Could not find or load main class
This might be a silly question but I'm getting "Error: Could not find or load main class python.hello" no matter what I try.
Following the [Tutorial](https://voc.readthedocs.io/en/latest/tutorial/tutorial-0.html), I've installed VOC, using OpenJDK that comes with Android Studio
```
(env) λ javac -version
javac 1.8.0_152-release
```
VOC is installed as per instructions, directly from GitHub. Virtualenv has been created and VOC installed via `pip -e`. Ant generated python-java-support.jar.

Then I've created a separate directory for the small test project. Created `hello.py` with only the print command. VOC compiled it into python\hello.class.

javap confirms the class was created:
```
(env) λ javap .\hello.class
Compiled from "hello.py"
public class python.hello extends org.python.types.Module {
  public void module$import();
  public python.hello();
  public org.python.Object __new__(org.python.Object);
  public static void main(java.lang.String[]);
}
```

However, whether I call from the directory where hello.py is, or from python directory, I'm getting the above error. I moved the .jar file into python directory and am running
```
(env) λ java -classpath python-java-support.jar:. python.hello
```
to no avail. Any hints, please? Thanks!

MisterY:


python/common/python/platform.java
```diff
@@ -27,6 +27,8 @@
             platform_class_name = "python._platform.AndroidPlatform";
         } else if (vendor.equals("Azul Systems, Inc.")) {
             platform_class_name = "python._platform.JavaPlatform";
+        } else if (vendor.equals("JetBrains s.r.o")) {
+            platform_class_name = "python._platform.JavaPlatform";
         } else {
             throw new org.python.exceptions.RuntimeError("Unknown platform vendor '" + vendor + "'");
```
eliasdorneles: To prevent this problem from happening in the future, I believe it will be good to stop raising this exception and just do this:

```
if (vendor.equals("The Android Project")) {
    platform_class_name = "python._platform.AndroidPlatform";
} else {
    platform_class_name = "python._platform.JavaPlatform";
}
```

Do you mind doing that change?

MisterY: Here we go

MisterY: That's a good suggestion.

PR #751, opened 2018-03-01
This fixes all tests for float type except test_power_complex

student-06:
#36.
@eliasdorneles  In Complex.java, I have changed line 450 since it is clearly wrong(kind of typo), [verify here].(https://github.com/python/cpython/blob/master/Objects/complexobject.c#L128).
Thank you.

### DIFF
--- a/python/common/org/python/types/Float.java
+++ b/python/common/org/python/types/Float.java
@@ -305,36 +305,47 @@ public class Float extends org.python.types.Object {
     )
     public org.python.Object __mul__(org.python.Object other) {
 
-        if (other instanceof org.python.types.Str) {
+        if (org.python.types.Object.isSequence(other)) {
             throw new org.python.exceptions.TypeError("can't multiply sequence by non-int of type '" + "float" + "'");
         } else if (other instanceof org.python.types.Int) {
             return new org.python.types.Float(this.value * ((org.python.types.Int) other).value);
         } else if (other instanceof org.python.types.Float) {
             return new org.python.types.Float(((double) this.value) * ((org.python.types.Float) other).value);
         } else if (other instanceof org.python.types.Bool) {
-            return new org.python.types.Float(this.value * (((org.python.types.Bool) other).value ? 1 : 0));
+            return new org.python.types.Float(this.value * (((org.python.types.Bool) other).__int__().value));
         } else if (other instanceof org.python.types.Complex) {
             return ((org.python.types.Complex) other).__mul__(this);
-        } else if (other instanceof org.python.types.Dict) {
-            throw new org.python.exceptions.TypeError("unsupported operand type(s) for *: 'float' and '" + other.typeName() + "'");
-        } else if (other instanceof org.python.types.NoneType) {
-            throw new org.python.exceptions.TypeError("unsupported operand type(s) for *: 'float' and '" + other.typeName() + "'");
-        } else if (other instanceof org.python.types.Set) {
-            throw new org.python.exceptions.TypeError("unsupported operand type(s) for *: 'float' and '" + other.typeName() + "'");
-        } else if (other instanceof org.python.types.FrozenSet) {
-            throw new org.python.exceptions.TypeError("unsupported operand type(s) for *: 'float' and '" + other.typeName() + "'");
-        } else if (other instanceof org.python.types.List) {
-            throw new org.python.exceptions.TypeError("can't multiply sequence by non-int of type 'float'");
-        } else if (other instanceof org.python.types.Tuple) {
-            throw new org.python.exceptions.TypeError("can't multiply sequence by non-int of type 'float'");
-        } else if (other instanceof org.python.types.Slice) {
-            throw new org.python.exceptions.TypeError("unsupported operand type(s) for *: 'float' and 'slice'");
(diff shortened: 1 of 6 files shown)


student-06:
Hi @eliasdorneles, this pull request failed on this test:```beekeeper:1/smoke-test```. In details I found this
```
________________ BuiltinTwoargPowFunctionTests.test_float_float ________________
[gw15] linux -- Python 3.4.6 /usr/local/bin/python
Unexpected success
```
 But on my system it passes the all tests(except one), but the weird part is there is no ```test_float_float``` in ```tests/datatypes/test_float.py```. After a while I did ```grep``` on ```test_float_float``` found it in ```/tests/builtins/test_pow.py```. Now I am bit confused, what has caused this. Could you please look into this ?

Thank you.

eliasdorneles:
Hi! Sorry for the brevity, I'm in my phone on a train with bad connection.

But from the message, it looks like that test_float_float wasn't passing
before your changes (that's why it is in the not_implemented list in
test_pow).

So, your changes seem to have fixed that test (note how it says "Unexpected
success" in the message), therefore you can now remove it from the list of
not implemented. :)

Le 2 mars 2018 6:42 AM, "Siddharth Mundada" <notifications@github.com> a
écrit :

student-06:
@eliasdorneles Hi, I removed it but it there weren't any errors on my system.  Thanks.

eliasdorneles:
> I removed it but it there weren't any errors on my system.

Are you sure you were running the tests for `tests/builtins/test_pow.py` when there was no error in your system?

python/common/org/python/types/Float.java
```diff
@@ -305,37 +305,48 @@ public boolean __setattr_null(java.lang.String name, org.python.Object value) {
...
+        } else if (other instanceof org.python.types.Float) {
+            return new org.python.types.Float(((double) this.value) * ((org.python.types.Float) other).value);
+        } else if (other instanceof org.python.types.Bool) {
+            return new org.python.types.Float(this.value * (((org.python.types.Bool) other).__int__().value));
+        } else if (other instanceof org.python.types.Complex) {
+            return ((org.python.types.Complex) other).__imul__(this);
+        } else {
+            super.__imul__(other);
+        }
+        throw new org.python.exceptions.NotImplementedError("float.__mul__() has not been implemented.");
```
eliasdorneles: This doesn't look right -- if we're adding an implementation, why are we throwing NotImplementedError here?
Let's add a test for a different type (say, passing an empty string `''`) and then catch and print the error -- this probably needs to throw TypeError.

student-06: @eliasdorneles  Oh yes, I made a terrible mistake there. Removing exception throwing will do right ?

eliasdorneles: To be fair, it looks like the mistake already existed in `__mul__` implementation.

> Removing exception throwing will do right ?

I suggest adding a test for the case and let the test tell you what to do (run it, see the failure, figure out what the output should be). :)

student-06: Hi, I actually added test case and it worked(```TypeError``` thing). Please have a look at [this](https://github.com/pybee/voc/pull/751/files#diff-0eec62afc15f3b0631bd40cb3d44b3af). I think I didn't understand the problem in the first place, could you please give me some hints ? Thank you.

student-06:
@eliasdorneles Hi, I am having little trouble understanding tests. for eg, for [this](https://github.com/pybee/voc/pull/751/commits/6b01ac0d2a710afeb629f80d1fddc6ee60a1ca0f) commit, it fails on these python3.5 and python3.6. Please have a look at this screenshot 
screenshot from 2018-03-03 09-55-24.
Here on my system with python3.6 it worked, but here it failed. 
To best of my knowledge, I have not changed/committed any files related to the test on which it is failing. 
So how do I approach this problem ? Sorry for the trouble.
Thank you.

cflee:
Ok, this is an issue with the test suite. There is a list of substitutions that it does to the java/python output, so that some 'random' or 'undefined' things like the ordering of items in a set don't fail the test. The test that is failing is `BuiltinSetFunctionTests.test_tuple`, and as you can see in the test output by clicking on the red cross, it's just a different order in how a set is being printed that is causing the test to fail:

```diff
- {3, 1.2, True}
?  ---
+ {1.2, 3, True}
?      +++
```

This is the list of tuples that are used in the test suite:

https://github.com/pybee/voc/blob/78d2360c7f9f94daa2eae2c40a66d74f66926797/tests/utils.py#L733-L739

These are the sets that will be substituted/normalised for output in the test suite:

https://github.com/pybee/voc/blob/78d2360c7f9f94daa2eae2c40a66d74f66926797/tests/utils.py#L757-L763

If you add the `{1.2, 3, True}` set to the list of standard substitutions, the test should reliably pass both locally _and_ on the automated builds.

student-06:
@cflee a line like this would fix  ``` "{'3', '1.2', 'True'}": _string_substitutions("{'1.2', '3', 'True'}"),``` ? red should be replaced by green right ? Thank you for the suggestion.

student-06:
@eliasdorneles Hi, could you please review this pr ? thank you.

PR #755, opened 2018-03-05
implement hashCode for complex datatype.

svmundada:
1. Complex number hashing is taken from cpython's [hashing function](https://github.com/python/cpython/blob/master/Objects/complexobject.c#L408).

2. I made __repr__ and __str__ methods same for complex numbers, I am not exactly sure if it is the right thing but from here cypthon ```complexobject.c``` module both seem to have same underlining function.
Link to cpython's corressponding functions:
[```__repr__```](https://github.com/python/cpython/blob/master/Objects/complexobject.c#L1126) and [```__str__```](https://github.com/python/cpython/blob/master/Objects/complexobject.c#L1132).

Thank you.

### DIFF
--- a/python/common/org/python/types/Complex.java
+++ b/python/common/org/python/types/Complex.java
@@ -23,6 +23,9 @@ public class Complex extends org.python.types.Object {
     }
 
     public int hashCode() {
+        // 0xf4243 is _PyHASH_MULTIPLIER macro defined in Cpython's Include/pyhash.h.
+        // Its just a prime number (happens to be multiplier) used in hashing.
+        // https://computinglife.wordpress.com/2008/11/20/why-do-hash-functions-use-prime-numbers/
         return this.real.hashCode() + 0xf4243 * this.imag.hashCode();
     }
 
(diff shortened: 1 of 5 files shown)


python/common/org/python/types/Complex.java
```diff
@@ -23,7 +23,7 @@
     }
 
     public int hashCode() {
-        return this.hashCode();
+        return this.real.hashCode() + 0xf4243 * this.imag.hashCode();
```
freakboy3742: What's the significance of 0xf4342? It feels like a magic number - was it a randomly chosen magic number, or does it have history in the CPython codebase? Whatever the explanation, this warrants a comment in the code to explain where the number came from, and what it's being used for.

python/common/org/python/types/Complex.java
```diff
@@ -162,31 +162,11 @@ private String partToStr(org.python.types.Float x) {
...
-        } else {
-            real_present = false;
-        }
-        if (this.real.value != 0 && this.imag.value >= 0) {
-            buffer.append("+");
-        }
-        if (((org.python.types.Bool) ((this.imag).__int__().__eq__(this.imag))).value) {
-            buffer.append(((org.python.types.Str) (this.imag).__int__().__repr__()).value);
+        if (this.real.value != 0.0 || this.real.isNegativeZero()) {
+            return new org.python.types.Str("(" + partToStr(this.real) + ((this.imag.value >= 0.0 && !this.imag.isNegativeZero()) ? "+" : "-") + partToStr(new org.python.types.Float(Math.abs(this.imag.value))) + "j)");
```
freakboy3742: I'm surprised this wasn't picked up by linting... but we don't use ternary operators, and the line length is *extremely* excessive.

freakboy3742 (changes requested):
Can't argue with the results, but there's some comments and style issues that need to be addressed.

python/common/org/python/types/Complex.java
student-06: Actually we are using ternary at many places in voc, so should we remove all of those ? I'll make that last line bit more readable.

freakboy3742: We shouldn't be - and the linter *should* be catching them. If it isn't, then we have another problem.

student-06:
@freakboy3742 could you please check the [new](https://github.com/pybee/voc/pull/755/commits/20c92c92b91016de0204fdf6a20263a71d850636) commit.
Thank you.

phildini:

PR #753, opened 2018-03-05
reorganised ByteArray, Bytes and test cases

kartik-karz:

PR #756, opened 2018-03-06
implement bytearray isalnum() and isdigit() with tests Refs #36

sziraqui:
A helper method Bytes._isnum() is created to keep code cleaner and exploit existing Bytes._isalpha(). Both are then used in ByteArray.isalnum()

### DIFF
--- a/python/common/org/python/types/ByteArray.java
+++ b/python/common/org/python/types/ByteArray.java
@@ -761,8 +761,16 @@ public class ByteArray extends org.python.types.Object {
     @org.python.Method(
             __doc__ = "B.isalnum() -> bool\n\nReturn True if all characters in B are alphanumeric\nand there is at least one character in B, False otherwise."
     )
-    public org.python.Object isalnum(java.util.List<org.python.Object> args, java.util.Map<java.lang.String, org.python.Object> kwargs, java.util.List<org.python.Object> default_args, java.util.Map<java.lang.String, org.python.Object> default_kwargs) {
-        throw new org.python.exceptions.NotImplementedError("bytearray.isalnum has not been implemented.");
+    public org.python.Object isalnum() {
+        if (this.value.length == 0) {
+            return new org.python.types.Bool(false);
+        }
+        for (byte ch: this.value) {
+            if (!Bytes._isalpha(ch) && !Bytes._isnum(ch)) {
+                return new org.python.types.Bool(false);
+            }
+        }
+        return new org.python.types.Bool(true);
     }
 
     @org.python.Method(
--- a/python/common/org/python/types/ByteArray.java
+++ b/python/common/org/python/types/ByteArray.java
@@ -762,15 +762,7 @@ public class ByteArray extends org.python.types.Object {
             __doc__ = "B.isalnum() -> bool\n\nReturn True if all characters in B are alphanumeric\nand there is at least one character in B, False otherwise."
     )
     public org.python.Object isalnum() {
-        if (this.value.length == 0) {
-            return new org.python.types.Bool(false);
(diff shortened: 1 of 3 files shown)


python/common/org/python/types/ByteArray.java
```diff
@@ -761,8 +761,16 @@ public boolean __setattr_null(java.lang.String name, org.python.Object value) {
...
             __doc__ = "B.isalnum() -> bool\n\nReturn True if all characters in B are alphanumeric\nand there is at least one character in B, False otherwise."
     )
-    public org.python.Object isalnum(java.util.List<org.python.Object> args, java.util.Map<java.lang.String, org.python.Object> kwargs, java.util.List<org.python.Object> default_args, java.util.Map<java.lang.String, org.python.Object> default_kwargs) {
-        throw new org.python.exceptions.NotImplementedError("bytearray.isalnum has not been implemented.");
+    public org.python.Object isalnum() {
+        if (this.value.length == 0) {
+            return new org.python.types.Bool(false);
+        }
+        for (byte ch: this.value) {
+            if (!Bytes._isalpha(ch) && !Bytes._isnum(ch)) {
```
goanpeca: Is this correct `!Bytes._isalpha(ch) && !Bytes._isnum(ch)` ?

Should it not be more like:
`!(Bytes._isalpha(ch) && Bytes._isnum(ch))`

sziraqui: It is correct because:
!AlphaNum = !(Alpha || Num) = !Alpha && !Num

But !(Alpha && Num) works too because any alphanum ch cannot be aplha and num at the same time. Your logic is slightly more efficient. Should I switch to your logic?

goanpeca: On second thought @sziraqui its fine I think as you have it

sziraqui: Cool! Thanks for reviewing.

sziraqui: Ready to be merged?

goanpeca: @eliasdorneles is the boss :-p

sziraqui: ohk

goanpeca: Also, you need to add tests!

goanpeca: Already being tacked on https://github.com/pybee/voc/pull/753

sziraqui: oops saw your message late. I had mentioned that I am working on these functions in issue 36. Hope someone will resolve the conflict

eliasdorneles: So, this is already implemented in Bytes, see: https://github.com/pybee/voc/blob/78d2360c7f9f94daa2eae2c40a66d74f66926797/python/common/org/python/types/Bytes.java#L1053

For ByteArray, you just need to add tests and reuse that implementation.

sziraqui: Ok. I thought these are supposed to be independent because they are non-static.

kartik-karz: @sziraqui I think I should have mentioned the pull request properly sorry, I have implemented in a similar way including the helper utility functions... and have done that for isdigit,isalpha,isalnum and used that in the ByteArray... even the test cases are done... waiting for the boss...

sziraqui: Ok no issues. I will look for existing active forks before pushing next time. You can make our life easier by referencing issue number in PR message :)

PR #760, opened 2018-03-09
Implement join() in Bytes and ByteArray

sziraqui:
join method is implemented for Bytes. The same is called in ByteArray.join() but the result here is casted to ByteArray as per python3 behaviour. Refs #38

### DIFF
--- a/python/common/org/python/types/ByteArray.java
+++ b/python/common/org/python/types/ByteArray.java
@@ -808,10 +808,13 @@ public class ByteArray extends org.python.types.Object {
     }
 
     @org.python.Method(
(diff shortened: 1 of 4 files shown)


sziraqui:
Pretty strange to see smoke test failing that too for isalnum which is not even changed. Its the same isalnum that was passed all tests in previous PR. Still I ran the test locally with py3.4.8, it passed all tests

sziraqui:
OMG! Theres a random change in isalnum test case itself. I might have pressed few keys mistakenly

sziraqui:
@freakboy3742 Is this ready to merged or any changes are required?

PR #761, opened 2018-03-10
Implemented ljust method in Bytes with suitable test cases.

kartik-karz:
The method **ljust** inside the Bytes.java file from python/common/org/python/types has been implemented with suitable test cases inside tests/datatypes/test_bytes.py
This method aligns the bytes in left justified manner iff width is more than the length of the byte [] provided, which also accepts custom filling character if specified, defaulted to " ".


kartik-karz:
forgot to squash the commits plz wait.

kartik-karz:
Oh that's lame... please wait

PR #762, opened 2018-03-10
Implemented ljust method in Bytes with suitable test cases.

kartik-karz:
The method **ljust** inside the Bytes.java file from python/common/org/python/types has been implemented with suitable test cases inside tests/datatypes/test_bytes.py
This method aligns the bytes in left justified manner iff width is more than the length of the byte [] provided, which also accepts custom filling character if specified, defaulted to " ".

### DIFF
--- a/tests/datatypes/test_bytes.py
+++ b/tests/datatypes/test_bytes.py
@@ -349,6 +349,32 @@ class BytesTests(TranspileTestCase):
             print(b'\x46\x55\x43\x4B'.upper())
         """)
 
+    def test_ljust(self):
+        self.assertCodeExecution("""
+            print(b'testMoreThanWidth'.ljust(5))
+            print(b'testEqualWidth'.ljust(14))
+            print(b'testLessThanWidth'.ljust(20))
+            print(b'testMoreWithFill'.ljust(2, b'x'))
+            print(b'testEqualWithFill'.ljust(17, b'x'))
+            print(b'testLessWithFill'.ljust(25, b'x'))
+            print(b'testNegative'.ljust(-20))
+            print(b''.ljust(5))
+            print(b'testNoChangeWidthOne'.ljust(True, b'x'))
+            print(b'testBArraySecondArg'.ljust(True, bytearray(b'x')))
+            try:
+                print(b'testStrArgError'.ljust('5'))
+            except Exception as e:
+                print(str(e))
+            try:
+                print(b'testMoreLengthError'.ljust(12, b'as'))
+            except Exception as e:
+                print(str(e))
+            try:
+                print(b'testStrFillingChar'.ljust(12, 'a'))
+            except Exception as e:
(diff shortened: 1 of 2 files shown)


kartik-karz:
@freakboy3742 it's stuck again...

freakboy3742:
/me waves fist at CI server...

I've just poked it again. Apologies for the unreliability.

tests/datatypes/test_bytes.py
```diff
@@ -349,6 +349,29 @@ def test_upper(self):
...
+            print(b'testEqualWithFill'.ljust(17, b'x'))
+            print(b'testLessWithFill'.ljust(25, b'x'))
+            print(b'testNegative'.ljust(-20))
+            print(b''.ljust(5))
+            print(b'testNoChangeWidthOne'.ljust(True, b'x'))
+            print(b'testErrorSecondArg'.ljust(True, bytearray(b'x')))
+        """)
+        self.assertCodeExecution("""
+            print(b'test'.ljust('5'))
+        """, exits_early=True)
```
freakboy3742: `exits_early` is a last resort testing mechanism; it's better to use a try-catch block in the sample code itself, printing something to the console if an exception *isn't* raised (so the test case has something to validate against).

As a side effect, this means you can do the entire test in a single assertCodeExecution block, which is slightly faster - not by a *lot*, but by enough that when you count up all the times we start assertCodeExecution blocks, it adds up.

kartik-karz:
@freakboy3742 Added try-catch can you have look again? but I didn't get the reason for printing it...as all the three test cases are sure to produce errors and will only go to the exception part. Maybe I mistook you do check if that's how you wanted it.

tests/datatypes/test_bytes.py
```diff
@@ -349,6 +349,26 @@ def test_upper(self):
...
+            print(b'testEqualWithFill'.ljust(17, b'x'))
+            print(b'testLessWithFill'.ljust(25, b'x'))
+            print(b'testNegative'.ljust(-20))
+            print(b''.ljust(5))
+            print(b'testNoChangeWidthOne'.ljust(True, b'x'))
+            print(b'testBArraySecondArg'.ljust(True, bytearray(b'x')))
+            try:
+                print(b'testStrArgError'.ljust('5'))
+                print(b'testMoreLengthError'.ljust(12, b'as'))
+                print(b'testStrFillingChar'.ljust(12, 'a'))
```
freakboy3742: You need to wrap each individual test in a try-catch; otherwise, the first test will raise an exception, and the second two wont ever be executed.

kartik-karz:
@freakboy3742 Done.

goanpeca:
Hmm question, are we testing that documentation is producing the same results in an automatic way?

eliasdorneles:
@goanpeca we're not, but that would be nice.

Issue #766, opened 2018-03-15
docker image


jedie:
Is there a docker image that contains a working VOC environment for development as described here: https://voc.readthedocs.io/en/latest/background/install.html ?!?

cflee:
I don't think so, but that would make an interesting contribution, if there aren't any existing images that already have the JDK, Python and Ant installed!

jedie:
Hm... OK, i will try it ;)

cflee:
For future visitors to this issue, please see discussion in #767 – @jedie is now continuing this at https://github.com/jedie/pybee-docker. Thanks!

PR #767, opened 2018-03-15
Docker

jedie:
I made a Dockerfile for easier startup. See #766 

It's currently installs:

- Python 3.5.5
- javac 1.8.0_111-internal
- Apache Ant(TM) version 1.10.2 compiled on February 3 2018

and VOC from github.

After install, it creates, build and run the "Hello World" tutorial-0

A complete output of `docker built` can you find here: https://travis-ci.org/jedie/voc/builds/353894630

It's possible to push a working travis CI run to docker hub.

Maybe it makes sense to include all other PyBee projects into one Docker Image? It's very simple to install all projects via pip from github ;)

Any interest in continuing this?!?


freakboy3742:
Thanks for the contribution; however, I'm not entirely sure I understand why this makes things "easier".

Firstly, it most definitely *won't* make our lives easier in sprint situations. Docker downloads are *huge*, and would (and have, in the past) crippled most networks in a sprint situation. Docker builds can't be (easily) distributed on a USB key, so any - and every - person who turned up to a sprint will need to download the complete image set. And, sprint attendees won't necessarily have docker pre-installed. In contrast, anyone coming to a Python sprint will almost certainly have Python installed - if they don't, we can given them the installer on a USB stick. We can also distribute Java and ANT installers on a USB stick.

Secondly, this isn't a situation that requires a complex setup. You literally need 3 tools installed: Python, Java, and Ant; and they've all got good installers.

Lastly, we don't use Travis. We have our own CI tool (beekeeper) because Travis simply wasn't flexible enough for our needs. Parallelization is the key issue - Beekeeper completes a VOC build for a single platform in a little over 5 minutes. Travis takes almost 2 hours - *if* it completes without hitting the build timeout.

So - thanks for the contribution, but I think we'll pass on this idea. However, if I've missed some particular benefit to a docker container that I'm missing, feel free to educate me 😄

cflee:
Sure, I can see how that doesn't make sense in a PyCon sprint situation. And installing Java and Ant isn't a problem, since you really only need a non-ancient version of each. 

Dev environments – when I was initially getting started with VOC, as someone familiar with Java but not Python, I had some issues sorting out pyenv and virtualenv and whatever else was needed to be able to test the output of various patch versions of 3.4 and 3.5. There _might_ be some benefit to potential contributors in a non-PyCon-sprint, new-to-Python context to have a way to easily setup a dev environment apart from tutorials, and that's an increasingly common way to use Docker images. But yes, we could equally argue that the tutorials are sufficient and all that we want to maintain.

Build environments – I guess Beekeeper/comb already use Docker containers to take care of that aspect.

All that said, even if we don't feel that it's suitable for inclusion in this repo, @jedie can always create their own repo and maintain/publish the Dockerfile and images there! That's another way to contribute to the world of open source :)

jedie:
Thanks for response. My focus was not on sprints ;)
Meanwhile I also think it makes more sense to do a separate project: 

https://github.com/jedie/pybee-docker

The interesting thing will be to install all projects in the container.

The start should be easy. But on the other hand, it is also important to me not to install all sorts of things locally, especially the Android stuff ;)

freakboy3742:
@cflee In addition to the sprint issue, my concern is that getting started with Docker is only easy *if you know and understand Docker*. If you don't, it's yet another tool you have to install, learn, and use. 

We're unapologetically a Python project. If you're going to write Python, you *definitely* need to learn how to use virtual environments, and if you're planning to use multiple versions of Python at once, learning pyenv is well worth your time as well. 

If there's something we can do to improve the onboarding process for new developers, then I'm in favour of that. However, to my mind, adding another complex tool to the requirements won't achieve that. 

That said, I'm also happy to be proven wrong :-) I'd definitely encourage you to put together a standalone project; If I end up in a sprint situation, I might even throw a couple of contributors in the direction of that dockerfile and see what their experience is like.

cflee:
> In addition to the sprint issue, my concern is that getting started with Docker is only easy if you know and understand Docker. If you don't, it's yet another tool you have to install, learn, and use.

Point taken – I think you understand the devs better, and if the typical new developer to the BeeWare projects isn't familiar with/using Docker yet, sure. 

I think where this is coming from is that in some people's worlds (like mine), Docker containers are what they are use day-to-day. So nope, not anyone trying to shove another complex tool somewhere, the intention is just to cater to these folks to get started quickly. A standalone project works just as well for this, and that's great too!

> We're unapologetically a Python project. If you're going to write Python, you definitely need to learn how to use virtual environments, and if you're planning to use multiple versions of Python at once, learning pyenv is well worth your time as well.

Oh, definitely. My point is about avoiding fiddling with _setting up_ the environment and fighting with all the OS-specific quirks that come with it, not about avoiding learning how to use the Python tools.

jedie:
@cflee maybe you can help to bring https://github.com/jedie/pybee-docker to a useable state? ;)

PR #770, opened 2018-03-18
Fix example shell command in tutorial for linux

ssikdar1:
Running this command:
$ java -classpath ../voc/dist/python-java-support.jar:. python.example

Will result in this message:
Error: Could not find or load main class python.example

Because after running ant java the dist folder is alongside voc/ not in it.

Changing the command to this:

$ java -classpath ../dist/python-java-support.jar:. python.example

gives out the correct output printed in the tutorial:

hello world


ssikdar1:
Ah, I created  tutorial0 inside the voc folder by accident.  Because the voc/ also has a folder voc/ inside it
when i read this
>>
Let’s start by creating a tutorial0 directory in the tutorial directory alongside the voc directory you just cloned into
>>
I accidentally made tutorial0 inside voc/tuorial0 and didn't notice.

Sorry, you can ignore the pull request!

PR #777, opened 2018-03-30
Implemented expandtabs method in Bytes with suitable test cases.

kartik-karz:
The method expandtabs inside the Bytes.java file from python/common/org/python/types has been implemented with suitable test cases inside tests/datatypes/test_bytes.py
This method expands the tabs according to the tabsize(default=8) with spaces.
It is to be noted that this method does not replace the tabs with space as many times the tab character ('\t') is encountered but replaces with position parameter for spaces with the input length.
eg
 b'test\t'.expandtabs() will **not be** b'test&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;' but instead will be b'test&nbsp;&nbsp;&nbsp;'

All test cases run properly, and further details can be noted through the test_bytes.py file.

### DIFF
--- a/python/common/org/python/types/Bytes.java
+++ b/python/common/org/python/types/Bytes.java
@@ -971,10 +971,44 @@ public class Bytes extends org.python.types.Object {
     }
 
     @org.python.Method(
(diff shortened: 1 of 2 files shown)


kartik-karz:
@freakboy3742 merge?

PR #784, opened 2018-04-16
Added gmtime(), tests and struct_time class in time.java

amaarahj:
I implemented gmtime() and the struct_time class to store/format the tuple that gmtime() returns. I also added some tests for gmtime().

### DIFF
--- a/python/common/python/time.java
+++ b/python/common/python/time.java
@@ -120,6 +120,8 @@ public class time extends org.python.types.Module {
                 case "Sun":
                     wday = 6;
                     break;
+                default:
+                    System.out.println("Invalid wday");
             }
             for (int i = 0; i < splited.length; ++i) {
                 if (i != 6){
--- a/python/common/python/time.java
+++ b/python/common/python/time.java
@@ -80,14 +80,14 @@ public class time extends org.python.types.Module {
         public org.python.types.Str __repr__() {
             org.python.Object iter = tuple.__iter__();
             String formatted = "time.struct_time(tm_year=" + iter.__next__() +
-                ", tm_mon=" + iter.__next__() +
-                ", tm_mday=" + iter.__next__() +
-                ", tm_hour=" + iter.__next__() +
-                ", tm_min=" + iter.__next__() +
-                ", tm_sec=" + iter.__next__() +
-                ", tm_wday=" + iter.__next__() +
-                ", tm_yday=" + iter.__next__() +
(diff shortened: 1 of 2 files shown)


python/common/python/time.java
```diff
@@ -53,8 +53,98 @@ public time() {
...
+                    "  minutes (0-59)\n" +
+                    "  seconds (0-59)\n" +
+                    "  weekday (0-6, Monday is 0)\n" +
+                    "  Julian day (day in the year, 1-366)\n" +
+                    "  DST (Daylight Savings Time) flag (-1, 0 or 1)\n" +
+                    "If the DST flag is 0, the time is given in the regular time zone;\n" +
+                    "if it is 1, the time is given in the DST time zone;\n" +
+                    "if it is -1, mktime() should guess based on the date and time.\n"
+    )
+    static class struct_time extends org.python.types.Module {
```
eliasdorneles: I believe it would make more sense to make this class to extend `org.python.types.Tuple`.
Can you tell me why is it extending `org.python.types.Module` ?

python/common/python/time.java
```diff
@@ -53,8 +53,98 @@ public time() {
...
+        org.python.types.Tuple tuple;
+
+        public org.python.Object __getitem__(org.python.Object index) {
```
eliasdorneles: It would be nice if this class also added support for accessing named attributes, like `time.struct_time` objects do in CPython:

```
>>> time.gmtime(0).tm_year
1970
>>> time.gmtime(0).tm_hour
0
>>> time.gmtime(0).tm_mday
1
```

eliasdorneles (changes requested):
Hi @amaarahj !
Thank you for contributing!
Sorry for taking long to review, I've been eyeing this PR for a while but only today I managed to give the attention it deserves.

So, I feel this is definitely going in the right direction!
I've pointed below some things which require some attention.
Please let me know if anything isn't clear.

python/common/python/time.java
amaarahj: I will look into adding this feature.

python/common/python/time.java
amaarahj: No particular reason.

eliasdorneles: Sorry, I don't get it, what was the update?
It's still extending `org.python.types.Module`.

phildini:

PR #786, opened 2018-05-08
Implement dict.keys() and the dict_keys type

Aniket1998:
#38 
Implemented the dict.keys method by adding a dict_keys type and written appropriate test cases for it such that changes in the keys of the original dict reflects in the corresponding dict_keys

Python's dict_keys behaves much like a read only set with operations like >, <, >=, <= etc behaving exactly like that of frozenset the hence the DictKeys class inherits from FrozenSet.

frozenset has some methods such as union, intersection etc. that aren't  present in dict_keys hence those methods are overriden to throw an AttributeError


eliasdorneles:
@Aniket1998 hi there!
Sorry for the delay, this looks great and almost ready to merge!
Can you please solve the conflicts?

Aniket1998:
@eliasdorneles Done, does it work fine now?
Sorry about not looking into the conflict earlier, probably should have combined this and #789  as a single PR instead of submitting both separately.

eliasdorneles:
@Aniket1998 no worries at all!
Thanks for contributing!

PR #791, opened 2018-05-14
Update java install instructions.

debonzi:
The documentation at https://voc.readthedocs.io/en/latest/how-to/contribute.html states that you need `a Java 7 or Java 8 JDK` but the install instructions states `at least 7` which can lead people to install Java 10 JDK.

### DIFF
--- a/docs/background/install.rst
+++ b/docs/background/install.rst
@@ -25,7 +25,7 @@ To check if you have the JDK installed, run ``javac -version``
     $ javac -version
     javac 1.7.0_101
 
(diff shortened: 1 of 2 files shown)


debonzi:
To be honest I have no idea. One reason Java 10 is not in the contribute documentation could be that there are a number of devices not using it but it is just a guess. Or it was not well tested yet with Java 10. Would like to have some input on that from people that are more familiar with it than me.

freakboy3742:
All we really need to verify this is to try and run a hello world app with Java 10 - it's very much a "it will either work, or break" situation. If it works, we can just say "install a >= 7 JVM".

debonzi:
Update done in the proper place.

PR #808, opened 2018-05-14
Unused functions don't break the compilation

Ad115:


voc/python/ast.py
```diff
@@ -1894,6 +1917,16 @@ def compare_to(arg):
 
     @node_visitor
     def visit_Call(self, node):
+        # Check if the callable function contains 
+        # not implemented features, if it does, abort
+        # compilation
+        if node.func.id in self.notimplemented:
+            error = self.notimplemented[node.func.id]
+            raise RuntimeError(
```
eliasdorneles: This is raising an error at compile time when the function is called.

We don't really want to crash the compiler: what we want instead is to crash the program that's being compiled in the case the function is called.

Namely, what you want here is to generate opcodes for raising RuntimeError.
You can find examples of that in this same file if you look for `java.THROW`.

voc/python/ast.py
```diff
@@ -1894,6 +1917,16 @@ def compare_to(arg):
 
     @node_visitor
     def visit_Call(self, node):
+        # Check if the callable function contains 
+        # not implemented features, if it does, abort
+        # compilation
+        if node.func.id in self.notimplemented:
+            error = self.notimplemented[node.func.id]
```
eliasdorneles: To avoid name clashing (e.g. functions in different modules/classes with the same name), it's best to store in this dictionary something that identify the specific node being compiled.
My suggestion is to use `'function-{}-{}'.format(node.func.id, id(node))`.

voc/python/ast.py
```diff
@@ -221,10 +227,27 @@ def visit_FunctionDef(self, node):
...
 
-        for child in node.body:
-            self.visit(child)
-        self.pop_context()
-
+        try:
+            for child in node.body:
+                self.visit(child)
+                
+        except NotImplementedError as e:
```
eliasdorneles: Since this error shouldn't happen at compile, these changes here aren't needed.

voc/python/ast.py
Ad115: Oh, ok, looks like I misinterpreted the intended behavior, so, what we want to do is then to replace the behavior of lines with "yield from" to throw an exception? And wouldn't it be better for them to throw a NotImplementedError exception?

PR #811, opened 2018-05-15
Test isprintable surrogate cases

jonkiparsky:
Changes in this PR:
* Update Java character type for SURROGATE (was 18, s/b 19)
* refactor `_buffer_encode_codepoint` logic
* handle chars in the surrogate range

### DIFF
--- a/python/common/org/python/types/Str.java
+++ b/python/common/org/python/types/Str.java
@@ -77,8 +77,8 @@ public class Str extends org.python.types.Object {
         // 15 - CONTROL
         // 16 - FORMAT
         // 17 - PRIVATE_USE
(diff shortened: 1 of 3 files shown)


tests/test_codecs.py
```diff
@@ -390,7 +390,6 @@ def test_decoder_state(self):
                                          u, u.encode(self.encoding))
 
     def test_lone_surrogates(self):
-        self.assertRaises(UnicodeEncodeError, "\ud800".encode, self.encoding)
```
eliasdorneles: Hello @jonkiparsky !
Can you tell me why is this test being removed?
It looks like this was passing before the changes in this PR.

PR #812, opened 2018-05-15
Add tests for two-arg version of round.

gkhaburzaniya:
`round` has both a single arg and a two-arg version. The two-arg version is not being tested yet, this adds those tests.

Makes this issue show up in tests
https://github.com/pybee/voc/issues/554


gkhaburzaniya:
`test_int_int` is the specific one there.

Issue #763, opened 2018-03-11
[GSoC] Implement asyncio support


BPYap:
# **Project Description**
Python 3.4 introduced asynchronous I/O support to the Python core language. Python asyncio module provides infrastructure for writing single-threaded concurrent code using coroutines, multiplexing I/O access over sockets and other resources, running network clients and servers, and other related primitives. 

This project focus on implementing asyncio’s coroutine and event loop (both timer & socket based) for voc in JVM. In order to achieve the end goal, base infrastructure such as coroutine class and specific statement such as ‘yield from’ would be implemented during first phase of the project and subsequent phases will focus in implementations of Python's asyncio library as Java API.

Successful implementation of asyncio in voc will greatly benefit toga_android development as event loop in Android API can be integrated in toga using the implemented asyncio library (similar to rubicon’s event loop for ios devices). Other than that, applications that utilize concurrent computing can be executed in Java Virtual Machine - users will be able to write concurrent functions (using asyc def/await in Python) and executes them in Java Virtual Machine after transpilation.

Testing, debugging, and documenting are integral parts of the project development. Therefore, they will be carried out continuously as the project progress. The project is broken down into three phases, with end of each phase corresponding to mentor’s evaluation.

#  **Project Details**
1. The following Python statements will be implemented through the ast module (i.e implements the conversion of ast nodes to Java bytecode):
   - `yield from`
   - `async`
   - `await`  

2. Some modules can be directly compiled through voc with some minor adjustments and debugging on voc, hence I will focus on debugging voc transpiler during the weeks allocated for the following modules:
   - `asyncio.base_futures`
   - `asyncio.futures`
   - `asyncio.base_tasks`
   - `asyncio.tasks`

3. The event loop (both timer and socket based) will be rewritten in Java, as the module `base_events.py` depends on too many stdlib modules in which voc currently does not support, hence providing supports for all of them _may_ not be feasible within my allocated timeline. Here's how some functionality in Python can be translated to Java:
   - `collections.deque()` --> Java's `Deque Interface`
   - `time.get_clock_info('monotonic')` --> Java's `System.nanotime()`
   - `socket.py` --> Java.net's `Socket` (prioritizes operations required by event loop, such as setup connection, receiving and sending data operations) **

** Java does not support Unix Domain Socket natively, hence I'm planning to restrict socket's address family to IPv4 and IPv6

# **Working Timeline**
## Phase 1 (May 14 – Jun 16)
**Aim** 
[...]

PR #817, opened 2018-05-17
Added coroutine related exception class [WIP]

BPYap:
Refer #763 week 2 log


freakboy3742:
While these exceptions will be required by the final implementation of asyncio, I don't believe they'll have to be defined in Java; we should be able to use Python definitions.

I'll close this PR for now; if it turns out Java implementations *are* required, we can re-open it later.

BPYap:
Alright :-)

PR #820, opened 2018-05-23
Added yield from support

BPYap:
This commit only added support for `yield from x` and `yield from foo()` by itself (see samples in test),
the assignment statement `x = yield from foo()` is currently not supported as generator.send() is WIP


BPYap:
I realized I messed up the commits, will open another pr with 'clean' commits

cflee:
@BPYap just a note for the future – it's actually possible to "fix" a messed up PR by pushing the fixed commits to the same remote branch on your fork. For example, if you do git rebase or reset and then re-assemble the correct commits into some local branch, then you do a git force push to the same remote branch that was used to create the PR (in this case `yield_from_support`), it will replace the remote branch with your new commits, and the PR will automatically be updated. Might be easier than closing and then re-opening a PR :)

BPYap:
@cflee Didn't know this is possible, will definitely keep this in mind 😃 Thanks!

PR #821, opened 2018-05-23
Added support for Yield from statement

BPYap:
This commit only added support for `yield from x` or `yield from foo()` by itself (see samples in test),
the assignment statement x = yield from foo() is currently not supported as generator.send() is currently WIP

### DIFF
--- a/tests/structures/test_generator.py
+++ b/tests/structures/test_generator.py
@@ -72,24 +72,48 @@ class GeneratorTests(TranspileTestCase):
                 print(i)
             """)
     
-    def test_yield_from_not_used(self):
-        """Yield from is currently not implemented.
-        Unused yield from statements must not break 
-        the program compilation, but only ensue a warning."""
+    def test_simplest_yieldfrom(self):
         self.assertCodeExecution("""
-            
-            def unused():
-                yield from range(5)
+            def gen1():
+                yield 1
+                yield 2
+                yield 3
+                
+            def gen2():
+                yield from gen1()
+                
+            for i in gen2():
+                print(i)
+        """)
 
-            print('Hello, world!')
-            """)
-    
-    @expectedFailure
-    def test_yield_from_used(self):
-        """Yield from is currently not implemented.
-        Yield from statements become NotImplementedErrors at runtime."""
+    def test_yieldfrom_list(self):
         self.assertCodeExecution("""
-            
+            def gen():
+                yield from [1, 2, 3]
+                
+            for i in gen():
+                print(i)
+        """)
+
+    def test_chainning_yieldfrom(self):
+        self.assertCodeExecution("""
+            def gen1():
+                yield 'a'
+                yield 'b'
+                
+            def gen2():
+                yield from gen1()
+                yield 'gen2'
+                
+            def gen3():
+                yield from gen2()
+                
+            for i in gen3():
+                print(i)
+        """)
+    
+    def test_yield_from_used(self):
+        self.assertCodeExecution("""
             def using_yieldfrom():
                 yield from range(5)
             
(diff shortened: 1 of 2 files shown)


tests/structures/test_generator.py
```diff
@@ -72,24 +72,48 @@ def fizz_buzz(self):
...
+                
+            def gen3():
+                yield from gen2()
+                
+            for i in gen3():
+                print(i)
+        """)
     
-    @expectedFailure
     def test_yield_from_used(self):
```
freakboy3742: This is one of those rare situations where a test can be removed. The test itself is a much simpler version of what you've tested thoroughly with your other `yield from` tests.

tests/structures/test_generator.py
```diff
@@ -72,24 +72,48 @@ def fizz_buzz(self):
...
+    def test_yieldfrom_list(self):
+        self.assertCodeExecution("""
+            def gen():
+                yield from [1, 2, 3]
+                
+            for i in gen():
+                print(i)
+        """)
+
+    def test_chainning_yieldfrom(self):
```
freakboy3742: spelling: `chaining`

voc/python/ast.py
```diff
@@ -1747,18 +1748,39 @@ def visit_Yield(self, node):
...
-            java.THROW(
-                'org/python/exceptions/NotImplementedError',
-                ['Ljava/lang/String;',
-                 JavaOpcodes.LDC_W('"yield from" statement not yet implemented')]
-            )
+            loop,
+        )
+        self.context.add_opcodes(
+            TRY(),
+        )
```
freakboy3742: As a style thing, it's preferable to include as many opcodes as possible in a single `add_opcodes` call. The end result is exactly the same in terms of functionality, but there's less code to *read*, which can make the intent of the code easier to interpret for those coming in later.

voc/python/ast.py
```diff
@@ -1712,7 +1712,8 @@ def visit_GeneratorExp(self, node):
 
     @node_visitor
     def visit_Yield(self, node):
-        self.visit(node.value)
+        if hasattr(node, "lineno"):
+            self.visit(node.value)
```
freakboy3742: It isn't clear to me why `lineno` is a relevant attribute here. Is there a type of AST node that *doesn't* have a linen that might be visited? 

At the very least, a comment explaining the need for this branch is required; better still would be a test case that exercises this case (assuming there isn't one already)

BPYap: This is sort of a 'hackish' way to differentiate a programmatically defined yield node from those generated normally by ast.parse. 

Combined with #823, there will be 3 types of yield node passed into visit_Yield method:
1.  Ordinary yield node generated from a line of code (eg `yield x`). In this case we need to visit x (the default implementation).
2. Ordinary yield node generated from a line of code, but without value/operand (eg `yield` or `x=yield`). In this case we push a `org.python.types.NoneType.NONE` object on stack without visiting the none value.
3. Programmatically defined yield node (in line 1778 `self.visit(ast.Yield(None))`) This node is created in visit_YieldFrom and the absent of `lineno` is use to indicate that a value has been pushed on stack hence there is no need to call self.visit(node.value)

Edit: I just remembered it is possible to add custom attribute to a class so I'm thinking of using a flag (i.e `skip_visit`) as custom attribute of the programmatically defined yield node instead of the `lineno` to increase clarity, what do you think?

freakboy3742: Ok - that reasoning makes sense; however, it isn't immediately obvious from the code. Adding a comment explaining what is going on would be helpful. A custom attribute isn't necessary; as long as the code explains what it's doing, that will be sufficient.

PR #823, opened 2018-05-25
Added generator send method

BPYap:
This PR enables `yield` keyword to be used in expression and introduces send method to generator in voc.

Some examples:
1.  Echoes what is sent
```
def gen():
    # echoes what is sent in
    while True:
        x = print((yield)) 

g = gen()
next(g)
g.send("Hi there!")
g.send("Sup")
```
Output:
```
Hi there!
Sup
```

2. Yield square of value sent
```
def gen():
    # yield square of value sent
    while True:
        x = yield("ready") 
        yield x ** 2

g = gen()
next(g)
print(g.send(2))
next(g)
print(g.send(-1))
```
Output:
```
4
1
```

Note:

- I'm currently working on augmented assignment with `yield`, hence the code below will not work at the moment:
```
def gen():
    k = 123
    k += yield
```

- Once this PR and #821 are confirmed stable, I will implement the same features to `yield from` keyword

### DIFF
--- a/tests/structures/test_generator.py
+++ b/tests/structures/test_generator.py
@@ -96,3 +96,64 @@ class GeneratorTests(TranspileTestCase):
             for i in using_yieldfrom():
                 print(i)
             """)
+            
+    def test_generator_send(self):
+        self.assertCodeExecution("""
+            def gen():
+                a = yield
+                print(a)
+                
+            g = gen()
+            next(g)
+            try:
+                g.send("Hello World")
+            except StopIteration:
+                pass
+            """)
+            
+    def test_generator_multi_send(self):
+        self.assertCodeExecution("""
+            def gen():
+                a = yield 1
+                print(a)
+                b = yield 2
+                print(b)
+                
+            g = gen()
+            next(g)
+            try:
+                print(g.send("a"))
+                print(g.send("b"))
+            except StopIteration:
+                pass
+            """)
+            
+    def test_generator_power_generator(self):
+        self.assertCodeExecution("""
+            def gen():
+                while True:
+                    x = yield 1
+                    yield x ** 2
+                
+            g = gen()
+            next(g)
+            print(g.send(6))
+            next(g)
+            print(g.send(-1))
+            """)
+            
+    def test_generator_send_loop(self):
+        self.assertCodeExecution("""
+            def gen():
+                for i in range(1, 5):
+                    a = yield i
+                    print("printing from generator " + str(a))
+            g = gen()
+            g.send(None)
+            try:
+                while True:
+                    b = g.send(1)
+                    print("printing from user " + str(b))
+            except StopIteration: 
+                pass
+            """)
--- a/tests/structures/test_generator.py
+++ b/tests/structures/test_generator.py
@@ -71,38 +71,38 @@ class GeneratorTests(TranspileTestCase):
             for i in myinterview.fizz_buzz():
                 print(i)
             """)
-    
+
     def test_yield_from_not_used(self):
         """Yield from is currently not implemented.
-        Unused yield from statements must not break 
+        Unused yield from statements must not break
         the program compilation, but only ensue a warning."""
         self.assertCodeExecution("""
-            
+
             def unused():
                 yield from range(5)
 
             print('Hello, world!')
             """)
-    
+
     @expectedFailure
     def test_yield_from_used(self):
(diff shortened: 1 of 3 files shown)


tests/structures/test_generator.py
```diff
@@ -71,28 +71,158 @@ def fizz_buzz(self):
...
-        Unused yield from statements must not break 
+        Unused yield from statements must not break
         the program compilation, but only ensue a warning."""
         self.assertCodeExecution("""
-            
             def unused():
                 yield from range(5)
 
-            print('Hello, world!')
+                print('Hello, world!')
```
freakboy3742: The (lack of) indentation on this line was intentional. The test needs *some* output to verify that *something* ran. The dummy "Hello world" is proof that the code ran. If it's in the `unused()` method, but `unused()` isn't invoked, then the print will never happen. 

This entire test `test_yield_from_not_used` case will also become redundant as soon as `yield from` exists; the purpose of this test is to validate that just having `yield from` in your code won't cause a problem, as long as the yield from isn't actually on a live code path.

tests/structures/test_generator.py
```diff
@@ -71,28 +71,158 @@ def fizz_buzz(self):
...
+            except StopIteration:
+                pass
+            """)
+
+    def test_generator_power_generator(self):
+        self.assertCodeExecution("""
+            def gen():
+                while True:
+                    x = yield 1
+                    yield x ** 2
```
freakboy3742: Is there any particular reason why a "power" operator is being used here? Is it just a mathematical operator to modify the value a bit? In terms of test naming, it seems like there is some sort of deeper significance to the choice of power specifically.

voc/python/ast.py
```diff
@@ -325,8 +325,79 @@ def visit_Delete(self, node):
 
     @node_visitor
     def visit_Assign(self, node):
-        # Evaluate the value
-        self.visit(node.value)
+        def push_message(node):
```
freakboy3742: This shouldn't be an closure method - it's not dependent on any of the context provided by `visit_Assign`. You can move it into a utility method at the level of the visitor (like e.g., `full_classspec`).

voc/python/ast.py
```diff
@@ -325,8 +325,79 @@ def visit_Delete(self, node):
...
+                    'org/python/types/Generator',
+                    'reset_message',
+                    args=[],
+                    returns='V'
+                )
+            )
+
+        # walk subtree with node.value as root to find any yield node
+        nested_yield_node_found = False
+        for descendant_node in ast.walk(node.value):
```
freakboy3742: Is there a reason that this is handled as a separate walk, rather than just visiting `node.value` within the existing AST visitor that we're evaluating?

voc/python/ast.py
```diff
@@ -1712,13 +1783,25 @@ def visit_GeneratorExp(self, node):
...
     def visit_Yield(self, node):
-        self.visit(node.value)
+        if node.value:
+            self.visit(node.value)
+            self.context.add_opcodes(
+                # Convert to a new value for return purposes
+                JavaOpcodes.INVOKEINTERFACE('org/python/Object', 'byValue', args=[], returns='Lorg/python/Object;')
+            )
+        elif hasattr(node, "lineno"):  # a yield expression
+            # push NoneType object to stack
```
freakboy3742: It's not clear to me why a None is needed on the stack here - is something consuming it later?

(To that end, comments like "push None onto stack" aren't that helpful - it's fairly easy to see that the next line is "add none to stack"; the comment needs to say *why* that is the required course of action)

voc/python/ast.py
BPYap: Yes. Completely agree with that, I realized expression is more than just assignment and augmented assignment. I'm currently working on turning the block of code into an utility function which can be called in any node that evaluate an expression.

tests/structures/test_generator.py
BPYap: I noticed pow is handled differently in `visit_BinOp` so I added this test case to see its behavior, it executes as intended and I have merged this test case under `test_generator_yield_expr_binary_op`

voc/python/ast.py
BPYap: This is to handle the statement `yield` which is equivalent to `yield None` . I have added a comment to it in latest PR

BPYap:
@freakboy3742 Conflicts resolved as per request :-)

PR #824, opened 2018-05-29
Introduce performance benchmarking suite 

patiences:
In the coming weeks, as part of my GSoC project, there will be changes that aim to improve the performance of the code and so we need a way to evaluate that. This is a first PR with some example tests. 
This test is not intended to be run like the other unit tests, rather it can be run from the command line and produces an output like 
```
Running test_small_integers ...
  Elapsed time:  37.10159900219878  ms
  CPU process time:  274.72000000000025  ms
Running test_booleans ...
  Elapsed time:  18.298730996320955  ms
  CPU process time:  181.10000000000014  ms
```

### DIFF
--- a/tests/bench_datatypes.py
+++ b/tests/bench_datatypes.py
@@ -1,25 +1,27 @@
 import sys
 from os import path
-sys.path.append( path.dirname( path.dirname( path.abspath(__file__) ) ) )
+sys.path.append(path.dirname(path.dirname(path.abspath(__file__))))
 
 from utils import *
 
 def test_small_integers(test_case):
-    test_case.runAndBenchAsJava("test_small_integers", """
+    print("Running", "test_small_integers")
+    test_case.runAsJava(adjust("""
         for i in range(10000):
             x = -5
             for i in range(-5, 257):
                 x = x + 1
-        """)
+        """), timed=True)
 
 def test_booleans(test_case):
-    test_case.runAndBenchAsJava("test_booleans", """
+    print("Running", "test_booleans")
+    test_case.runAsJava(adjust("""
         x = True
         y = False
         for i in range(10000):
             x = (not x and y)
             y = x != y
-        """)
+        """), timed=True)
 
 def main():
     test_case = TranspileTestCase()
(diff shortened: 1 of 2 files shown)


tests/utils.py
```diff
@@ -534,20 +534,27 @@ def makeTempDir(self):
...
+
             transpiler.transpile_string("test.py", main_code)
 
             if extra_code:
                 for name, code in extra_code.items():
                     transpiler.transpile_string("%s.py" % name.replace('.', os.path.sep), adjust(code))
 
+            t1_stop = time.perf_counter()
+            t2_stop = time.process_time()
+
```
freakboy3742: This won't be benchmarking what you think it is (or, at least, it's not benchmarking the thing that needs to be benchmarked. These timing calls are wrapped around the transpilation process - which is the process of converting Python code to Java bytecode. While this is something that certainly *can* be benchmarked, it's not the thing that is a performance concern - it's *runtime* performance that is a problem. 

So - we need the start and end timers to be executed *in Java*, so that we're performance testing the *execution* of the code, not the *compilation*. 

The simplest version of this would be to wrap the timer calls around the call to subprocess (since that's the part that spawns the actual invocation).

tests/bench_datatypes.py
```diff
@@ -0,0 +1,31 @@
+import sys
+from os import path
+sys.path.append( path.dirname( path.dirname( path.abspath(__file__) ) ) )
```
freakboy3742: The spacing on this is a little odd.

tests/utils.py
```diff
@@ -583,8 +590,16 @@ def runAsJava(self, main_code, extra_code=None, args=None):
             )
             out = proc.communicate()[0].decode('utf8')
 
+        if timed:
+            print("  Elapsed time: ", (t1_stop-t1_start) * 1000, " ms")
+            print("  CPU process time: ", (t2_stop-t2_start) * 10000, " ms")
+
         return out
 
+    def runAndBenchAsJava(self, test_name, code):
```
freakboy3742: I can see what you're doing here, but it's probably better served as 2 lines of code in the external wrapper invoking the code, rather than something on the utility class.

tests/utils.py
patiences: Ah-ha.. I see. Thanks! Done

tests/bench_datatypes.py
patiences: Oops, fixed. Done

tests/utils.py
patiences: Done

PR #825, opened 2018-05-29
Small integer preallocation optimization 

patiences:
This PR implements the preallocation and initialization of the integers [-5, 257) so that they can be reused. In general, this means that when the python code is parsed and an integer object is detected, if it is in this range we will not create a new Integer object but rather use the preallocated one. 

Object construction is expensive memory and time-wise. We can see a performance gain even running a simple performance test like [this one](https://github.com/pybee/voc/pull/824/commits/a89fc6617b582c3b5981df88fb4a2d9c2376edeb): 

Pre-optimization 
```
Running test_small_integers ...
  Elapsed time:  33.07889499713201  ms
  CPU process time:  289.66  ms
```
With optimization 
```
Running test_small_integers ...
  Elapsed time:  23.559533001796808  ms
  CPU process time:  240.0899999999998  ms
```
(Of course there is always some variance between different runs, but from what I've seen there is consistent improvement.)

### DIFF
--- a/voc/python/blocks.py
+++ b/voc/python/blocks.py
@@ -196,11 +196,14 @@ class Block(Accumulator):
                 )
             else:
                 if isinstance(value, bool):
-                    self.add_opcodes(
-                        java.New('org/python/types/Bool'),
-                        ICONST_val(value),
-                        java.Init('org/python/types/Bool', 'Z'),
-                    )
+                    if value is True:
+                        self.add_opcodes(
+                            JavaOpcodes.GETSTATIC('org/python/types/Bool', 'TRUE', 'Lorg/python/Object;'),
+                        )
+                    else:
+                        self.add_opcodes(
+                            JavaOpcodes.GETSTATIC('org/python/types/Bool', 'FALSE', 'Lorg/python/Object;'),
+                        )
 
                 elif isinstance(value, int):
                     self.add_int(value)
--- a/voc/python/blocks.py
+++ b/voc/python/blocks.py
@@ -198,11 +198,11 @@ class Block(Accumulator):
                 if isinstance(value, bool):
                     if value is True:
                         self.add_opcodes(
-                            JavaOpcodes.GETSTATIC('org/python/types/Bool', 'TRUE', 'Lorg/python/Object;'),
+                            JavaOpcodes.GETSTATIC('org/python/types/Bool', 'TRUE', 'Lorg/python/types/Bool;'),
                         )
                     else:
                         self.add_opcodes(
-                            JavaOpcodes.GETSTATIC('org/python/types/Bool', 'FALSE', 'Lorg/python/Object;'),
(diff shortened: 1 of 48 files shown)


patiences:
As a further piece of work I think we should consider restricting access to the `Int` constructor altogether. This is perhaps a small issue on the transpilation side, for example, when we have an instruction that creates an integer from a value inside the variable so we're not able to check if the integer is small and generate lookup instructions. However, the math operations on `Int` objects create new `Int` objects all the time. It should be fairly straightforward to refactor the code to check for small integers, if it's something we want to do.

patiences:
Some updated numbers: 

Running test_small_integers
  Elapsed time:  122.01105412701145  sec
  CPU process time:  0.0006999999999999784  sec

Running test_small_integers
  Elapsed time:  128.66034104398568  sec
  CPU process time:  0.0008010000000000517  sec

Running test_small_integers
  Elapsed time:  122.50824150099652  sec
  CPU process time:  0.00048599999999998644  sec

Running test_small_integers
  Elapsed time:  122.89947666198714  sec
  CPU process time:  0.0008429999999999827  sec

Running test_small_integers
  Elapsed time:  122.97956180298934  sec
  CPU process time:  0.00045000000000000595  sec

freakboy3742:
@patiences Do you have baseline numbers for the benchmark? The ticket only has the "progress" numbers...

patiences:
@freakboy3742  Yes, of course! Here we go: 

**Pre-optimization** 
Running test_small_integers
  Elapsed time:  144.74749750000774  sec
  CPU process time:  0.00047200000000002795  sec

Running test_small_integers
  Elapsed time:  143.55906505999155  sec
  CPU process time:  0.0010949999999999571  sec

Running test_small_integers
  Elapsed time:  149.2436827890051  sec
  CPU process time:  0.0008159999999999834  sec

Running test_small_integers
  Elapsed time:  146.150813241984  sec
  CPU process time:  0.0008259999999999934  sec

Looks like about a 15% improvement.

patiences:
> python/testfile.class has also been inadvertently added.

>That file needs to be stripped out (ideally, rolling back the commit where it was added; if you just >delete the file, it will still be in the commit history, and so every person who clones the repo will >download the file, even though it doesn't exist any more).

Really? That is so weird. Is that the case with .py files too (some of those got added and deleted at other points too, oops)? I actually am not sure how to revert just the specific commit (bb40b5b) -- I tried to rebase it without that commit locally but I'm not sure that's going to work...

voc/python/blocks.py
```diff
@@ -16,6 +16,10 @@
 
 # from .debug import DEBUG, DEBUG_value
 
+# The number of negative and positive small org/python/types/Int objects that are preallocated
+# [-NSMALLNEGINTS, NSMALLPOSINTS)
+NSMALLNEGINTS = 5
+NSMALLPOSINTS = 257
```
freakboy3742: I don't think this is needed anymore

PR #826, opened 2018-06-01
Fix test_set issues for test suite

AmirYalamov:
test_tuple was giving "unexpected success" when the tests were ran in the terminal.
Line entered: `$ python setup.py test -s tests.builtins.test_set.BuiltinSetFunctionTests`
Line yielded: `test_tuple (tests.builtins.test_set.BuiltinSetFunctionTests) ... unexpected success`.
I deleted the `not_implemented` reference for test_tuple from the file "test_set".

My second commit addresses the issue that there was no error handling for the assert statement in utils.py. The line `self.assertEqual(java_out, py_out, context)` showed up to yield a failure in the smoke-test of my first commit.

Third commit added the output for CPython from the CI server's error log in the `test_tuple` file which was `+ {1.2, True, 3}`, so I added it as `substitutions = { (1.2, True, 3  }`. This was done since CPython is the original implementation of Python, so perhaps I need to go off its output.

### DIFF
--- a/tests/datatypes/test_tuple.py
+++ b/tests/datatypes/test_tuple.py
@@ -272,3 +272,7 @@ class BinaryTupleOperationTests(BinaryOperationTestCase, TranspileTestCase):
 
 class InplaceTupleOperationTests(InplaceOperationTestCase, TranspileTestCase):
     data_type = 'tuple'
+
+    substitutions = {
+        (1.2, True, 3)
+    }
--- a/tests/datatypes/test_tuple.py
+++ b/tests/datatypes/test_tuple.py
@@ -272,3 +272,10 @@ class BinaryTupleOperationTests(BinaryOperationTestCase, TranspileTestCase):
 
 class InplaceTupleOperationTests(InplaceOperationTestCase, TranspileTestCase):
     data_type = 'tuple'
+
+    substitutions = {
+        # output, keyed to all possible inputs
+        "{3, 1.2, True}": [
+            "{1.2, 3, True}", "{True, 1.2, 3}", "{True, 3, 1.2}", "{3, 1.2, True}", "{3, True, 1.2}", "{1.2, True, 3}",
+        ]
+    }
--- a/tests/datatypes/test_tuple.py
+++ b/tests/datatypes/test_tuple.py
@@ -274,5 +274,5 @@ class InplaceTupleOperationTests(InplaceOperationTestCase, TranspileTestCase):
     data_type = 'tuple'
 
     substitutions = {
-        (1.2, True, 3)
+        1.2, True, 3
     }
--- a/tests/datatypes/test_tuple.py
+++ b/tests/datatypes/test_tuple.py
@@ -275,19 +275,19 @@ class InplaceTupleOperationTests(InplaceOperationTestCase, TranspileTestCase):
 
     substitutions = {
         # all possible outputs
-        "(1.2, 3, True)": [
-            "(1.2, True, 3)"
+        "{1.2, 3, True}": [
+            "{1.2, True, 3}"
         ],
-        "(3, 1.2, True)": [
-            "(1.2, True, 3)"
+        "{3, 1.2, True}": [
+            "{1.2, True, 3}"
         ],
-        "(3, True, 1.2)": [
-            "(1.2, True, 3)"
+        "{3, True, 1.2}": [
+            "{1.2, True, 3}"
         ],
-        "(True, 3, 1.2)": [
-            "(1.2, True, 3)"
+        "{True, 3, 1.2}": [
+            "{1.2, True, 3}"
         ],
-        "(True, 1.2, 3)": [
-            "(1.2, True, 3)"
+        "{True, 1.2, 3}": [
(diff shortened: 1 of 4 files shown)


tests/utils.py
```diff
@@ -404,12 +404,15 @@ def assertCodeExecution(
...
-            self.assertEqual(java_out, py_out, context)
-
+            try:
+                if message:
+                    context = 'Global context: %s' % message
+                else:
+                    context = 'Global context'
+                self.assertEqual(java_out, py_out, context)
+            except AssertionError as e:
+                self.fail(e)
```
freakboy3742: It's not entirely clear what you're trying to do with this change - you're effectively catching an AssertionError so you can raise another AssertionError (which is all `self.fail()` does).

The error you're seeing is going to need to use the `substitutions` feature of the VOC test suite; if you search other tests, you should be able to find an example of the use of this feature.

AmirYalamov: Ok, I see the examples of the `substitutions` feature. As for the values, is there a standard set of tuples that is used for testing, or would I be making up random tuples for the `UnaryTupleOperationTests`, `BinaryTupleOperationTests`, and `InplaceTupleOperationTests`?

freakboy3742: Yes, there is a standard set of test data - that's what you can see in the test output. The failure output shows you the example that was tried, and the output that was received (the `-` line is what VOC produced, and the `+` line is what CPython produced). If you define a substitution that neutralises that difference, the test should pass.

freakboy3742:
@AmirYalamov You're on the right path, but you've missed something from the existing examples. 

`substitutions` isn't just a *set* of patterns - it's a *dictionary* of "preferred forms" and "possible forms". So, for example `datatypes/test_float.py` contains the following: 
```
    substitutions = {
        "(-0.8946025309573877+0.446862743585361j)": [
            "(-0.8946025309573877+0.44686274358536093j)"
        ]
```
What this is telling the test case is that every time it sees any of the strings in the list, it should be replaced with the version from the dictionary key (in this case - if you get a long complex number ending in `6093j`, it will be replaced by the same complex number ending in `6j`) This normalises the output so that a simple comparison will works.

In your case, you have a test that can produce `{1.2, True, 3}` - but because sets are unordered, it could also be presented as `{3, 1.2, True}`, or `{True, 1.2, 3}`, or 4 other possiblities. You need to capture the various possible outputs as substitutions, and normalize them to a common ordering. It doesn't matter which one - it just needs to be consistent.

tests/datatypes/test_tuple.py
```diff
@@ -272,3 +272,22 @@ class BinaryTupleOperationTests(BinaryOperationTestCase, TranspileTestCase):
...
+        ],
+        "{3, True, 1.2}": [
+            "{1.2, True, 3}"
+        ],
+        "{True, 3, 1.2}": [
+            "{1.2, True, 3}"
+        ],
+        "{True, 1.2, 3}": [
+            "{1.2, True, 3}"
+        ],
```
freakboy3742: Closer, but we're still not there. 

The format is *output*: *input list*. We want a *single* output, keyed to *all* possible inputs - the idea is that *any* ordered input will be mapped to a single, canonical ordering.

AmirYalamov: Is my formatting all correct? Am I supposed to remove the quotation marks?

Also, should I be adding a `substitutions` feature into `class UnaryTupleOperationTests` and `class BinaryTupleOperationTests`?

freakboy3742: The quotation marks are required; the substitutions are being done on the string output.

You *don't* need to add this to UnaryTupleOperationTests and class BinaryTupleOperationTests. However, you do need to  make the change to the test that is actually raising the error. The test failure is with BuiltinSetFunctionTests; that's the only class that needs substitutions. You're adding a substitution definition to `datatypes/test_tuple.py`; the error is being raised by `builtins/test_tuple.py`

You also need to revert the indentation change you've made to `tests/builtins/test_tuple.py`.

AmirYalamov:
@freakboy3742 How long does the CI server usually take to do a complete check? I submitted my last commit 2 days ago and it still has not completed the tests. Is it possible it is stuck in an infinite loop?

cflee:
@AmirYalamov I've just restarted the CI build! I think it's under an hour nowadays, so check back in a while.

freakboy3742:
@AmirYalamov Again - you need to read *all* the error message* (and, *all* my comments). The error is with the `test_tuple` test - in the `test_set.py` file. A change won't have any effect if you're making it in the wrong place.

(I'll also note that your substitution list includes the target pattern in the list of patterns you want to match - which, while it isn't an error, is redundant; and the patch still contains a change to `test/utils.py` that isn't required.

freakboy3742:
@AmirYalamov Ok, so we're getting close now. There's three files that contain changes that aren't needed - including one that I've specifically flagged as part of a review. If you can revert those changes, we should be good to go!

AmirYalamov:
@freakboy3742 Thanks for all the help along! Are there any particular Issues that you would recommend I try to tackle next?

PR #829, opened 2018-06-04
Use python requires to require Python 3.4

danyeaw:
Toga was recently updated to include python_requires. This PR updates VOC as well to enforce python 3.4 using the new python_requires syntax. Also updated the documentation to make it more clear that only Python 3.4 is currently supported.

Signed-off-by: Dan Yeaw <dan@yeaw.me>

### DIFF
--- a/docs/background/faq.rst
+++ b/docs/background/faq.rst
@@ -82,7 +82,7 @@ that caused the problem.
 What version of Python does VOC require?
 ----------------------------------------
 
-VOC runs under Python 3.4.
+VOC runs under Python 3.4+.
 
 What version of Java does VOC require?
 --------------------------------------
--- a/docs/background/faq.rst
+++ b/docs/background/faq.rst
@@ -82,7 +82,7 @@ that caused the problem.
 What version of Python does VOC require?
 ----------------------------------------
 
-VOC runs under Python 3.4+.
+VOC runs under Python 3.4.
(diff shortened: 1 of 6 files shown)


docs/background/faq.rst
```diff
@@ -82,7 +82,7 @@ that caused the problem.
 What version of Python does VOC require?
 ----------------------------------------
 
-VOC runs under Python 3.4+.
+VOC runs under Python 3.4.
```
eliasdorneles: So, even if you can't yet transpile code from 3.5+, you can still run VOC with Python 3.4+.

setup.py
```diff
@@ -28,6 +24,7 @@
     author_email='russell@keith-magee.com',
     url='http://pybee.org/voc',
     packages=find_packages(exclude=['docs', 'tests']),
+    python_requires='==3.4',
```
eliasdorneles: I think this should probably be `>=3.4`, because this is about the Python version you're running VOC, right?

danyeaw:
@eliasdorneles Thanks for taking the time to review and provide me feedback. I thought the requirements for using Python 3.4 were more strict. I have updated the python_requires to be >= 3.4 and removed the documentation changes.

eliasdorneles:
Thanks Dan ! :1st_place_medal:

PR #830, opened 2018-06-04
Boolean preallocation optimization

patiences:
This PR improves execution time by using preallocated instances of True and False, never creating new `Boolean` objects. 

Performance benchmark stats ([test here](https://github.com/patiences/voc/commit/f161e5c97aaa5dab5c8bfba7ed1d7e150f596c89)): 

**Pre-change** 
Running test_booleans
  Elapsed time:  85.63200672800303  sec
  CPU process time:  0.0009250000000000091  sec

Running test_booleans
  Elapsed time:  83.78228437301004  sec
  CPU process time:  0.00013499999999999623  sec

Running test_booleans
  Elapsed time:  80.98856862200773  sec
  CPU process time:  0.00012000000000000899  sec

**With this change** 
Running test_booleans
  Elapsed time:  42.00830070799566  sec
  CPU process time:  0.0008920000000000039  sec

Running test_booleans
  Elapsed time:  41.94127046901849  sec
  CPU process time:  0.0007800000000000029  sec

Running test_booleans
  Elapsed time:  43.51021371400566  sec
  CPU process time:  0.0009509999999999796  sec

### DIFF
--- a/python/common/org/Python.java
+++ b/python/common/org/Python.java
@@ -384,12 +384,12 @@ public class Python {
             while (true) {
                 org.python.Object next = iter.__next__();
                 if (!next.toBoolean()) {
-                    return new org.python.types.Bool(false);
+                    return org.python.types.Bool.getBool(false);
                 }
             }
         } catch (org.python.exceptions.StopIteration si) {
         }
-        return new org.python.types.Bool(true);
+        return org.python.types.Bool.getBool(true);
     }
 
     @org.python.Method(
@@ -405,12 +405,12 @@ public class Python {
             while (true) {
                 org.python.Object next = iter.__next__();
                 if (next.toBoolean()) {
-                    return new org.python.types.Bool(true);
+                    return org.python.types.Bool.getBool(true);
                 }
             }
         } catch (org.python.exceptions.StopIteration si) {
         }
-        return new org.python.types.Bool(false);
+        return org.python.types.Bool.getBool(false);
     }
 
     @org.python.Method(
@@ -462,7 +462,7 @@ public class Python {
             args = {"object"}
     )
     public static org.python.types.Bool callable(org.python.Object object) {
-        return new org.python.types.Bool(org.python.Callable.class.isAssignableFrom(object.getClass()));
+        return org.python.types.Bool.getBool(org.python.Callable.class.isAssignableFrom(object.getClass()));
     }
 
     @org.python.Method(
@@ -777,9 +777,9 @@ public class Python {
         }
         try {
             object.__getattribute__(name);
-            return new org.python.types.Bool(true);
+            return org.python.types.Bool.getBool(true);
         } catch (org.python.exceptions.AttributeError ae) {
-            return new org.python.types.Bool(false);
+            return org.python.types.Bool.getBool(false);
         } catch (org.python.exceptions.TypeError te) {
             throw new org.python.exceptions.TypeError(te.getMessage().replace("__hasattribute__", "hasattr"));
         }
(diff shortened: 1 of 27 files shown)


python/common/org/Python.java
```diff
@@ -384,12 +384,12 @@ public static void initializeModule(java.lang.Class cls, java.util.Map<java.lang
             while (true) {
                 org.python.Object next = iter.__next__();
                 if (!next.toBoolean()) {
-                    return new org.python.types.Bool(false);
+                    return org.python.types.Bool.getBool(false);
```
freakboy3742: In the case where you're referencing a constant, is there any reason to not just use the static member reference directly? Variable access should always be faster than a function call...

patiences: Ah, yes! Done (Here and many other places)

patiences:
With this change, things seem to be marginally better: 

Running test_booleans
  Elapsed time:  41.18798470598995  sec
  CPU process time:  0.00010700000000002374  sec

Running test_booleans
  Elapsed time:  41.623120495001785  sec
  CPU process time:  0.00012199999999995548  sec

Running test_booleans
  Elapsed time:  41.56958808301715  sec
  CPU process time:  0.000154000000000043  sec

Running test_booleans
  Elapsed time:  43.66924121600459  sec
  CPU process time:  0.00013999999999997348  sec

python/common/org/python/types/Bool.java
```diff
@@ -91,9 +100,9 @@ public Bool(org.python.Object[] args, java.util.Map<java.lang.String, org.python
     )
     public org.python.Object __lt__(org.python.Object other) {
         if (other instanceof org.python.types.Int) {
-            return new org.python.types.Bool((((org.python.types.Bool) this).value ? 1 : 0) < ((org.python.types.Int) other).value);
+            return org.python.types.Bool.getBool((((org.python.types.Bool) this).value ? 1 : 0) < ((org.python.types.Int) other).value);
         } else if (other instanceof org.python.types.Bool) {
-            return new org.python.types.Bool((((org.python.types.Bool) this).value ? 1 : 0) < (((org.python.types.Bool) other).value ? 1 : 0));
+            return org.python.types.Bool.getBool((((org.python.types.Bool) this).value ? 1 : 0) < (((org.python.types.Bool) other).value ? 1 : 0));
```
freakboy3742: Another tweak, but this expression can be simplified as well; if we're comparing two bools, there's no need to do a transition through integer logic to get back to a boolean.

python/common/org/python/types/Bool.java
```diff
@@ -302,7 +311,7 @@ public boolean __setattr_null(java.lang.String name, org.python.Object value) {
                 return new org.python.types.Int(0);
             } else {
                 if (org.Python.VERSION < 0x03060000) {
-                    return new org.python.types.Bool(false);
+                    return (org.python.types.Bool) FALSE;
```
freakboy3742: The cast shouldn't be needed here (and the constant should be fully qualified to avoid ambiguity)

python/common/org/python/types/ByteArray.java
```diff
@@ -413,33 +413,33 @@ public ByteArray(org.python.Object[] args, java.util.Map<java.lang.String, org.p
             byte[] other_bytes = (byte[]) ((org.python.types.Bytes) other).value;
             for (int i = 0; i < Math.min(this.value.length, other_bytes.length); i++) {
                 if (this.value[i] > other_bytes[i]) {
-                    return new org.python.types.Bool(1);
+                    return org.python.types.Bool.getBool(1);
```
freakboy3742: `getBool(1)` should have a very predictable output... :-)

python/common/org/python/types/Bool.java
patiences: Done!

python/common/org/python/types/ByteArray.java
patiences: Done, here and in other places :-)

python/common/org/python/types/Bool.java
patiences: I'm not sure we can use `<` directly on booleans here as that's not defined in Java, but other places in the file like `==` or `|` have been tweaked!

Issue #827, opened 2018-06-01
Maps reserved Java keywords to Python built-in function/method call


BPYap:
Some built-in methods in Python has names that conflict with reserved keyword in Java. One example is Python's generator `throw` method vs Java's `throw` syntax.

The current approach is by renaming the method to something else in Java implementation and adding checks to convert the name during Attribute node traversal in `ast.py`. In the case of generator `throw` method, the check below is added in `visit_Attribute`:
```
if node.attr == "throw":
    node.attr = "throw_"  # refer org.python.types.Generator._throw
``` 
The `throw` will be converted to `throw_` method name defined in Generator.java during transpilation.

As you can see, it is not very 'scalable', plus it lacks mechanism to standardize naming convention for converted name. 

I spoken with @freakboy3742 just now and he suggested to have a dictionary to store mappings of reserved Java syntax to standardized Python's call name so that the transpiler can refer to it during transpilation.

BPYap:
Closing this. This can be achieved by defining `name` field in `@org.python.Method` decorator.
See https://voc.readthedocs.io/en/latest/reference/signatures.html

PR #831, opened 2018-06-04
Support exceptions handling in generator [WIP]

BPYap:
This PRs include:
- Allows yield to be used in try-finally suite
- Allows return statement to be used in a generator
- Implements Generator's throw method**
- Implements Generator's close method
- Implements Generator's __del__ method

**Currently working on parsing *args and **kwargs of throw method

### DIFF
--- a/python/common/org/python/types/Generator.java
+++ b/python/common/org/python/types/Generator.java
@@ -33,6 +33,11 @@ public class Generator extends org.python.types.Object {
         this.message = org.python.types.NoneType.NONE;
     }
 
+    protected void finalize() throws Throwable {
+        super.finalize();
+        this.__del__();
+    }
+
     public void yield(java.util.Map<java.lang.String, org.python.Object> stack, int yield_point) {
         // System.out.println("YIELD: " + yield_point);
         // for (org.python.Object obj: stack) {
@@ -102,16 +107,20 @@ public class Generator extends org.python.types.Object {
         }
 
         if (just_started) {
-            // TODO: close the generator before throwing exception
+            this.close();
             throw (org.python.exceptions.BaseException) this.exception;
         }
 
         try {
             return this.__next__();
         } catch (org.python.exceptions.BaseException e) {
-            this.close(); // close this generator if it did not catch the exception
-            throw e; // re-throw exception after closing
+            if (!(e instanceof org.python.exceptions.StopIteration)) {
+                this.cleanup(); // close this generator if it did not catch the exception
+                throw e; // re-throw exception after closing
+            }
         }
+
+        return org.python.types.NoneType.NONE;
     }
 
     public void throw_exception() {
@@ -147,8 +156,14 @@ public class Generator extends org.python.types.Object {
             throw new org.python.exceptions.RuntimeError("generator ignored GeneratorExit");
         }
 
-        this.expression = null; // cleanup
+        this.cleanup();
         return org.python.types.NoneType.NONE;
+    }
+
+    private void cleanup() {
+        this.expression = null;
+        this.message = null;
+        this.exception = null;
     }
 
     @org.python.Method(
--- a/python/common/org/python/types/Generator.java
+++ b/python/common/org/python/types/Generator.java
@@ -1,5 +1,8 @@
 package org.python.types;
 
+import java.lang.reflect.Constructor;
+import java.lang.reflect.InvocationTargetException;
+
 public class Generator extends org.python.types.Object {
     java.lang.String name;
     java.lang.reflect.Method expression;
@@ -8,6 +11,7 @@ public class Generator extends org.python.types.Object {
 
     private boolean just_started = true;
     public org.python.Object message;
+    public org.python.Object exception;
 
     public int hashCode() {
         return this.expression.hashCode();
@@ -26,7 +30,7 @@ public class Generator extends org.python.types.Object {
         this.expression = expression;
         this.yield_point = 0;
         this.stack = stack;
-        this.message = new org.python.types.NoneType();
+        this.message = org.python.types.NoneType.NONE;
     }
 
     public void yield(java.util.Map<java.lang.String, org.python.Object> stack, int yield_point) {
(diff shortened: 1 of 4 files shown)


python/common/org/python/types/Generator.java
```diff
@@ -26,7 +30,12 @@ public Generator(
         this.expression = expression;
         this.yield_point = 0;
         this.stack = stack;
-        this.message = new org.python.types.NoneType();
+        this.message = org.python.types.NoneType.NONE;
```
freakboy3742: We're acquiring a lot of "state" variables in the Generator; I can see that they're all needed (with the possible exception of just_started), but it would be helpful to have some explanatory notes of what the values all mean in the generator context. (e.g., `expression == null` means the generator has exhausted itself). A class-level documentation block explaining all the possible "states" of a generator is all we'd need here.

tests/structures/test_generator.py
```diff
@@ -231,28 +247,246 @@ def gen():
...
+    def test_generator_throw_on_starting(self):
+        self.assertCodeExecution("""
+            def gen():
+                yield "Hello World"
+
+            g = gen()
+            try:
+                g.throw(ZeroDivisionError)
+            except ZeroDivisionError:
+                pass
```
freakboy3742: The VOC test suite is based on comparing output, so having tests that "pass" and do nothing in a branch is potentially risky. Ideally, you want to print something that confirms this branch was actually executed (and was executed in the right order); you also want to have print statements in places that *can't* be reached (e.g., after the throw statement) to confirm that code that *should* be dead is actually dead.

voc/python/ast.py
```diff
@@ -694,7 +712,7 @@ def visit_Try(self, node):
         # Finally content is duplicated at the end of the body
         # if it is defined.
         if node.finalbody:
-            for child in node.finalbody:
+            for child in copy.deepcopy(node.finalbody):
```
freakboy3742: Why the deep copy?

voc/python/ast.py
```diff
@@ -1809,13 +1827,37 @@ def visit_Yield(self, node):
...
+            self.context.add_opcodes(
+                JavaOpcodes.INVOKEVIRTUAL(
+                    'org/python/types/Generator',
+                    'reset_message',
+                    args=[],
+                    returns='V'
+                )
+            )
+
+        # throw exception if there is one
```
freakboy3742: The comment says "*if* there is one" - but this code is executed regardless. It would be worth noting (both here, and in a docstring in the Java code) that `throw_exception` is a no-op if the generator doesn't have an exception state.

voc/python/ast.py
```diff
@@ -221,10 +224,17 @@ def convert_to_Name(_node):
                 lineno=_node.lineno,
                 col_offset=_node.col_offset
             )
+
+        # The node is a regular yield statement
+        # Mark it as 'regular_yield' to flush the generator's message upon restored
+        if isinstance(node, ast.Yield):
+            setattr(node, "regular_yield", True)
```
freakboy3742: What's the alternative to "regular" yield?

python/common/org/python/types/Generator.java
```diff
@@ -54,6 +63,120 @@ public void reset_message() {
         this.message = org.python.types.NoneType.NONE;
     }
 
+    @org.python.Method(
+            __doc__ = "Implement throw(type, value=None, traceback=None).",
+            args = {"type"},
+            default_args = {"value", "traceback"}
+    )
+    public org.python.Object _throw(org.python.Object type, org.python.Object value, org.python.Object traceback) {
```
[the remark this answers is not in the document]
BPYap: Resolved. Can't believe I missed that `name` attribute when browsing through voc documentation!
Now that there is already a mechanism to resolve name conflicts, shall I closed issue #827 ?

python/common/org/python/types/Generator.java
```diff
@@ -8,6 +11,7 @@
 
     private boolean just_started = true;
```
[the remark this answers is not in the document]
BPYap: Resolved. It seems `yield_point == 0` is sufficient to differentiate a just started generator and `yield_point` with any integer value indicates a already started generator.

python/common/org/python/types/Generator.java
```diff
@@ -54,6 +63,120 @@ public void reset_message() {
...
+    }
+
+    @org.python.Method(
+            __doc__ = "Return del(self)."
+    )
+    public void __del__() {
+        try {
+            this.close();
+        } catch (Exception e) {
+            //TODO: log error message in java equivalent of sys.stderr
```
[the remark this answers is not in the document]
BPYap: Resolved. Thanks! 😄

python/common/org/python/types/Generator.java
BPYap: Resolved. Added a class-level documentation block explaining possible states of a generator.

tests/structures/test_generator.py
BPYap: Resolved.

voc/python/ast.py
BPYap: Currently `parse_yield` from previous PR is modifying yield expression nodes in place, if I did not copy the list, subsequent visit to final body's childs (that contains yield expression nodes) will visit the transformed nodes and generate different opcodes.

I'm trying to refactor either `parse_yield` or come up with a way to copy generated opcodes from first visit since using `deepcopy` might not be efficient and optimal when the code is large.

voc/python/ast.py
BPYap: Resolved.

voc/python/ast.py
BPYap: Would `is_yield_stament` be better?

python/common/org/python/types/Generator.java
```diff
@@ -1,13 +1,23 @@
...
 
+import java.lang.reflect.Constructor;
+import java.lang.reflect.InvocationTargetException;
+
+/**
+ * Possible states of a generator:
+ * 1. yield_point == 0    : The generator is 'fresh' i.e. (just started)
+ * 2. expression  == null : The generator is exhausted
+ * 3. message     != none : The generator currently holds value sent from caller via send() method
+ * 4. exception   != null : The generator currently holds exception sent from caller via throw() method
```
freakboy3742: Are any of these states mutually exclusive? e.g., can I have both a message *and* an exception?

voc/python/ast.py
freakboy3742: I don't know if that's a likely outcome - the body of a yield has to be an expression, not a statement block, so there's a practical limit to how complex a yield body can be.

python/common/org/python/types/Generator.java
BPYap: Yes every states are mutually exclusive

voc/python/ast.py
freakboy3742: It's not (necessarily) a question of naming - I don't understand what a "regular yield" *is*. What's an "irregular" yield? How do you get an "irregular" yield?

BPYap: Initially I want to differentiate a yield statement (e.g. the plain old `yield 123`, which I called "regular yield") from the yield expression (`x = yield 123`).

However, I did some code refactoring just now and the attribute is no longer needed.

voc/python/ast.py
BPYap: @freakboy3742 I've refactored the code in latest commit so that I'm no longer modify/transform the nodes. However, deepcopy is still needed in the following lines:
 ```        
node_to_resolve = copy.deepcopy(node)  # deepcopy is required for multiple visits of finalbody from visitTry
self.context.yield_points.append(node_to_resolve)
self.context.next_resolve_list.append((node_to_resolve, OpcodePosition.YIELD))
```
in order to resolve the next opcode correctly for duplicated final body as the code snippet below in `blocks.py` changed node's attribute in place:
```
# Resolve any references to the "next" opcode.
for (obj, attr) in self.next_resolve_list:
    # print("        resolve %s reference on %s %s with %s %s" % (
    #     attr, obj, id(obj), opcode, id(opcode))
    # )
    setattr(obj, attr.value, opcode)
```

If everything is fine then it is ready for merge 😃

PR #833, opened 2018-06-08
[WIP] Modules dict optimization

patiences:
The `modules` dictionary is repeatedly loaded which is naive. Let's keep it in a local variable when possible.


patiences:
Closing this again. We can't/shouldn't do "lazy modules loading" (as in 6e20d27) because we can't rely on "first use" actually being the first use, leading to NPEs coming from expecting modules to be load when it actually wasn't.

PR #834, opened 2018-06-08
Implementing int(x, base)

henne90gen:
See #36 

Changes in this PR:
- implemented conversion of string to int with a specific base
- added test to show that it is working

### DIFF
--- a/python/common/org/python/types/Int.java
+++ b/python/common/org/python/types/Int.java
@@ -82,12 +82,18 @@ public class Int extends org.python.types.Object {
             try {
                 org.python.types.Str argStr = (org.python.types.Str) args[0];
                 org.python.types.Int argInt = (org.python.types.Int) args[1];
-                this.value = java.lang.Integer.parseInt(argStr.value, (int) argInt.value);
-            } catch (Exception e) {
+                int base = (int) argInt.value;
+                if (base == 0) {
+                    base = 10;
+                } else if (base < 2 || base > 36) {
+                    throw new org.python.exceptions.ValueError("int() base must be >= 2 and <= 36, or 0");
+                }
+                this.value = java.lang.Integer.parseInt(argStr.value, base);
+            } catch (java.lang.NumberFormatException e) {
                 throw new org.python.exceptions.ValueError("invalid literal for int() with base " + args[1] + ": '" + args[0] + "'");
             }
         } else if (args.length > 2) {
-            throw new org.python.exceptions.NotImplementedError("int() with a base is not implemented");
+            throw new org.python.exceptions.TypeError("int() takes at most 2 arguments (" + args.length + " given)");
         }
     }
 
--- a/python/common/org/python/types/Int.java
+++ b/python/common/org/python/types/Int.java
@@ -81,12 +81,19 @@ public class Int extends org.python.types.Object {
         } else if (args.length == 2) {
             try {
                 org.python.types.Str argStr = (org.python.types.Str) args[0];
+                if (args[1] instanceof org.python.types.NoneType) {
+                    throw new org.python.exceptions.TypeError("'NoneType' object cannot be interpreted as an integer");
+                }
                 org.python.types.Int argInt = (org.python.types.Int) args[1];
                 int base = (int) argInt.value;
                 if (base == 0) {
                     base = 10;
                 } else if (base < 2 || base > 36) {
-                    throw new org.python.exceptions.ValueError("int() base must be >= 2 and <= 36, or 0");
+                    if (org.Python.VERSION < 0x03060000) {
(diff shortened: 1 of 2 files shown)


python/common/org/python/types/Int.java
```diff
@@ -78,7 +78,15 @@ public Int(org.python.Object[] args, java.util.Map<java.lang.String, org.python.
                     );
                 }
             }
-        } else if (args.length > 3) {
+        } else if (args.length == 2) {
+            try {
+                org.python.types.Str argStr = (org.python.types.Str) args[0];
+                org.python.types.Int argInt = (org.python.types.Int) args[1];
+                this.value = java.lang.Integer.parseInt(argStr.value, (int) argInt.value);
+            } catch (Exception e) {
```
eliasdorneles: This should probably be catching only `NumberFormatException`, it's better to catch specific exceptions than use a catch-all with `Exception`.

python/common/org/python/types/Int.java
```diff
@@ -78,7 +78,15 @@ public Int(org.python.Object[] args, java.util.Map<java.lang.String, org.python.
...
+                throw new org.python.exceptions.ValueError("invalid literal for int() with base " + args[1] + ": '" + args[0] + "'");
+            }
+        } else if (args.length > 2) {
             throw new org.python.exceptions.NotImplementedError("int() with a base is not implemented");
```
eliasdorneles: This `NotImplementedError` message doesn't make sense, we should throw a different error message here, as CPython does:

```
>>> int('a', 16, 3)
TypeError: int() takes at most 2 arguments (3 given)
```

tests/datatypes/test_int.py
```diff
@@ -22,12 +22,18 @@ def test_getattr(self):
...
-            int('q', 16)
+            try:
+                print(int('q', 16))
+            except ValueError as err:
+                print(err)
             """)
 
+    def test_constructor_with_base(self):
+        self.assertCodeExecution("""
+            print(int('a', 16))
```
eliasdorneles: Can you please add some more cases here, and make it work for the edge cases?

```
print(int('22', 3))
print(int('22', 1))
print(int('5', 0))
print(int('5', -10))
print(int('5', 100))
```

henne90gen:
I've addressed almost all of your comments, except for the last one.
When I tried running int('1', 2, 3), an ArrayIndexOutOfBoundsException is being thrown before it even gets to the Int constructor.

python/common/org/python/types/Int.java
```diff
@@ -78,8 +78,22 @@ public Int(org.python.Object[] args, java.util.Map<java.lang.String, org.python.
...
-            throw new org.python.exceptions.NotImplementedError("int() with a base is not implemented");
+        } else if (args.length == 2) {
+            try {
+                org.python.types.Str argStr = (org.python.types.Str) args[0];
+                org.python.types.Int argInt = (org.python.types.Int) args[1];
+                int base = (int) argInt.value;
+                if (base == 0) {
+                    base = 10;
+                } else if (base < 2 || base > 36) {
+                    throw new org.python.exceptions.ValueError("int() base must be >= 2 and <= 36, or 0");
```
eliasdorneles: Almost there, now the tests are failing because the message isn't supposed to contain `, or 0` at the end.
The `0` here is considered as `falsy` value, e.g. `bool(0)` (you might add a test for the `None` case too).

henne90gen: Running it under Python 3.4.8 has the `, or 0` at the end. I'll add a switch for the version.

eliasdorneles: Are you sure?

```
$ $(pythonz locate 3.4.8)
Python 3.4.8 (default, Jun  9 2018, 11:17:16) 
[GCC 5.4.0 20160609] on linux
>>> int('22', 1)
ValueError: int() base must be >= 2 and <= 36
```

```
$ $(pythonz locate 3.5.0)
Python 3.5.0 (default, Nov  4 2015, 10:43:10) 
[GCC 4.8.4] on linux
>>> int('22', 1)
ValueError: int() base must be >= 2 and <= 36
>>> 
```

```
$ $(pythonz locate 3.6.0)
Python 3.6.0 (default, Feb  6 2017, 21:07:35) 
[GCC 4.8.4] on linux
>>> int('22', 1)
ValueError: int() base must be >= 2 and <= 36
```

henne90gen: Then it must be one of the minor versions under 3.6 that changed that behavior. I'll check which one it is.

henne90gen: The error message changed from 3.6.3 to 3.6.4

eliasdorneles:
Nice work, merci @henne90gen ! :rocket:

PR #836, opened 2018-06-11
Changed Bool construction to use getBool instead

BPYap:


patiences:
This should fix all the unit test failures in the builds, thanks!

PR #837, opened 2018-06-11
fix Method repr()

patiences:
While debugging I came across lines like: 
```
- Comparator: <bound method null.__eq__ of test>
``` 
Kept me on the wrong trail for a bit. Turns out `org.python.types.Type` objects on `org.python.types.Method`s don't have a `__name__` registered. This change uses `org.python.types.Type`'s `repr` method, resulting in logs like: 
```
- Comparator: <bound method <class 'str'>.__eq__ of test>
```

### DIFF
--- a/python/common/org/python/types/DictItems.java
+++ b/python/common/org/python/types/DictItems.java
@@ -111,7 +111,7 @@ public class DictItems extends org.python.types.Object {
             __doc__ = ""
     )
     public org.python.Object __bool__() {
-        return new org.python.types.Bool(!this.value.isEmpty());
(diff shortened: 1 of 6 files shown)


eliasdorneles:
Any idea why compilation failed in the CI?

patiences:
@eliasdorneles Ah, yes -- those failures have since been fixed in https://github.com/pybee/voc/pull/836. I'm not sure how to retrigger the build?

eliasdorneles:
@patiences gotcha! So, if you synchronize the branch of this PR with master (either with merge or rebase), it will incorporate those changes and trigger a new build.

PR #838, opened 2018-06-12
[WIP] keep module name string objects

patiences:


patiences:
Closing for now. This is probably not worth it, because we should try to store the modules themselves which would be more helpful than just the full_name.

PR #840, opened 2018-06-18
Implement iter() builtin

AmirYalamov:
Refs #76 

Going off the previous code for `iter()` in the case that there was no sentinel value, I wrote similar code, but with a while loop to detect when the iterator object reaches `sentinel`.


freakboy3742:
@AmirYalamov FYI - you can run the style checks locally with `ant checkstyle`

phildini:

PR #843, opened 2018-06-18
implemented builtin function tuple()

raraizan:
The current state trows an indexOutOfBoundsException, still working on some tests


raraizan:
Pending catch None and some other cases

PR #844, opened 2018-06-18
Fix exception handling

### LINKED ISSUE #255: Custom exception raising.
It is now possible to raise only exceptions from package org.python.exceptions. So for raise custom python exception there are 2 option:
1) Rebuild java-python-support.jar with your exception.
2) Write on java exception proxy-class.

vincent-prz:
This aims to fix https://github.com/pybee/voc/issues/255

Does this look good?


eliasdorneles:
So I've been pairing on this with @Heisenberg815 for some time, I think we're pretty close now.

@Heisenberg815: i was looking at the build failure, it looks like we missed the `except (exc1, exc2)` case =)
To be continued !

voc/python/ast.py
```diff
@@ -2300,7 +2300,120 @@ def visit_Index(self, node):
...
+            "UnicodeEncodeError",
+            "UnicodeError",
+            "UnicodeTranslateError",
+            "UnicodeWarning",
+            "UserWarning",
+            "ValueError",
+            "Warning",
+            "ZeroDivisionError",
+        ]
+        if node.type in builtin_exceptions:
```
BPYap: @Heisenberg815 @eliasdorneles, so I was compiling asyncio modules using the codes from this PR and I found a minor bug where voc transpilation fails when `node.type` is None. 

A potential fix would be adding `or node.type == None` condition on this line to redirect the program flow to `self._visitCoreException(node)`.

Cheers 😃

Heisenberg815:
@BPYap I have made the fix you suggested, plus one other small fix. It seems I still have one test to fix, will do it shortly.

Sorry for the late response.

Cheers!

tests/structures/test_try_catch.py
```diff
@@ -912,3 +912,78 @@ def test_try_multiple_except_else_finally(self):
...
+            class MyException(Exception):
+                pass
+            try:
+                raise MyException
+            except (MyException,):
+                print("Got a custom exception")
+            print('Done.')
+         """)
+
+        self.assertCodeExecution("""
```
eliasdorneles: Can you please put this test in a separate method, so that it can run as well?

This seems to be a regression (i.e., if i'm not mistaken, this second `assertCodeExecution` passes on master)

voc/python/ast.py
```diff
@@ -2474,9 +2473,128 @@ def visit_ExtSlice(self, node):
...
+            "ZeroDivisionError",
+        ]
+        return node.type is None or (
+                isinstance(node.type, ast.Name) and
+                (node.type.id in builtin_exceptions or
+                 node.type.id in self.symbol_namespace)
+        )
+
     @node_visitor
+    # [FIXME] - make it work for Tuples
```
eliasdorneles: leftover to remove?

eliasdorneles (changes requested):
This is almost good, some minor changes before it's good!

tests/structures/test_try_catch.py
vincent-prz: Done. Indeed, there seems to be a regression on `master`. But it seems it's the other way around: the second `assertCodeExecution` passes, while the first fails.

Do you think it's blocking the merge?

eliasdorneles: Thanks!
Yeah, that's what I said -- the second passes. :)

So, while the patch seems to indeed already make things better, I would like to have a look with more patience to see what's causing that regression.
I'm a bit swamped lately but I'll have some days off soon and will look into it!

phildini:

PR #846, opened 2018-06-21
Include SAMPLE_SUBS in custom substitutions 

patiences:


patiences:
This will fix the current CI build problems!

PR #847, opened 2018-06-25
Add custom exceptions test cases

BPYap:
Initially this PR includes attempts to support importing custom exceptions via ast parsing method. However, there are far more things and bugs to be resolved using this approach and the code get messier and messier. Then I saw PR #844 which is also solving the same problem (issue #255) but the approach is far more elegant and clean. Hence I revert all my previous commits (except the test cases).

**I'm using the same branch from my fork to show my progress and keep it as record for GSOC. :)

### DIFF
--- a/tests/structures/test_exception.py
+++ b/tests/structures/test_exception.py
@@ -105,6 +105,64 @@ class ExceptionTests(TranspileTestCase):
                     print(e, e.args)
         """)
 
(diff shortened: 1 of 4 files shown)


BPYap:
Closed as #844 is currently handling this in much more elegant and comprehensive fashion.

PR #848, opened 2018-06-25
Branch 1

ChetanKhanna:
Implemented the following methods:-
1. float() on complex numbers
2. bytes.splitlines()
3. byte.zfill()
4. a small test to str.splitlines() and byte.splitlines() that I found while goofing around with python terminal :)

Please let me know how I may improve upon them if needed. Thanks.


phildini:

PR #849, opened 2018-06-26
Fixed Unknown constant type <class 'frozenset'> in function definition

BPYap:
Previously when a set/frozenset appears as a constant in function definition, voc will fail during transpilation, for example, the code below will raise `AssertionError: Unknown constant type <class 'frozenset'>` during transpilation:
```
def func():
    1 in {1, 2, 3, 4, 5, 6}
```

This PR solve this issue by converting frozenset constant to a tuple when `blocks.add_tuple` is executing.

### DIFF
--- a/voc/python/blocks.py
+++ b/voc/python/blocks.py
@@ -181,6 +181,58 @@ class Block(Accumulator):
             java.Init('java/util/ArrayList'),
         )
 
+        def add_value(val):
+            if isinstance(val, bool):
+                if val is True:
+                    self.add_opcodes(
+                        JavaOpcodes.GETSTATIC('org/python/types/Bool', 'TRUE', 'Lorg/python/types/Bool;'),
+                    )
+                else:
+                    self.add_opcodes(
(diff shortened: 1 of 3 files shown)


voc/python/blocks.py
```diff
@@ -190,57 +242,22 @@ def add_tuple(self, data):
...
-
-                elif isinstance(value, str):
-                    self.add_opcodes(
-                        python.Str(value),
-                    )
-
-                elif isinstance(value, bytes):
+            elif isinstance(value, frozenset):
+                self.add_opcodes(
+                    python.Set(),
```
freakboy3742: Shouldn't this be FrozenSet?

voc/python/blocks.py
```diff
@@ -181,6 +181,58 @@ def add_tuple(self, data):
             java.Init('java/util/ArrayList'),
         )
 
+        def add_value(val):
```
freakboy3742: It's better to avoid inner methods if you can.

Issue #850, opened 2018-06-27
List and try-except combo raises AnalyzerException 


patiences:
This is a minimal example reproducing the ASM verifier error I got trying to run pystone. 

```
[x] = [0]

try:
    a = 1
except:
    a = 2
```

As far as I can tell all those bits have to be there: the list variable with the insides named, the try and except which is in this example obviously unrelated to `x` (strange combo I know). Full ASM error: 
https://gist.github.com/patiences/ab612f1af22cd426a58d6b67afce65db

PR #852, opened 2018-06-27
Do not leave List elements on the stack after visiting 

patiences:
When visiting a list node that's being stored into the list object should be popped from the stack when finished (this is already the case for `Tuple`s in ast.py a few lines below my change). This fixes https://github.com/pybee/voc/issues/850.

### DIFF
--- a/tests/structures/test_assignment.py
+++ b/tests/structures/test_assignment.py
@@ -1,3 +1,5 @@
+from unittest import expectedFailure
+
 from ..utils import TranspileTestCase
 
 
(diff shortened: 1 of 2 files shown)


tests/structures/test_assignment.py
```diff
@@ -69,7 +69,15 @@ def test_list_assignment(self):
...
             print(y)
             print(z)
-            """)
+        """)
+
+    def test_bad_list_assignment(self):
+        self.assertCodeExecution("""
+            try:
+                [x, y, z, a] = range(3)
+            except:
```
patiences: If you are wondering why the exception is ignored, it's because VOC throws a StopIteration while Cpython throws a ValueError (this behavior is captured in SequenceTests and they are expected failures).

patiences: Added the tests that catch the correct error in the next commit (as expected failures), so these can be removed when handling bad unpacking of sequences is implemented.

tests/structures/test_assignment.py
```diff
@@ -69,7 +71,37 @@ def test_list_assignment(self):
...
+        this test should be deleted when it is.
+        """
+        self.assertCodeExecution("""
+            try:
+                [x, y, z, a] = range(3)
+                print(x)
+                print(y)
+                print(z)
+                print(a)
+            except Exception as e:
```
freakboy3742: In retrospect, it's probably better to specifically catch `StopIteration` - that way the test will fail if/when we get around to fix the underlying problem, *this* test will fail, but hopefully the documentation will point at the way to make the test pass - to delete it :-)

patiences: Yes it would be -- but then the test will fail because there will be no `StopIteration` to catch on the Python side. Or are you suggesting to use `except (StopIteration, ValueError) as e:`? I think that when this problem is fixed, the correct test below will result in an unexpected pass, and the comment there should be enough to point the human to delete this test?

freakboy3742: Sorry - yes, I meant to catch `(StopIteration, ValueError) as e`

patiences: Done! :-)

PR #854, opened 2018-06-29
Nonlocal statement support

BPYap:
Nonlocal is used in some built-in module (e.g. tasks.py in asyncio module). Support for `nonlocal` statement should enable a number of files in `KNOWN_PROBLEM_MODULES` to compile.

Refer to ~~[GSOC week 8 log](https://github.com/pybee/voc/issues/763#issuecomment-403778552)~~ (currently *outdated*, will update soon) for implementation details.

**New features added in this PR:**
- `nonlocal` is now supported in function closure and generator closure. e.g:
```
def func():
    a = 'a from outer'
    def nested_func():
        nonlocal a
        print(a)  # will print 'a from outer'
        a = 'a from inner'
        print(a)  # will print 'a from inner'

    nested_func()
    print(a)  # will print 'a from inner'

func()
```

- Generator is now able to read variables from enclosing scopes. e.g:
```
def func():
    a = 'a from outer'
    def gen():
        print(a)  # will print 'a from outer'
        yield

    next(gen())

func()
```


BPYap:
Passing `#locals` as storage space for closure variables seem to work well, to my surprise it even fixes `test_recursive_in_function` case 😄

PR #864, opened 2018-07-03
Callables should handle null args and kwargs

patiences:
When functions are called, two data structures are created to hold their arguments: a java array for the args and a java hashmap for the kwargs. These are created even when the callable takes no arguments. We can prevent creating these empty data structures by making sure there are sufficient null checks when the callable is invoked. 

(This is a copy of [this PR](https://github.com/pybee/voc/pull/845))


patiences:
Current status: 

I'm unable to show that this change results in a performance boost, and I think this is because while not creating some arrays and hashmaps that will just be empty is an improvement, the extensive null checking on invocation cancels out the savings. If that is the case, I'm not sure this PR is a good idea because it could mean that performance worsens a bit for callables that *do* take arguments but we have to do the null checking anyway. Thoughts @freakboy3742? :-)

python/common/org/python/java/Function.java
```diff
@@ -269,8 +269,8 @@ public static MatchType parameterMatch(java.lang.Class<?> from_type, java.lang.C
 
         java.lang.reflect.Method method = null;
         java.lang.StringBuilder signature = new java.lang.StringBuilder();
-        java.lang.Class<?>[] arg_types = new java.lang.Class<?>[args.length];
-        int n_args = args.length;
+        int n_args = (args == null) ? 0 : args.length;
+        java.lang.Class<?>[] arg_types = new java.lang.Class<?>[n_args];
```
freakboy3742: This may be the place where the performance hit is actually taking place (or, at least, somewhere else you can potentially optimise); every 0-arg call is creating an array of size 0.

python/common/org/python/java/Function.java
```diff
@@ -381,9 +381,10 @@ public static MatchType parameterMatch(java.lang.Class<?> from_type, java.lang.C
     }
 
     public java.lang.Object[] adjustArguments(java.lang.reflect.Method method, org.python.Object[] args, java.util.Map<java.lang.String, org.python.Object> kwargs) {
-        java.lang.Object[] adjusted = new java.lang.Object[args.length];
+        int n_args = (args == null) ? 0 : args.length;
+        java.lang.Object[] adjusted = new java.lang.Object[n_args];
```
freakboy3742: Similarly - this is an empty array being created.

freakboy3742: In fact, I don't think there's no need to call this method *at all* if there aren't any arguments.

patiences: You're absolutely right! Done

python/common/org/python/java/Function.java
patiences: Done!

patiences:
It appears that the reason why I can't write a benchmarking test to show the performance improvement is because the preparation for making a function invocation (the site of this optimization) is miniscule compared to the cost of making a function call. I reached this conclusion because a) the bytecode of the benchmark test I wrote *is* smaller (see gist [here](https://gist.github.com/patiences/31365d01f3af577f77001acd3752b18b)) and b) there is no noticeable performance difference between invoking functions that don't take arguments and functions that do (see gist [here](https://gist.github.com/patiences/a453e5b73882d4d758cd6680510ef8b6)).

PR #874, opened 2018-07-25
Implement collections.defaultdict

BPYap:


BPYap:
@freakboy3742 Sorry my brain got short-circuited just now 😅. Actually the reason I make a private `__getitem__` on org.python.types.Dict is to preserve dict behavior on defaultdict. You see, other than `__getitem__`, defaultdict has exactly the same behavior as dict when other methods are call. 

e.g. `defaultdict.get(x)` should return `None` (same as dict) when `x` is not a key instead of invoking the `default_factory` for missing value. 
However the implementation of `get(x)` in voc calls `__getitem__(x)` which indirectly calls `__missing__(x)`, causing `defaultdict` to return `default_factory` value instead of `None`, hence the need of another `__getitem__` without calls to `__missing__`.

PR #877, opened 2018-07-27
Contains/Not Contains throws wrong errors 

patiences:
1. `in`/`__contains__` throws the wrong error on non-iterable types (like int) -- should be `TypeError` not `AttributeError` 
2. `not in`/`__not_contains__` should probably use `__contains__` instead of looking for `__not_contains__` attribute 
2. org.python.types.Str.__contains__ should return a `org.python.types.Bool` 

This PR makes the following changes: 
1. Addresses the problems above 
2. Removes the `__not_contains__` attribute from org/python/Objects. I verified that each `__not_contains__` attribute that was defined was just an inverse of its `__contains__` attribute. Since *the operator not in is defined to have the inverse true value of in* (from https://docs.python.org/3/reference/expressions.html#membership-test-operations), there is no need to define separate `__not_contains__` attributes, they should all direct to `__contains__`. See org/python/types/Object.java#__not_contains__.

### DIFF
--- a/python/common/org/python/types/Object.java
+++ b/python/common/org/python/types/Object.java
@@ -1371,24 +1371,14 @@ public class Object extends java.lang.RuntimeException implements org.python.Obj
             try {
                 result = v.__contains__(w);
             } catch (org.python.exceptions.AttributeError e) {
-                if (org.Python.VERSION < 0x03060000) {
-                    throw new org.python.exceptions.TypeError(String.format(
-                        "unorderable types: %s() %s %s()", v.typeName(), "in", w.typeName()));
-                } else {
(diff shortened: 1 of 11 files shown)


python/common/org/python/types/Object.java
```diff
@@ -552,14 +552,6 @@ public void __delitem__(org.python.Object index) {
         throw new org.python.exceptions.AttributeError(this, "__contains__");
     }
```
patiences: I'm not sure that this AttributeError is strictly necessary -- seems like a `__contains__` call is either successful (it's defined elsewhere) or a `TypeError` is thrown (and this is constructed after this `AttributeError` is thrown and caught). Could this be replaced by `org.python.types.NotImplemented`?

PR #880, opened 2018-07-30
[WIP] do not create `__dict__`s for builtin types

patiences:


patiences:
This PR has been stalling for the past week because I've been trying to understand why the performance gain (~1% improvement is what I've been seeing on my benchmarking test) is so modest, when this change was supposed to have a bigger effect. It turns out that creating java.util.HashMaps is not expensive at all as the table is not created until something is put in the table (a source [here](https://stackoverflow.com/questions/31419029/why-is-creating-a-hashmap-faster-than-creating-an-object)). In my opinion it's not worth making this change for such a small benefit plus then there is the distinction between `org/python/Object`s that have `__dict__`s and ones that don't.

Issue #895, opened 2018-08-07
Inner generator function throws NameError when accessing outer function's arg 


patiences:
```
def fun(lst):
    def gen():
        for x in lst:
            yield x

    return list(gen())

print(fun([1, 2, 3, 4]))
```

This results in: 

```
test_generator_in_function (tests.structures.test_generator.GeneratorTests) ... java.lang.reflect.InvocationTargetException
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:498)
	at python.testdaemon.TestDaemon.main(TestDaemon.java:65)
Caused by: NameError: name 'lst' is not defined
	at org.python.types.Module.__getattribute__(Module.java:38)
	at python.test.fun$gen.invoke$generator(test.py:3)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:498)
	at org.python.types.Generator.__next__(Generator.java:78)
	at org.python.types.List.<init>(List.java:71)
	at sun.reflect.NativeConstructorAccessorImpl.newInstance0(Native Method)
	at sun.reflect.NativeConstructorAccessorImpl.newInstance(NativeConstructorAccessorImpl.java:62)
	at sun.reflect.DelegatingConstructorAccessorImpl.newInstance(DelegatingConstructorAccessorImpl.java:45)
	at java.lang.reflect.Constructor.newInstance(Constructor.java:423)
	at org.python.types.Type.invoke(Type.java:478)
	at python.test.fun(test.py:6)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:498)
[...]

PR #898, opened 2018-08-08
Add test for inner generator function bug

patiences:
Test for https://github.com/pybee/voc/issues/895


patiences:
Thanks @BPYap!! Looks good.

PR #911, opened 2018-08-15
Do not re-create the same immutable object

patiences:
This is a cleanup of methods in various immutable object classes that call `new` on themselves. That seems unnecessary since they are immutable. 

Changes: 

1. Removed `org.python.Object.byValue()`, which was being used in yield statements (removed those too). 
2. Removed `setValue()`, which was being used internally in `Super.java` and `Str.java` (inconsistently). Removed those usages too. 
3. Replaced statements in immutable datatypes (`Str`, `Int`, etc) like `return new org.python.types.X(this.value)` with `return this` to prevent creating the same object multiple times.

### DIFF
--- a/python/common/org/python/types/Bool.java
+++ b/python/common/org/python/types/Bool.java
@@ -34,7 +34,7 @@ public class Bool extends org.python.types.Object {
     }
 
     public org.python.Object byValue() {
-        return org.python.types.Bool.getBool(this.value);
+        return this;
     }
 
     public int hashCode() {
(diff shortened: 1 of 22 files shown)


python/common/org/python/types/Bool.java
```diff
@@ -34,7 +34,7 @@ void setValue(org.python.Object obj) {
     }
 
     public org.python.Object byValue() {
-        return org.python.types.Bool.getBool(this.value);
+        return this;
     }
```
patiences: I saw the comment on org.python.Object: "Return a version of the object that can be used when returning by value. For most objects, this will be itself; but primitive types need to return a copy of themselves to ensure that they aren't modified." Which makes sense, but at the same time, there are only 2 `Bool` objects ever, so it seems unneeded.

Issue #912, opened 2018-08-17
tuple() function does not handle range 


patiences:
```
tuple(range(3))
```
fails with 
```
java.lang.reflect.InvocationTargetException
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:498)
	at python.testdaemon.TestDaemon.main(TestDaemon.java:65)
Caused by: java.lang.ArrayIndexOutOfBoundsException: 0
	at org.python.types.Type.invoke(Type.java:464)
	at python.test.module$import(test.py:1)
	at python.test.main(test.py)
	... 5 more
```

Issue #907, opened 2018-08-14
tuple() method does not handle improper arguments 


patiences:
```
tuple(0)
``` 
 and 
```
tuple(0, 1) 
``` 
causes VOC to throw: 
```
java.lang.reflect.InvocationTargetException
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:498)
	at python.testdaemon.TestDaemon.main(TestDaemon.java:65)
Caused by: java.lang.ArrayIndexOutOfBoundsException: 0
	at org.python.types.Type.invoke(Type.java:464)
	at python.test.module$import(test.py:1)
	at python.test.main(test.py)
	... 5 more
```
The former should be a `TypeError: 'int' object is not iterable`. The latter should be a `TypeError: tuple() takes at most 1 argument (2 given)`.

PR #913, opened 2018-08-17
Throw TypeError when constructor receives too many args

patiences:
This is a fix for https://github.com/pybee/voc/issues/907 and https://github.com/pybee/voc/issues/912. If too many arguments are provided to a builtin constructor, the right `TypeError` should be thrown. However, VOC was throwing `ArrayIndexOutOfBoundsException`s because this case wasn't being handled in `Type.java` during argument population.

### DIFF
--- a/python/common/org/python/types/Type.java
+++ b/python/common/org/python/types/Type.java
@@ -466,14 +466,8 @@ public class Type extends org.python.types.Object implements org.python.Callable
                             kwargs.put(arg_name, args[a]);
                         }
                     } else {
-                        java.lang.String name = org.Python.typeName(this.klass);
-                        if (n_args == 0) {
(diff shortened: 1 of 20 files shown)


python/common/org/python/types/Type.java
```diff
@@ -465,6 +465,9 @@ public boolean __setattr_null(java.lang.String name, org.python.Object value) {
                             }
                             kwargs.put(arg_name, args[a]);
                         }
+                    } else {
+                        // More arguments than expected, call the constructor anyway to handle too many arguments
+                        adjusted_args = args;
```
patiences: I thought about throwing a TypeError here, but the specific constructor has to do it because the error messages are not consistent across datatypes.

Issue #923, opened 2018-09-19
Missing bracket in voc/tests/microbenchmarks.py line 206


nikhilmudholkar:
I am starting to contribute to this repository. While running the test cases i observed that my test cases were not running. Further analysis showed that there was a bracket missing right after """ in the code. In all other similiar methods, the bracket is present. Adding that bracket made all the tests run correctly. Is this an issue? What can be done to rectify it?

PR #924, opened 2018-09-22
Add missing parenthesis in microbenchmarks

auscompgeek:
Fixes: #923

### DIFF
--- a/tests/microbenchmarks.py
+++ b/tests/microbenchmarks.py
@@ -1,8 +1,10 @@
 import sys
 from os import path
+
 sys.path.append(path.dirname(path.dirname(path.abspath(__file__))))
 
-from utils import *
+from utils import TranspileTestCase, adjust  # noqa: E402
+
 
 def test_small_integers(test_case):
     print("Running", "test_small_integers")
@@ -13,6 +15,7 @@ def test_small_integers(test_case):
                 for k in range(-5, 257):
                     x = x + 1
         """), timed=True)
+
 
 def test_booleans(test_case):
     print("Running", "test_booleans")
@@ -49,6 +52,7 @@ def test_booleans(test_case):
             y = (False != False) != False
         """), timed=True)
 
+
 def test_global_var_load(test_case):
     print("Running", "test_global_var_load")
     test_case.runAsJava(adjust("""
@@ -67,6 +71,7 @@ def test_global_var_load(test_case):
             print(x)
             print(y)
     """), timed=True)
+
 
 def test_class_var_load(test_case):
     print("Running", "test_class_var_load")
@@ -92,6 +97,7 @@ def test_class_var_load(test_case):
 
     """), timed=True)
 
+
 def test_function_var_load(test_case):
     print("Running", "test_function_var_load")
     test_case.runAsJava(adjust("""
@@ -112,6 +118,7 @@ def test_function_var_load(test_case):
         for i in range(100000):
             foo()
     """), timed=True)
+
 
 def test_code(test_case):
     print("Running", "test_code")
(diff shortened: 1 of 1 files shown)


tests/microbenchmarks.py
```diff
@@ -203,7 +203,7 @@ def test_dict_set(test_case):
             dict["meow"] = 4
             dict["cockadoodledoo"] = 5
             dict["hiss"] = 6
-    """, timed=True)
+    """), timed=True)
 
 def test_class_init(test_case):
```
brutusthebee: At column 1: (E302) expected 2 blank lines, found 1

tests/microbenchmarks.py
```diff
@@ -203,7 +203,7 @@ def test_dict_set(test_case):
...
     print("Running" , "test_class_init")
```
brutusthebee: At column 20: (E203) whitespace before ','

auscompgeek:
Hmm. From the looks of things, this entire file doesn't follow PEP 8 whitespace. Should I reformat the entire file, or just leave it as is?

freakboy3742:
@auscompgeek Oh, *please* reformat it :-) 

It appears that the pre-commit linting check wasn't working for a while, so a lot of problems have drifted in over time. Unfortunately, this now means that a lot of people are getting tagged with problems that aren't actually theirs. If you're able to fix them, that would be most appreciated.

auscompgeek:
Ok. I'd probably want to run an autoformatter in that case. I am seeing some sort of vague style preference in this file that black and yapf trample over though — both black and yapf want to put the `adjust` calls on separate lines.

black:

```diff
--- tests/microbenchmarks.py    2018-09-22 03:03:08.239819 +0000
+++ tests/microbenchmarks.py    2018-09-22 03:03:19.308244 +0000
@@ -1,24 +1,34 @@
 import sys
 from os import path
+
 sys.path.append(path.dirname(path.dirname(path.abspath(__file__))))
 
 from utils import *
+
 
 def test_small_integers(test_case):
     print("Running", "test_small_integers")
-    test_case.runAsJava(adjust("""
+    test_case.runAsJava(
+        adjust(
+            """
         for i in range(100):
             for j in range(100):
                 x = -5
                 for k in range(-5, 257):
                     x = x + 1
-        """), timed=True)
+        """
+        ),
+        timed=True,
+    )
+
 
 def test_booleans(test_case):
     print("Running", "test_booleans")
-    test_case.runAsJava(adjust("""
+    test_case.runAsJava(
+        adjust(
+            """
         for i in range(100000):
             x = True
             y = False
 
             x = (not x and y)
```

yapf (default config):
```diff
--- tests/microbenchmarks.py    (original)
+++ tests/microbenchmarks.py    (reformatted)
@@ -4,19 +4,24 @@
 
 from utils import *
 
+
 def test_small_integers(test_case):
     print("Running", "test_small_integers")
-    test_case.runAsJava(adjust("""
+    test_case.runAsJava(
+        adjust("""
         for i in range(100):
             for j in range(100):
                 x = -5
                 for k in range(-5, 257):
                     x = x + 1
-        """), timed=True)
+        """),
+        timed=True)
+
 
 def test_booleans(test_case):
     print("Running", "test_booleans")
-    test_case.runAsJava(adjust("""
+    test_case.runAsJava(
+        adjust("""
         for i in range(100000):
             x = True
             y = False
```

Now, usually I'm not a fan of autopep8, but would you rather just make pycodestyle happy, or should we go for what yapf is suggesting?

freakboy3742:
@auscompgeek The immediate requirement is just to make pycodestyle happy; I've been contemplating a project-wide move to black because it seems to be the way things are moving. I can't claim I'm in love with 100% of black's settings, but I can't say I've seen anything I *actively* hate either, and having an opinionated and automatable set of defaults is vastly preferable to vague styles.

PR #929, opened 2018-10-27
Update os.java

sinbycosmay:
Perfectly indented some of the errors that might be popping up from time to time.


phildini:

PR #930, opened 2018-11-05
updated exception messages in str.java

mvanderbeek123:


tests/datatypes/test_str.py
```diff
@@ -914,6 +914,17 @@ def test_too_many_arguments(self):
                 print(err)
 
         """)
+
```
brutusthebee: At column 1: (W293) blank line contains whitespace

PR #932, opened 2018-12-10
update exception messages

mvanderbeek123:


mvanderbeek123:
for version 3.7.1 of python

freakboy3742 (changes requested):
Thanks for the PR! This change is definitely needed; but it needs to be conditional based on Python version. We run our test suite on Python 3.4, 3.5 and 3.6, so we need to preserve the "old" error messages for those versions. There are some existing examples of this kind of conditional error message elsewhere in the code - for example, [this error](https://github.com/pybee/voc/blob/master/python/common/org/Python.java#L1417) in the pow() function.

We also need to add 3.7 testing to the Beekeeper configuration. A pybee/py37 image is available, so it should be a relatively straightforward configuration change. That image is currently using 3.7.0; I don't know if the error message changes you've made here require 3.7.1, but if they do, a PR to the py37 branch of the pybee/comb GitHub repository should be all that is required.

phildini:

PR #937, opened 2019-01-03
Hashing Functions

AnimeshSinha1309:
## Functions Implemented
* **`Float.__hash__()`** which implements `__hash__()` for a floating-point type.
* **`Boolean.__hash__()`** which implements `__hash__()` for a boolean type.
* **`Int.__hash__()`** which implements `__hash__()` for a integral numbers.
* **`Complex.__hash__()`** which implements `__hash__()` for a complex numbers.
* **`Float.as_integer_ratio()`** which converts a Python Float to a tuple containing the numerator and denominator.
## Tests fixed
Fixed the tests:
* `tests.builtins.test_hash.BuiltinHashFunctionTests.test_bool`
* `tests.builtins.test_hash.BuiltinHashFunctionTests.test_float`
* `tests.builtins.test_hash.BuiltinHashFunctionTests.test_complex`
* `tests.builtins.test_hash.BuiltinHashFunctionTests.test_int`
## Other Notes (Plans)
* The **`hashCode()`** function needs to be fixed for types other than Integer, Float, and Boolean, so that they are consistent with Python. 
* We could get rid `@expectedFailure` in `SetTests.test_equal_value_different_datatypes`, atleast for **Python 3.6**. The test gets fixed as the hash functions are now consistent across the types tested there. However, the tests are still failing on **Python 3.4**, so I am not implementing `hashCode()` using `__hash__()` here.


AnimeshSinha1309:
Please tell me if I should start converting all the `hashCode()` functions to **return long**. I will start on it, but I wanted to get permission first.

freakboy3742:
@AnimeshSinha1309 Well.. no - because the [return type of `hashCode` is specified by Java](https://docs.oracle.com/javase/7/docs/api/java/lang/Object.html#hashCode()).

tests/datatypes/test_set.py
```diff
@@ -4,7 +4,8 @@
 
 
 class SetTests(TranspileTestCase):
-    def test_complex_element(self):
+
+  def test_complex_element(self):
```
brutusthebee: At column 3: (E111) indentation is not a multiple of four

AnimeshSinha1309: Yeah, I just realized that, had overlooked it. Thanks!

python/common/org/python/types/Int.java
```diff
@@ -45,7 +49,7 @@ public int hashCode() {
         }
     }
 
-    private Int(long value) {
+    public Int(long value) {
```
freakboy3742: This constructor is private for an important reason - you shouldn't be instantiating integers directly, because it undermines instance caching.

python/common/org/python/types/Bool.java
```diff
@@ -27,6 +27,11 @@ public int hashCode() {
         return new java.lang.Boolean(this.value).hashCode();
     }
 
+    public org.python.types.Int __hash__() {
+        if (this.value == true) return org.python.types.Int.getInt(1);
+        else return org.python.types.Int.getInt(0);
```
freakboy3742: I'm surprised this isn't picked up by the code linter... but this formatting needs to be cleaned up.

Firstly, the body of if/else clauses should always be indented. Secondly "naked" if clauses shouldn't be used.

This code should look like:
```
if (this.value == true) {
    return org.python.types.Int.getInt(1);
} else {
    return org.python.types.Int.getInt(0);
}
```

python/common/org/python/types/Float.java
```diff
@@ -14,6 +14,46 @@ public int hashCode() {
         return new java.lang.Double(this.value).hashCode();
     }
 
+    public org.python.types.Int __hash__() {
+        final long _PyHashINF = 314159L, _PyHashNAN = 0L, _PyHashMULTIPLIER = 1000003L;
```
freakboy3742: This hashing algorithm is complex, so I'm guessing it wasn't something you invented yourself - in which case, you should reference where the original came from. This is partially to give appropriate credit to whoever wrote the original, but also to assist whoever comes here later to work out what is going on.

python/common/org/python/types/Float.java
```diff
@@ -737,6 +777,84 @@ public void __delitem__(org.python.Object other) {
         throw new org.python.exceptions.TypeError("'float' object does not support item deletion");
     }
 
+    @org.python.Method(
+        __doc__ = ""
+    )
+    public org.python.types.Tuple as_integer_ratio() {
```
freakboy3742: This method isn't referenced anywhere else in this PR - so what is it for?

It's also not a method that is on Python integers (as far as I can tell), so assuming it *is* required, there's no need for it to be annotated for exposure into the Python namespace.

python/common/org/python/types/Float.java
```diff
@@ -737,6 +777,84 @@ public void __delitem__(org.python.Object other) {
...
+
+    /**
+     * Private Helper Function
+     * Breaks down a floating-point value into it's component mantissa and exponent.
+     *
+     * @param x - double, the value of the number that needs to be broken down.
+     * @return MantissaExponent - pair containing the mantissa and exponent.
+     *
+     * @notes
+     *  Code Stolen from Jython
```
freakboy3742: "stolen" isn't a nice word. Jython's license allows re-use, so what we're doing here isn't criminal. "Based on a version from Jython" is a more appropriate phrasing.

freakboy3742 (changes requested):
A good start - but there's some code formatting changes that need to be cleaned up, and some documentation issues.

python/common/org/python/types/Float.java
AnimeshSinha1309: `float.as_integer_ratio()` is a part of the **Python standard library for Floats, but not for Integers**. So if this function is not exposed to the Python namespace, the following code would execute on Python and fail on Voc.
```python
print((3.5).as_integer_ratio()) # prints (7, 2)
print((3).as_integer_ratio()) # throws an error
```
So I do believe that this function should be a part of the PR, I have checked that the diff is empty in my implementation between CPython and Voc runs. Tell me if I am wrong / should add tests for this.

python/common/org/python/types/Float.java
AnimeshSinha1309: Sure, :stuck_out_tongue:!

python/common/org/python/types/Int.java
AnimeshSinha1309: I switched everything over to `Int.getInt()`, I hope.

python/common/org/python/types/Float.java
freakboy3742: Huh - TIL. I didn't know about this one, and for some reason, it didn't come up when I did an autocomplete check in iPython (i'm guessing I mistyped the name of the method). 

I did some digging, and it appears [`as_integer_ratio()` *will* be on integers in Python 3.8](https://docs.python.org/3.8/library/stdtypes.html?highlight=as_integer_ratio#int.as_integer_ratio) - but we don't need to handle that yet.

So yes - it should definitely exist, and yes, there should be tests.

freakboy3742 (changes requested):
The implementation is looking strong now - all that is needed is tests. You've added the 'as_integer_ratio' method - and there are some non-trivial edge-cases in the `__hash__` implementations for Float and Complex; they all need to be tested.

AnimeshSinha1309:
Okay, I am sorry I was not here. But I am not really sure how to test edge cases for the Hashing function.
I will try getting familiar with this, but please advise me on what / how to test.

tests/datatypes/test_complex.py
```diff
@@ -39,6 +39,16 @@ def test_equality_with_numbers_when_zero_imag(self):
...
+            print(hash(complex(2, 5)))
+            print(hash(complex(2, 0)))
+            print(hash(complex(20893129038, 1903813213)))
+            print(hash(complex(12908.19238, 1293.19238)))
+            print(hash(complex(0, 0)))
+        """)
+
+
 
 class UnaryComplexOperationTests(UnaryOperationTestCase, TranspileTestCase):
```
brutusthebee: At column 1: (E303) too many blank lines (3)

AnimeshSinha1309: Thanks so much Brutus-the-Bee, I fixed it.

tests/builtins/test_bytearray.py
```diff
@@ -6,4 +6,4 @@ class BytearrayTests(TranspileTestCase):
 
 
 class BuiltinBytearrayFunctionTests(BuiltinFunctionTestCase, TranspileTestCase):
-    functions = ["bytearray"]
```
freakboy3742: This doesn't make the test pass - it removes the test, and in a way that can't be observed by the test suite.

The preferred approach is to use a `not_implemented` list - see `test_eval` or `test_help` for an example of usage.

freakboy3742 (changes requested):

phildini:

PR #855, opened 2018-06-29
Implement built-in ascii()

AmirYalamov:
Refs #76

PR #942, opened 2019-02-01
ascii bug (not fixed) #855

0xrushi:
Refs #855 

is the __repr__() definition correct ? it seems that is printing unicode characters too.Anyways i tried this segment can anyone explain where is it wrong.
```AssertionError: ">>> f = ascii\n>>> x = 0\n>>> f(x)\n'0'\[194 chars]==\n" != '>>> f = ascii\n>>> x = 0\n>>> f(x)\n0\n\[184 chars]==\n'
  >>> f = ascii
  >>> x = 0
  >>> f(x)
- '0'
+ 0
```
 
I am getting errors like this, => actual python implementation wants Integer.
but in python shell if i type type(ascii(3.4)) it gives String

### DIFF
--- a/python/common/org/Python.java
+++ b/python/common/org/Python.java
@@ -484,7 +484,54 @@ public class Python {
             args = {"object"}
     )
     public static org.python.types.Str ascii(org.python.Object object) {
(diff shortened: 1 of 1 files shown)


phildini:

PR #952, opened 2019-05-16
Format and f-string support (and minor other changes)

jorants:
Major additions
- Implements `format()` as per python standard
- Implements standard `__format__` for all objects
- Implements `__format__` for `Str`, `Float`, `Int` and `Complex`
- Fixes `__repr__` for `Complex` by using the already present code from the `__str__` method instead.
- `__repr__` uses `__format__` for float type
- Implements f-strings by manipulating the AST.

Small changes:
- Added `generic_visit` to `ast.py`. Unimplemented node types previously where ignored but now give an error.
- Added a fail in `build-android.xml` when `android.jar` is not found

### DIFF
--- a/voc/python/ast.py
+++ b/voc/python/ast.py
@@ -99,7 +99,7 @@ class NameVisitor(ast.NodeVisitor):
 class LocalsVisitor(ast.NodeVisitor):
     def __init__(self, context):
         self.context = context
-        
+
     def visit_Name(self, node):
         if type(node.ctx) == ast.Store:
             if node.id not in self.context.local_vars:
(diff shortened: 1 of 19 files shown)


voc/python/ast.py
```diff
@@ -1,4 +1,5 @@
 import ast
+import _ast
 import copy
 import sys
 import traceback
```
brutusthebee: At column 1: (F401) 'traceback' imported but unused

jorants: This indeed seems like it can be removed, but has noting to do with this pull request.

jorants: I removed it in the latest commit.

PR #955, opened 2019-06-30
Be explicit about python3

dboddie:
Some Linux distributions will use `python3` not `python` for Python 3 scripts. We should use that instead of assuming that `python` is now Python 3.


dboddie:
Fair enough. I was just fixing a problem I encountered working outside a virtual environment. In the virtual environment, `python3` is also a valid way to invoke the interpreter.

PR #956, opened 2019-06-30
Exclude a test module from being built

dboddie:
The `__phello__.foo` module from Ouroboros breaks jar->dex conversion.


dboddie:
The failure is the result of the bytearray tests being fed data from container types: set, list, dictionary, etc.

I can't see exactly how this happens right now. Presumably this should have failed when the change that combined these types was introduced, but I'm not seeing where and when that occurred.

PR #965, opened 2019-07-24
Imported a common PythonActivity class

dboddie:
The class will look for the application's main function using the application's own metadata.

This potentially allows each application to be written without the need to generate and build application-specific Java files.


phildini:

PR #966, opened 2019-07-24
Mark no-parameter methods static

whydoubt:
Python methods with no parameters are implicitly static. However, in the Java class file, if the method meta-data does not have it explicitly marked as such, DEXing results in the following translation error: "com.android.dx.util.MutabilityException: immutable instance".  This PR explicitly marks no-parameter methods as static.

### DIFF
--- a/python/common/org/python/types/Function.java
+++ b/python/common/org/python/types/Function.java
@@ -203,8 +203,8 @@ public class Function extends org.python.types.Object implements org.python.Call
                     }
                 }
 
-                throw new org.python.exceptions.TypeError(this.name + "() missing " + missingArgs.size() + " required positional "
-                        + (missingArgs.size() == 1 ? "argument" : "arguments") + ": " + sb.toString());
+                throw new org.python.exceptions.TypeError(this.name + "() missing " + missingArgs.size()
+                        + " required positional argument" + (missingArgs.size() == 1 ? ": " : "s: ") + sb.toString());
             }
         }
     }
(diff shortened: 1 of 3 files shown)


freakboy3742 (changes requested):
While a 0-parameter method will definitely behave like a static method in many circumstances, I feel like this is a hack around a very specific case, rather than a general solution. There's a bigger issue around `@staticmethod` handling; and while you may not necessarily want to fully implement `@staticmethod` right now, at least setting up the infrastructure for future staticmethod handling, rather than hard-coding the 'no parameters' == static hack in multiple locations in the code, is something we should aim for. 

A test case for this is also required.

whydoubt:
> While a 0-parameter method will definitely behave like a static method in many circumstances, I feel like this is a hack around a very specific case, rather than a general solution.

Yes it was a hack. I am now researching a more-general solution, which will of course have deeper implications.

> There's a bigger issue around `@staticmethod` handling; and while you may not necessarily want to fully implement `@staticmethod` right now, at least setting up the infrastructure for future staticmethod handling,

After looking into how `@staticmethod` / `staticmethod()` fit into the equation, I would say that the particular case I was trying to deal with was really a distinct issue.  However, the more-general solution I am looking into may, as you indicate, have some bearing on dealing with the `staticmethod` issue in the future.

>rather than hard-coding the 'no parameters' == static hack in multiple locations in the code

While perhaps a moot point, this was the _only_ place the hack was needed to fix the particular class of DEX faults.
 
> A test case for this is also required.

Agreed

whydoubt:
> > While a 0-parameter method will definitely behave like a static method in many circumstances, I feel like this is a hack around a very specific case, rather than a general solution.
>
> Yes it was a hack. I am now researching a more-general solution, which will of course have deeper implications.

After further investigation, I keep coming back to my original solution.  For a Python method that takes zero parameters, there is no instance.  A Java method that takes no instance is static.  Simple as that.

> > There's a bigger issue around `@staticmethod` handling; and while you may not necessarily want to fully implement `@staticmethod` right now, at least setting up the infrastructure for future staticmethod handling,
 
I still regard how _Python_ creates static methods (i.e. `@staticmethod`) as only tangentially related to this case.

> > rather than hard-coding the 'no parameters' == static hack in multiple locations in the code

This only affects the transpile_wrapper method of the Method class.

> > A test case for this is also required.

Done, and it will included in the next push.

phildini:

PR #967, opened 2019-07-24
Override local vars when creating base method for generator

whydoubt:
Fixes: https://github.com/beeware/voc/issues/950


whydoubt:
I added a testcase, and reworked the code to not trample self.local_vars.  Hopefully it also makes the intent a bit clearer.  Note that this is the only place it is needed, because this is the only nested wrapper (one transpile_wrapper calling another transpile_wrapper)

PR #970, opened 2019-08-17
Python 3.7

whydoubt:
Updates exception messages and tests to make test suite complete successfully with Python 3.7 (without affecting results for Python 3.4-3.6)


freakboy3742:
Apologies for the noise on this PR; I thought I'd make the "simple" fix of updating the Python 3.7 test environment from 3.7.2 to 3.7.4... and the more I pulled on that thread, the worse things got...

However, it looks like the errors picked up by the first pass of CI were introduced in 3.7.3 or 3.7.4, so they're all passing now! Thanks for the patch!

PR #987, opened 2019-10-19
added clarification to tutorial-1.rst

BellaTrix25336:
added clarification to avoid others running into the same silly mistake I made... after following so many instructions for installations and tutorials it's easy to get into a copy-paste without thinking mode. So maybe being overly explicit cannot hurt.

### DIFF
--- a/docs/tutorial/tutorial-1.rst
+++ b/docs/tutorial/tutorial-1.rst
@@ -19,8 +19,10 @@ that by writing in your current terminal::
 
     VOC_DIR=/path/to/voc
 
-In case you don't really want to do that for some reason, you'll need to
-remember to replace ``$VOC_DIR`` by the full VOC directory any time it
+Please remember to replace ``/path/to/voc`` with the actual path to your VOC 
+repository (e.g. /home/elias/src/voc-dev/voc). In case you don't really want 
+to do that for some reason, you'll need to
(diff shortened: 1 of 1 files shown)


docs/tutorial/tutorial-1.rst
```diff
@@ -19,8 +19,10 @@ that by writing in your current terminal::
 
     VOC_DIR=/path/to/voc
 
-In case you don't really want to do that for some reason, you'll need to
-remember to replace ``$VOC_DIR`` by the full VOC directory any time it
+Please remember to replace ``/path/to/voc`` with the actual path to your VOC 
+repository (e.g. /home/elias/src/voc-dev/voc). In case you don't really want
```
freakboy3742: Two small fixes here - firstly, replace `elias` with something generic like `myusername`; secondly, since this is a literal, it should be wrapped with backticks.

freakboy3742 (changes requested):
A good suggestion - thanks! Two minor suggestions for improvement inline.

docs/tutorial/tutorial-1.rst
BellaTrix25336: thanks for the feedback, much appreciated!

freakboy3742 (approved):
Fantastic! Thanks for the contribution!

